PR #18, opened 2022-05-13
[SYCLomatic] Fix an issue in math API migration

zhiweij1:
Use CanonicalType when checking if type-cast is needed for math
APIs in ENTRY_RENAMED_SINGLE and ENTRY_RENAMED_DOUBLE categories.


clang/test/dpct/cuda-math-intrinsics.cu
```diff
@@ -1067,7 +1067,7 @@ __global__ void kernelFuncFloat(float *deviceArrayFloat) {
 
   // CHECK: i = sycl::isinf(f0);
   i = isinf(f0);
-  // CHECK: i = sycl::isinf((float)i);
+  // CHECK: i = sycl::isinf((double)i);
```
zhimingwang36: suggest to requrest SYCL extesion to support isinf(integer).  
https://en.cppreference.com/w/cpp/numeric/math/isinf (C++11 has isinf(integer).

zhiweij1: > suggest to requrest SYCL extesion to support isinf(integer). https://en.cppreference.com/w/cpp/numeric/math/isinf (C++11 has isinf(integer).

OK, but for now, I think we can add `(double)` cast when migrating it.
When building isinf(Interger) with clang as CUDA, this call will match `__CUDA_CLANG_FN_INTEGER_OVERLOAD_1(bool, isinf);`
in clang/lib/Headers/__clang_cuda_cmath.h
It is expanded to
```
// Defines an overload of __fn that accepts one integral argument, calls
// __fn((double)x), and returns __retty.
#define __CUDA_CLANG_FN_INTEGER_OVERLOAD_1(__retty, __fn)                      \
  template <typename __T>                                                      \
  __DEVICE__                                                                   \
      typename __clang_cuda_enable_if<std::numeric_limits<__T>::is_integer,    \
                                      __retty>::type                           \
      __fn(__T __x) {                                                          \
    return ::__fn((double)__x);                                                \
  }
```

zhimingwang36: from test,  std::isinf(int) works and it will be const expr in C++23.  So for now,  I'd suggest just keep unchanged.

ziranzha (approved):

zhimingwang36 (approved):
LGTM

clang/test/dpct/cuda-math-intrinsics.cu
zhiweij1: Done

PR #23, opened 2022-05-16
[SYCLomatic] Fix the dpct::detail::gemm_batch_impl() in blas_utils.hpp

zhiweij1:
This patch uses pointers to heap memory instead of pointers to local variable
as the input arguments for oneapi::mkl::blas::column_major::gemm_batch() since
that API is an asynchronous API.


zhimingwang36 (approved):
LGTM

ziranzha (approved):

PR #24, opened 2022-05-16
[SYCLomatic] Fix the issue that empty Makefile is generated when the generated binary object files are not the in-root directory

tomflinda:


tomflinda:
The test case for this patch will be submitted into the test repo https://github.com/oneapi-src/SYCLomatic-test

PR #68, opened 2022-06-09
[SYCLomatic] Fix migration internal error due to dispatching nullptr in expression analysis

delecui:

### DIFF
--- a/clang/lib/DPCT/ExprAnalysis.cpp
+++ b/clang/lib/DPCT/ExprAnalysis.cpp
@@ -363,6 +363,8 @@ ExprAnalysis::ExprAnalysis(const Expr *Expression)
 }
 
 void ExprAnalysis::dispatch(const Stmt *Expression) {
+  if (!Expression)
+    return;
   switch (Expression->getStmtClass()) {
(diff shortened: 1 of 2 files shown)


clang/lib/DPCT/ExprAnalysis.cpp
```diff
@@ -363,6 +363,8 @@ ExprAnalysis::ExprAnalysis(const Expr *Expression)
 }
 
 void ExprAnalysis::dispatch(const Stmt *Expression) {
+  if (!Expression)
```
delecui, line 366: The change to fix `analyzeExpr(const IfStmt *IS)` at https://github.com/oneapi-src/SYCLomatic/blob/05ceacff8c4f6e27ee935afbca3f21399ffeacbe/clang/lib/DPCT/ExprAnalysis.cpp#L802

because `IS->getElse()` will return nullptr if there is no Else statment. `dispatch` should return early for nullptr input.
e.g.
`
if (v > 0) {
  return v;
}
`

tomflinda (approved):
LGTM

ziranzha (approved):
LGTM

PR #102, opened 2022-06-29
[SYCLomatic][Refactor] Refactor getStmtSpelling and remove getStmtExpansion

AndyCHHuang:
Original implementation:
In most cases, using getDefinitionRange in getStmtSpelling.
In some corner cases, using heuristic to get the range needed.

New implementation:
When parent range is given, using getRangeInRange.
When parent range is invalid, using getDefinitionRange.


clang/lib/DPCT/Utility.cpp
```diff
@@ -208,47 +208,47 @@ size_t calculateExpansionLevel(const SourceLocation Loc) {
...
+    EndLoc = Range.second;
+    Length = SM.getFileOffset(EndLoc) - SM.getFileOffset(BeginLoc);
   }
 
-  int Length =
-      SM.getFileOffset(EndLoc) - SM.getFileOffset(BeginLoc) +
-      Lexer::MeasureTokenLength(
-          EndLoc, SM, dpct::DpctGlobalInfo::getContext().getLangOpts());
+
   Str = std::string(SM.getCharacterData(BeginLoc), Length);
```
tomflinda, line 249: Suggest adding assert(Length > 0) before Line248 to detect the unexpected overflow.

ziranzha (approved):
LGTM

tomflinda (approved):
LGTM

PR #109, opened 2022-07-07
[SYCLomatic] Fix incorrect template arg location

AndyCHHuang:
Fix incorrect source location of template args when the
entire kernel call is a macro arg.


tomflinda (approved):
LGTM

ziranzha (approved):
LGTM

PR #131, opened 2022-07-20
[SYCLomatic] Adjust error-handling to avoid hanging issue on windows

AndyCHHuang:

### DIFF
--- a/clang/lib/DPCT/SignalProcess.cpp
+++ b/clang/lib/DPCT/SignalProcess.cpp
@@ -118,7 +118,7 @@ static LONG CALLBACK ExceptionHandler(PEXCEPTION_POINTERS ExceptionInfo) {
   case STATUS_ACCESS_VIOLATION:
     recoverCheckpoint(SIGSEGV);
   default:
-    return EXCEPTION_CONTINUE_EXECUTION;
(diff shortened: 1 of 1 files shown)


clang/lib/DPCT/SignalProcess.cpp
```diff
@@ -118,7 +118,7 @@ static LONG CALLBACK ExceptionHandler(PEXCEPTION_POINTERS ExceptionInfo) {
   case STATUS_ACCESS_VIOLATION:
     recoverCheckpoint(SIGSEGV);
   default:
-    return EXCEPTION_CONTINUE_EXECUTION;
+    return EXCEPTION_CONTINUE_SEARCH;
```
delecui, line 121: LGTM

ExceptionHandler should return EXCEPTION_CONTINUE_SEARCH here.

According to Win32 API documentation at https://docs.microsoft.com/en-us/windows/win32/api/winnt/nc-winnt-pvectored_exception_handler#return-value,

"To return control to the point at which the exception occurred, return EXCEPTION_CONTINUE_EXECUTION (0xffffffff). To continue the handler search, return EXCEPTION_CONTINUE_SEARCH (0x0)"

AndyCHHuang:
Another solution merged.

PR #161, opened 2022-08-18
[SYCLomatic] Fix get_info in device.hpp

LU-JOHN:
Fix get_info call in device.hpp to match SYCL 2020 standard.

### DIFF
--- a/clang/runtime/dpct-rt/include/device.hpp.inc
+++ b/clang/runtime/dpct-rt/include/device.hpp.inc
@@ -562,13 +562,12 @@ public:
     prop.set_minor_version(minor);
 
     prop.set_max_work_item_sizes(
-// SYCL 2020 uses structs as information descriptors. Previous versions of SYCL use enum class elements
-#if (SYCL_LANGUAGE_VERSION && SYCL_LANGUAGE_VERSION>=202001)
-        // SYCL 2020, max_work_item_sizes is a struct templated by an int
-        get_info<cl::sycl::info::device::max_work_item_sizes<3>>());
-#else
-        // Pre-SYCL 2020, max_work_item_sizes is an enum class element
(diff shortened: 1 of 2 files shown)


YuriPlyakhin (changes requested):
Changes are needed:
1.	Please, introduce DPC++ compiler-specific check to make sure the code works fine with older and newer DPC++ compilers. Just clearly comment it out that this code is specific to a certain compiler
2.	Please, do not add compatibility with SYCL 1.2.1

Hence, the code should be something like:

#if (DPC++ compiler specific check for older compiler)
        // comment saying that it is a branch specific for syclos compiler older than …, where max_work_item_sizes is an enum class element
        get_info<cl::sycl::info::device::max_work_item_sizes>());
#else
        // SYCL 2020-conformant code, max_work_item_sizes is a struct templated by an int
        get_info<cl::sycl::info::device::max_work_item_sizes<3>>());
#endif

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -562,7 +562,13 @@ public:
     prop.set_minor_version(minor);
 
     prop.set_max_work_item_sizes(
+#if (__SYCL_COMPILER_VERSION && __SYCL_COMPILER_VERSION<20220902)
+        // syclos compiler older than 2022/09/02, where max_work_item_sizes is an enum class element
```
YuriPlyakhin:
```suggestion
        // oneAPI DPC++ compiler older than 2022/09/02, where max_work_item_sizes is an enum class element
```

YuriPlyakhin (changes requested):

YuriPlyakhin (approved):

0x12CC (approved):

zhimingwang36 (approved):
LGTM

PR #170, opened 2022-08-24
[SYCLomatic] Fix the bug of dpct not inserting [[intel::reqd_sub_group_size(32)]] in some cases

delecui:

### DIFF
--- a/clang/lib/DPCT/AnalysisInfo.cpp
+++ b/clang/lib/DPCT/AnalysisInfo.cpp
@@ -2631,7 +2631,8 @@ void DeviceFunctionDecl::LinkDecl(const FunctionDecl *FD, DeclList &List,
       TSK_ExplicitInstantiationDefinition)
     return;
   if (FD->isImplicit() ||
-      FD->getTemplateSpecializationKind() == TSK_ImplicitInstantiation) {
+      (FD->getTemplateSpecializationKind() == TSK_ImplicitInstantiation &&
+       FD->getPrimaryTemplate())) {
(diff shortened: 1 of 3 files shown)


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2631,7 +2631,8 @@ void DeviceFunctionDecl::LinkDecl(const FunctionDecl *FD, DeclList &List,
       TSK_ExplicitInstantiationDefinition)
     return;
   if (FD->isImplicit() ||
-      FD->getTemplateSpecializationKind() == TSK_ImplicitInstantiation) {
+      (FD->getTemplateSpecializationKind() == TSK_ImplicitInstantiation &&
```
delecui, line 2634: Please refer to the code snippet in https://godbolt.org/z/P95YWEE4o

`FD` could be a non-template member function of local class(e.g., `MaxOp::warp_shfl_down`) that is defined inside CUDA kernel template function. 
`FD->getTemplateSpecializationKind() == TSK_ImplicitInstantiation` is `true` in such case, but `FD` itself is not template function instantiation.

delecui: FunctionDecl::getPrimaryTemplate()
https://clang.llvm.org/doxygen/classclang_1_1FunctionDecl.html#a3701fbf27221163b5f533f7d0a6f418c

zhiweij1 (approved):

ziranzha (approved):
LGTM

PR #171, opened 2022-08-24
[SYCLomatic] Support drcp* intrinsics w/o loss of precision

LU-JOHN:
sycl::native::recip() does not support double input.  Thus previously dpct would cast double inputs to a float in order to use sycl::native::recip().  Change dpct to now generate (1.0/X) if X is a double so that no precision is lost.


LU-JOHN:
@zhimingwang36 can you review this fix?  A testcase was added with:

https://github.com/oneapi-src/SYCLomatic-test/pull/66

clang/test/dpct/cuda-math-intrinsics.cu
```diff
@@ -660,36 +660,24 @@ __global__ void kernelFuncDouble(double *deviceArrayDouble) {
   d2 = __dmul_rz(d0, d1);
 
   // CHECK: /*
-  // CHECK-NEXT: DPCT1017:{{[0-9]+}}: The sycl::native::recip call is used instead of the __drcp_rd call. These two calls do not provide exactly the same functionality. Check the potential precision and/or performance issues for the generated code.
-  // CHECK-NEXT: */
-  // CHECK-NEXT: /*
   // CHECK-NEXT: DPCT1013:{{[0-9]+}}: The rounding mode could not be specified and the generated code may have different precision then the original code. Verify the correctness. SYCL math built-ins rounding mode is aligned with OpenCL C 1.2 standard.
   // CHECK-NEXT: */
-  // CHECK-NEXT: d1 = sycl::native::recip((float)d0);
+  // CHECK-NEXT: d1 = (1.0/(d0));
```
ziranzha: can we remove the parens here?

LU-JOHN: I don’t think we can remove the parenthesis.
If we have:

	10.0 * __drcp_rn(d+123.1)

We want it to be evaluated as:

	10.0 * (1.0/(d+123.1))

Without the outer parenthesis it will be:

	10.0*1.0/(d+123.1)

which will be evaluated as:

	(10.0 * 1.0)/(d+123.1)

ziranzha: But in this case, the parens are redundant.
So can we remove the parens if possible?

LU-JOHN: Updated dpct and testcase to remove inner-parens if argument is a variable.

ziranzha (approved):
LGTM

tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #185, opened 2022-08-31
[SYCLomatic] Support migration of cuda thrust pinned_allocator

delecui:


clang/runtime/dpct-rt/include/memory.hpp.inc
```diff
@@ -1906,6 +1906,63 @@ using constant_memory = detail::device_memory<T, constant, Dimension>;
...
+  using propagate_on_container_swap =
+      typename std::allocator_traits<Alloc>::propagate_on_container_swap;
+  using is_always_equal =
+      typename std::allocator_traits<Alloc>::is_always_equal;
+
+  template <typename U> struct rebind {
+    typedef allocator<U, AllocKind> other;
+  };
+
+  allocator() : _impl(dpct::get_default_queue()) {}
```
delecui: `_impl` is a `sycl::usm_allocator` constructed with `dpct::get_default_queue()`, so there is limitation that `_impl` will go invalid if users switch queue/context.

ziranzha (approved):
LGTM

clang/runtime/dpct-rt/include/memory.hpp.inc
```diff
@@ -1906,6 +1906,63 @@ using constant_memory = detail::device_memory<T, constant, Dimension>;
...
 using shared_memory = detail::device_memory<T, shared, Dimension>;
 // DPCT_LABEL_END
+
+// DPCT_LABEL_BEGIN|dpct_allocator|dpct
+// DPCT_DEPENDENCY_BEGIN
+// Device|get_default_queue
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+template <typename T, sycl::usm::alloc AllocKind = sycl::usm::alloc::host>
+class allocator {
```
reble: Naming the wrapper class just `allocator` seems too generic to me. Maybe we can be more specific like the mapping target `pinned_allocator`?

delecui: > Naming the wrapper class just `allocator` seems too generic to me. Maybe we can be more specific like the mapping target `pinned_allocator`?

What about naming the wrapper to `host_allocator` ? At the same time, we remove the second template argument and use `sycl::usm::alloc::host` as fixed allocation type inside the wrapper.

`sycl::usm_allocator<T, sycl::usm::alloc::host>` seems to be a close mapping for `pinned_allocator<T>`. Both allocate/deallocate memory in host side, except the gap there is no "pinned memory" in SYCL.

delecui: @reble
I uploaded da9515dbaf933dfdcf5b9bcde34978ea3dd8c0e6 to refine the wrapper class name. Please help review.
- Moved `dpct::allocator` to `dpct::detail::usm_allocator`
- Added `dpct::usm_host_allocator<T>` alias to `dpct::detail::usm_allocator<T, sycl::usm::alloc::host>`

Below is an migration example:
CUDA:
`std::vector<float, thrust::cuda::experimental::pinned_allocator<float>> hVecCopy = hVec;`
Migrated SYCL:
`std::vector<float, dpct::usm_host_allocator<float>> hVecCopy = hVec`

0x12CC (approved):

clang/test/dpct/helper_files_ref/include/memory.hpp
```diff
@@ -610,6 +610,58 @@ dpct_memcpy(cl::sycl::queue &q, void *to_ptr, const void *from_ptr,
...
+
+template <typename T, sycl::usm::alloc AllocKind>
+class usm_allocator {
+private:
+  using Alloc = sycl::usm_allocator<T, AllocKind>;
+  Alloc _impl;
+
+public:
+  using value_type = typename std::allocator_traits<Alloc>::value_type;
+  using pointer = typename std::allocator_traits<Alloc>::pointer;
```
reble, line 624: Some of these member functions have been deprecated in C++17 (removed in C++20) https://en.cppreference.com/w/cpp/memory/allocator . Can we add a warning to **not** make it look like we're reintroducing deprecated functionality.

reble (commented):
Changes look okay to me. Can we please add a warning in general that this is for mapping an experimental feature which has been removed in recent Thrust versions.

YuriPlyakhin:
> Changes look okay to me. Can we please add a warning in general that this is for mapping an experimental feature which has been removed in recent Thrust versions.

I would avoid making statements about other products in SYCLomatic warnings.

reble:
> > Changes look okay to me. Can we please add a warning in general that this is for mapping an experimental feature which has been removed in recent Thrust versions.
> 
> I would avoid making statements about other products in SYCLomatic warnings.

If there are any concerns with this regard, how about introducing a nested namespace `detail::deprecated` with the added functionality? It is generic and should make it obvious that we're not unburying any old C++ features. Just for the sake of SYCL migration.

YuriPlyakhin:
> > > Changes look okay to me. Can we please add a warning in general that this is for mapping an experimental feature which has been removed in recent Thrust versions.
> > 
> > 
> > I would avoid making statements about other products in SYCLomatic warnings.
> 
> If there are any concerns with this regard, how about introducing a nested namespace `detail::deprecated` with the added functionality? It is generic and should make it obvious that we're not unburying any old C++ features. Just for the sake of SYCL migration.

Yes, I think this is good idea to use "dpct::deprecated" for any public API and "dpct::detail::deprecated" for any implementation detail, where we want to highlight that the API we introduce is not something we want to support long term.

delecui:
> > > > Changes look okay to me. Can we please add a warning in general that this is for mapping an experimental feature which has been removed in recent Thrust versions.
> > > 
> > > 
> > > I would avoid making statements about other products in SYCLomatic warnings.
> > 
> > 
> > If there are any concerns with this regard, how about introducing a nested namespace `detail::deprecated` with the added functionality? It is generic and should make it obvious that we're not unburying any old C++ features. Just for the sake of SYCL migration.
> 
> Yes, I think this is good idea to use "dpct::deprecated" for any public API and "dpct::detail::deprecated" for any implementation detail, where we want to highlight that the API we introduce is not something we want to support long term.

Thanks for your comments and suggestions, @YuriPlyakhin and @reble
I've pushed [2ac8dd1](https://github.com/oneapi-src/SYCLomatic/pull/185/commits/2ac8dd1613a38f51b55e16ff12be82f09cd29584) to update the namespace. 
Please help review.

YuriPlyakhin (approved):
Would be good to add some comment in the helper header, that would explain what does "deprecated" namespace mean. Something along these lines: "dpct::deprecated namespace is for functionality that was introduced for compatibility purpose, but relies on deprecated C++ features, which are either removed or will be removed in the future standards. Please, avoid usage of deprecated functionality."

Also, we need to add description of "dpct::deprecated" in https://www.intel.com/content/www/us/en/develop/documentation/intel-dpcpp-compatibility-tool-user-guide/top/dpct-namespace-usage-guide.html. We have description of "dpct::detail" there already.

zhimingwang36 (approved):
LGTM

PR #189, opened 2022-09-03
[SYCLomatic]  In intercept-build do not filter out -shared option

LU-JOHN:
Update compilation.py to not filter out -shared and --shared options.

Also, updated __init__.py to find libear.so in a second directory for LIT testing (intercept-build uses libear.so).
LIT testing uses intercept-build from the SYCLomatic build area, which has libear.so at  lib/libear.so
The release area has the shared library at lib/libear/libear.so.   Previously  __init__.py could only find this location.
This was not encountered before because this  commit has the first intercept-build LIT test.

Also added a LIT test intercept.c to check that the -shared option is in the compilation database.

### DIFF
--- a/clang/test/dpct/intercept.c
+++ b/clang/test/dpct/intercept.c
@@ -1,3 +1,6 @@
+// UNSUPPORTED: -windows-
+// intercept-build not found on windows
+//
 // ------ prepare test directory
 // RUN: cd %T
 // RUN: rm -rf intercept-build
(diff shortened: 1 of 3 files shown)


YuriPlyakhin:
@zhimingwang36 , could you please assign someone to review/approve this request?

tomflinda (approved):
LGTM

AndyCHHuang (approved):
LGTM

PR #212, opened 2022-09-09
[SYCLomatic] Remove option "--parse-build-log <file>", "--in-root <path>" and their functional codes from intercept-build tool

tomflinda:


tomflinda:
Not needed anymore

PR #217, opened 2022-09-13
[SYCLomatic] Fix vector type constructors not migrating in certain contexts

jzc:
Migration of calls like `make_double3` would usually be handled by `VectorTypeCtorRule`. However, for example, migration of the argument `make_double3(1., 2., 3.)` passed to a  `cub::BlockReduce<>::Sum` method call (see added tests) would conflict with `CubRule`. Now, migration of these calls are eventually handled via `MathFunctionsRule` or `ExprAnalysis`.


clang/lib/DPCT/APINamesMath.inc
```diff
@@ -600,3 +600,54 @@ ENTRY_RENAMED("mul24", MapNames::getClNamespace(false, true) + "mul24")
 
 // impl with new rewriter
 ENTRY_REWRITE("rsqrt")
+
+// vector types
+ENTRY_RENAMED("make_char1", "char")
```
LU-JOHN, line 605: I think these entries would be clearer if they were grouped in this order:

char/unsigned char [1-4]
short/unsigned short [1-4]
int/unsigned int [1-4]
long/unsigned long [1-4]
long long/unsigned long long [1-4]
float [1-4]
double [1-4]

clang/test/dpct/vector_type_cub.cu
```diff
@@ -0,0 +1,534 @@
+// UNSUPPORTED: cuda-8.0, cuda-9.0, cuda-9.1, cuda-9.2, cuda-10.0, cuda-10.1, cuda-10.2
+// UNSUPPORTED: v8.0, v9.0, v9.1, v9.2, v10.0, v10.1, v10.2
+// RUN: dpct --format-range=none --usm-level=none -out-root %T/vector_type_cub %s --cuda-include-path="%cuda-path/include" --sycl-named-lambda -- -x cuda --cuda-host-only
+// RUN: FileCheck --input-file %T/vector_type_cub/vector_type_cub.dp.cpp --match-full-lines %s
+
+#include <cub/cub.cuh>
+
+__device__ char1 operator+(char1 a, char1 b) {
+  return make_char1(a.x + b.x);
```
LU-JOHN: Does it make sense to add CHECKs for device code?

Do your testcases exercise the changes for abs/min/max or are the changes needed in other testcases?

clang/lib/DPCT/APINamesMath.inc
```diff
@@ -600,3 +600,54 @@ ENTRY_RENAMED("mul24", MapNames::getClNamespace(false, true) + "mul24")
...
+ENTRY_RENAMED("make_uint4", MapNames::getClNamespace(false, false) + "uint4")
+ENTRY_RENAMED("make_ulong2", MapNames::getClNamespace(false, false) + "ulong2")
+ENTRY_RENAMED("make_ulong3", MapNames::getClNamespace(false, false) + "ulong3")
+ENTRY_RENAMED("make_ulong4", MapNames::getClNamespace(false, false) + "ulong4")
+ENTRY_RENAMED("make_ulonglong2", MapNames::getClNamespace(false, false) + "ulonglong2")
+ENTRY_RENAMED("make_ulonglong3", MapNames::getClNamespace(false, false) + "ulonglong3")
+ENTRY_RENAMED("make_ulonglong4", MapNames::getClNamespace(false, false) + "ulonglong4")
+ENTRY_RENAMED("make_ushort2", MapNames::getClNamespace(false, false) + "ushort2")
+ENTRY_RENAMED("make_ushort3", MapNames::getClNamespace(false, false) + "ushort3")
+ENTRY_RENAMED("make_ushort4", MapNames::getClNamespace(false, false) + "ushort4")
```
0x12CC:
```suggestion
ENTRY_RENAMED("make_ushort4", MapNames::getClNamespace(false, false) + "ushort4")

```

0x12CC (approved):

clang/test/dpct/vector_type_cub.cu
jzc: Good point, I think it does make sense to add checks here because the logic through `getNewFuncName` will flow through a different branch. 

abs/min/max are tested by `math-functions.cu`.

zhiweij1 (approved):

tomflinda (approved):
LGTM

PR #224, opened 2022-09-14
[SYCLomatic] keep include of CL/opencl.h

LU-JOHN:

Do not remove include of CL/opencl.h
Only CUDA code is migrated.  OpenCL code is not, thus the OpenCL
header must be kept.

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -1089,10 +1089,11 @@ void IncludesCallbacks::InclusionDirective(
     }
   }
 
-  // Always keep include of CL/opencl.h
+  // Always keep include of CL/*.  Do not delete even if
+  // they are found in a CUDA include directory.
   // Only CUDA code is migrated, not OpenCL.
-  // Thus CL/opencl.h header must be kept
-  if (FileName == "CL/opencl.h")
+  // Thus CL/* headers must be kept
+  if (FileName.startswith("CL/"))
     return;
   
   // Replace the complete include directive with an empty string.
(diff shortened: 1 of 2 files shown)


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -1082,6 +1082,17 @@ void IncludesCallbacks::InclusionDirective(
...
   }
 
+  // Previous code tried to keep any headers (approximately) not in a cuda directory, 
+  // and remove remove any headers in a cuda directory.  But we must also
+  // keep include of CL/opencl.h even though it can be found in a cuda directory like:
+  //
+  //   /usr/local/cuda/include/CL/opencl.h
+  //
+  // OpenCL header must be kept since only cuda code is migrated, not OpenCL
+
```
0x12CC:
```suggestion
```

0x12CC (approved):

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -1089,6 +1089,12 @@ void IncludesCallbacks::InclusionDirective(
     }
   }
 
+  // Always keep include of CL/opencl.h
+  // Only CUDA code is migrated, not OpenCL.
+  // Thus CL/opencl.h header must be kept
+  if (FileName == "CL/opencl.h")
```
zhimingwang36: can the patch more generic, works for all CL/* files,  not specidifc opencl.h file. 
There are other file in CL/ should be keep, like.
cl_egl.h  cl_ext.h  cl_gl_ext.h  cl_gl.h  cl.h  cl.hpp  cl_platform.h  opencl.h

zhimingwang36: suggest to compare "CL/ "folder, if so, keep it.

LU-JOHN: Updated patch to keep includes of all CL* files.  Test updated also.

zhimingwang36 (approved):
LGTM

PR #227, opened 2022-09-16
[SYCLomatic] Make sure dpct inserts SYCL_EXTERNAL correctly for lambda expression

delecui:


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2175,7 +2175,8 @@ void CallFunctionExpr::buildInfo() {
 
   const std::string &DefFilePath = FuncInfo->getDefinitionFilePath();
   if (!DefFilePath.empty() && DefFilePath != getFilePath() &&
```
delecui, line 2177: SYCL_EXTERNAL is not needed if the definition of device function callee is in current compilation unit.

the device function might be
- called in .cuh header file([clang/test/dpct/wrong_sycl_external/test3.cuh](https://github.com/oneapi-src/SYCLomatic/pull/227/files#diff-edd62b9cd22fc7c5032428911618431753e7e06e023402ee0dcd7ba65ca8c803))
`__device__ acc_t reduce(acc_t x, acc_t y) const { return combine(x, y); }`
- defined in .cu file ([clang/test/dpct/wrong_sycl_external/test3.cu](https://github.com/oneapi-src/SYCLomatic/pull/227/files#diff-8fe4219e39a7aa9a3be74cd7c79a388c149156128c3564c7043d5e886a3b6c43))
`func_wrapper<T>([] __host__ __device__ (T a, T b) -> T { return a * b; }).reduce(x, y);`
- the `DefFilePath`(test.cu) includes `getFilePath()`(test.cuh) in this case where the device function is a lambda expression.
 
`isIncludedFile(getFilePath(), DefFilePath)` is not enough in this case, so the fix is to check "including" relationship in both directions.

ziranzha (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #234, opened 2022-09-18
[SYClomatic] Fix migration error of cudaStream_t with initArg 0

wenhuiNi:
cudaStream_t s{0} should be migrated by using get_default_queue


wenhuiNi:
Chen Sheng will refine cudaStream_t migration ,close it

PR #238, opened 2022-09-19
[SYCLomatic] make max_nd_range_size type compatible with CUDA

LU-JOHN:

Update SYCLomatic runtime to make the default return type of max_nd_range_size and max_work_item_sizes to be "int *" which
is compatible with CUDA.  This allows max_nd_range_size[X]/max_work_item_sizes[X] to be used in std::max and std::min expressions.

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -4041,19 +4041,10 @@ void DeviceInfoVarRule::runRule(const MatchFinder::MatchResult &Result) {
     // migrate to get_XXX() eg. "b=a.minor" to "b=a.get_minor_version()"
     requestFeature(PropToGetFeatureMap.at(MemberName), ME);
     std::string TmplArg = "";
-    if (MemberName == "maxGridSize" || MemberName == "maxThreadsDim") {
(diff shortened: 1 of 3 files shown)


clang/test/dpct/helper_files_ref/include/device.hpp
```diff
@@ -59,7 +59,7 @@ class device_info {
   // get interface
   const char *get_name() const { return _name; }
   char *get_name() { return _name; }
-  template <typename WorkItemSizesTy = sycl::id<3>,
+  template <typename WorkItemSizesTy = int *,
```
tomflinda: This change will break the compatibility of dpct helper functions.
I advise that this header file device.hpp is not touched, change the dpct migration process.
For example:
Migrate 'int *maxGridSize_ptr = deviceProp.maxGridSize;' 
to 
'int *maxGridSize_ptr = deviceProp.get_max_nd_range_size<int*>();'

LU-JOHN: This fix has been implemented within dpct now.  device.hpp is not changed.

tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #252, opened 2022-09-25
[SYCLomatic] Merge branch 'origin/sycl' into SYCLomatic

0x12CC:


clang/lib/DPCT/Rules.cpp
```diff
@@ -488,7 +488,7 @@ void clang::dpct::UserDefinedTypeRule::registerMatcher(
 
 void clang::dpct::UserDefinedTypeRule::runRule(
     const clang::ast_matchers::MatchFinder::MatchResult &Result) {
-  if (auto TL = getNodeAsType<TypeLoc>(Result, "typeLoc")) {
+  if (auto TL = getNodeAsType<TypeLoc>(Result, "typeLoc", false)) {
```
AndyCHHuang: I think this change is not relevant to the lit fail.

0x12CC: I've updated the `runRule` function so that this workaround is not needed.

YuriPlyakhin (approved):

PR #270, opened 2022-10-05
[SYCLomatic] Fix use after move issues

0x12CC:

### DIFF
--- a/clang/lib/DPCT/Diagnostics.h
+++ b/clang/lib/DPCT/Diagnostics.h
@@ -369,7 +369,7 @@ bool report(const std::string FileAbsPath, unsigned int Offset, IDTy MsgID,
   std::string FileAndLine = clang::dpct::buildString(
       NativeFormPath, ":", Fileinfo->getLineNumber(Offset));
   std::string WarningIDAndMsg = clang::dpct::buildString(
-      std::to_string(static_cast<int>(MsgID)), ":", std::forward<Ts>(Vals)...);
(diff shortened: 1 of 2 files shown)


clang/lib/DPCT/Diagnostics.h
```diff
@@ -369,7 +369,7 @@ bool report(const std::string FileAbsPath, unsigned int Offset, IDTy MsgID,
   std::string FileAndLine = clang::dpct::buildString(
       NativeFormPath, ":", Fileinfo->getLineNumber(Offset));
   std::string WarningIDAndMsg = clang::dpct::buildString(
-      std::to_string(static_cast<int>(MsgID)), ":", std::forward<Ts>(Vals)...);
+      std::to_string(static_cast<int>(MsgID)), ":", Vals...);
```
jzc, line 372: This `std::forward` was removed because `Vals` can be used on the lines 390 and 399:
https://github.com/oneapi-src/SYCLomatic/blob/f3b56482e82025cb193ddbef2db12e1dd5390e22/clang/lib/DPCT/Diagnostics.h#L385-L400

However, wouldn't the `std::forward` on line 390 also cause a use after move on line 399 if `IsInsertWarningIntoCode` is true?

0x12CC: Yes. I've fixed this.

jzc (approved):

AndyCHHuang (approved):
LGTM

PR #264, opened 2022-09-30
[SYCLomatic] Fix constant_iterator: declare __my_value_ as non const

SergeyKopienko:

PR #271, opened 2022-10-05
[SYCLomatic] Fix constant_iterator: declare __my_value_ as non const

0x12CC:
Rebases this PR: https://github.com/oneapi-src/SYCLomatic/pull/264.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/iterators.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/iterators.h.inc
@@ -155,7 +155,7 @@ public:
   }
 
 private:
-  const _Tp __my_value_;
+  _Tp __my_value_;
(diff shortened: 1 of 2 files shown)


jzc (approved):

0x12CC:
@SergeyKopienko, could you clarify why this change is needed?

SergeyKopienko:
> @SergeyKopienko, could you clarify why this change is needed?

When you pass constant_iterator into oneDPL library algorithms, sometimes occurred operations which want to assign state of one iterator instance to another instance. But as far as one field of const_iterator defined as const, this assignments are unavailable. This is the root cause of compilation error.

YuriPlyakhin (approved):
The fix is to allow assigning one instance of iterator to another instance. The data accessed by iterator still can not be modified which is controlled by public methods.

PR #273, opened 2022-10-08
[SYClomatic] Fix coverity 'Dereference before null check' and 'Uninitialized scalar field'

wenhuiNi:


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -9610,9 +9610,6 @@ void KernelCallRule::runRule(
       }
     }
 
-    if (!FD)
```
tangjj11, line 9597: `getAssistNodeAsType` may return nullptr, so do we need move this null pointer check before line 9589 (`if (!FD->isTemplateInstantiation()){`) ?

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -9943,10 +9940,11 @@ void MemVarRule::removeHostConstantWarning(Replacement &R) {
   R.setReplacementText(Result);
 }
 
-void MemVarRule::processTypeDeclaredLocal(const VarDecl *MemVar,
+bool MemVarRule::processTypeDeclaredLocal(const VarDecl *MemVar,
                                           std::shared_ptr<MemVarInfo> Info) {
   auto &SM = DpctGlobalInfo::getSourceManager();
   auto DS = Info->getDeclStmtOfVarType();
+  if(!DS) return false;
```
tangjj11: this line should be formatted.

clang/lib/DPCT/ASTTraversal.cpp
wenhuiNi: when `getAssistNodeAsType` return nullptr, `if (auto FD = getAssistNodeAsType<FunctionDecl>(Result, "deviceFuncDecl"))` will be false

yihwang-nv (approved):
LGTM

AndyCHHuang (approved):
LGTM

tangjj11 (approved):
LGTM

ziranzha (approved):
LGTM

PR #297, opened 2022-10-17
[SYCLomatic] Refine kernel migration that making the lambda function as the argument passed to kernel function directly instead of declaring a variable for lambda function passed to kernel function

tomflinda:


zhiweij1 (approved):

ziranzha (approved):
LGTM

PR #312, opened 2022-10-20
Adding metafunction to correct type used in buffer_allocator when dev…

timmiesmith:
…ice_pointer value type is void.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h.inc
@@ -50,13 +50,31 @@
 // DPCT_LABEL_END
 namespace dpct {
 namespace detail {
-// DPCT_LABEL_BEGIN|__buffer_allocator|dpct::detail
+template <typename T>
+// DPCT_LABEL_BEGIN|make_allocatable|dpct::detail
 // DPCT_DEPENDENCY_EMPTY
+// DPCT_CODE
+struct make_allocatable
+{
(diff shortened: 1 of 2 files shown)


timmiesmith:
@ShengchenJ please review.

clang/runtime/dpct-rt/include/dpl_extras/memory.h.inc
```diff
@@ -50,13 +50,31 @@
 // DPCT_LABEL_END
 namespace dpct {
 namespace detail {
-// DPCT_LABEL_BEGIN|__buffer_allocator|dpct::detail
+template <typename T>
+// DPCT_LABEL_BEGIN|make_allocatable|dpct::detail
 // DPCT_DEPENDENCY_EMPTY
```
ShengchenJ: The code depends on the dpct::byte_t, please help update it.

ShengchenJ (changes requested):
The patch is LGTM, but the Lit test failed, the check number is updated. Could you update the lit cases? Many thanks.

clang/runtime/dpct-rt/include/dpl_extras/memory.h.inc
timmiesmith: Dependence information has been updated.

zhiweij1:
I have merged https://github.com/oneapi-src/SYCLomatic/pull/316 based on this one and fixed all LIT failures.

PR #315, opened 2022-10-21
[SYCLomatic] report warning when macro in array index are replaced.

tangjj11:


ziranzha (approved):
LGTM

wenhuiNi (approved):

clang/test/dpct/curand-device.cu
```diff
@@ -76,6 +76,9 @@ int main(int argc, char **argv) {
   int *dOut;
   //CHECK:q_ct1.submit(
   //CHECK-NEXT:  [&](sycl::handler &cgh) {
+  //CHECK-NEXT:    /*
+  //CHECK-NEXT:    DPCT1099:{{[0-9]+}}: The size of the accessor is a literal value to avoid potential out of scope issue of the used macro/const value.
```
YuriPlyakhin:
```suggestion
  //CHECK-NEXT:    DPCT1099:{{[0-9]+}}: 'WARP_SIZE' expression was replaced with a value. Modify the code to use original expression, provided in comments, if it is correct.
```

clang/test/dpct/macro_test.cu
```diff
@@ -1153,6 +1153,10 @@ template<class T1, class T2, int N> __global__ void foo31();
 #define FOO31(DIMS) foo31<unsigned int, float, DIMS><<<1,1>>>();
 
 //CHECK:   q_ct1.submit([&](sycl::handler &cgh) {
+//CHECK-NEXT:     /*
+//CHECK-NEXT:     DPCT1099:{{[0-9]+}}: The size of the accessor is a literal value to avoid
+//CHECK-NEXT:     potential out of scope issue of the used macro/const value.
+//CHECK-NEXT:     */
 //CHECK-NEXT:     sycl::local_accessor<double, 2> red_acc_acc_ct1(
 //CHECK-NEXT:         sycl::range<2>(8 /*8*/, 8 /*BLOCK_PAIR / SIMD_SIZE*/), cgh);
```
YuriPlyakhin, line 1166:
```suggestion
//CHECK-NEXT:         sycl::range<2>(8, 8 /*BLOCK_PAIR / SIMD_SIZE*/), cgh);
```
/*8*/ is not a macro, why do we need to add it in comment in addition to value?

clang/test/dpct/sharedmem-dim-noUSM.cu
```diff
@@ -83,6 +83,9 @@ __global__ void kernel3() {
 }
 
 void foo3() {
+  // CHECK: /*
+  // CHECK-NEXT: DPCT1099:{{[0-9]+}}: The size of the accessor is a literal value to avoid potential out of scope issue of the used macro/const value.
```
YuriPlyakhin: Size of the accessor in this case is not a macro, there is no "out-of-scope" problem with using original expression in this case? Why do we need to replace it with value and provide warning?

YuriPlyakhin (changes requested):

clang/test/dpct/macro_test.cu
tangjj11: It is a bug about macro source location finding exactly. But it is it not related to this issue, and I prefer to fix it in a separated PR.

clang/test/dpct/sharedmem-dim-noUSM.cu
tangjj11: The `sizeof` can not be keep in some cases and it hard to idetify the scenarios. For example:
```c++
__global__ void foo29();
void foo30(){
  foo29<<<1,1,0>>>();
}
class A{};
__global__ void foo29() {
  __shared__ double red_acc[sizeof(A)];
}
```
There will be "out-of-scope" problem if we don't replace the expression. (because class A is invisible in function `foo30`)
But if the expression only contain unary/binary/ternary expression consist of integer literal, we can keep the expression not changed, and I prefer to fix it in a new PR.

clang/test/dpct/curand-device.cu
tangjj11: fixed

clang/test/dpct/sharedmem_var_static.cu
```diff
@@ -82,7 +82,13 @@ void testTemplate() {
 
   // CHECK: dpct::get_default_queue().submit(
   // CHECK-NEXT:   [&](sycl::handler &cgh) {
+  // CHECK-NEXT:     /*
+  // CHECK-NEXT:     DPCT1099:{{[0-9]+}}: 'size * 4' expression was replaced with a value. Modify the code to use original expression, provided in comments, if it is correct.
+  // CHECK-NEXT:     */
   // CHECK-NEXT:     sycl::local_accessor<T, 2> s_acc_ct1(sycl::range<2>(64/*size * 2*/, 128/*size * 4*/), cgh);
```
YuriPlyakhin, line 91: There were 2 different expressions replaced, but only one warning. I think, there should be 2 warnings (1 per expression) or one warning, that would list both expressions.

YuriPlyakhin (changes requested):

clang/test/dpct/sharedmem_var_static.cu
tangjj11: my mistake, fixed.

clang/test/dpct/macro_test.cu
```diff
@@ -1153,6 +1153,11 @@ template<class T1, class T2, int N> __global__ void foo31();
 #define FOO31(DIMS) foo31<unsigned int, float, DIMS><<<1,1>>>();
 
 //CHECK:   q_ct1.submit([&](sycl::handler &cgh) {
+//CHECK-NEXT:     /*
+//CHECK-NEXT:     DPCT1099:{{[0-9]+}}: '8, BLOCK_PAIR / SIMD_SIZE' expression was replaced with a
```
YuriPlyakhin:
```suggestion
//CHECK-NEXT:     DPCT1099:{{[0-9]+}}: '8', 'BLOCK_PAIR / SIMD_SIZE' expressions were replaced with
```

clang/test/dpct/macro_test.cu
```diff
@@ -1153,6 +1153,11 @@ template<class T1, class T2, int N> __global__ void foo31();
...
+//CHECK-NEXT:     value. Modify the code to use original expression, provided in comments,
```
YuriPlyakhin, line 1162:
```suggestion
//CHECK-NEXT:     values. Modify the code to use original expressions, provided in comments,
```

clang/test/dpct/sharedmem_var_static.cu
YuriPlyakhin: If you use 1 warning for several expressions, the wording needs to be updated. See my comment above with changes required. The point is, this is not one expression, these are several expressions.

YuriPlyakhin (changes requested):

clang/test/dpct/sharedmem_var_static.cu
tangjj11: I think it is unnecessary to add 2 kinds of Diags to put the 2 different expressions warning in one warning. So I will put them in 2 warning.

clang/lib/DPCT/Diagnostics.inc
```diff
@@ -211,3 +211,5 @@ DEF_WARNING(PRIMITIVE_WORKSPACE, 1097, "The function \"%0\" may require the work
 DEF_COMMENT(PRIMITIVE_WORKSPACE, 1097, "The function \"{0}\" may require the workspace used to save intermediate results from function \"{1}\". By default, a workspace from engine_ext is selected according to the source data pointer, but this may be incorrect and cause a workspace data race. You may need to rewrite this code.")
 DEF_WARNING(MATH_EMULATION_EXPRESSION, 1098, "The %1 expression is used instead of the %0 call. These two expressions do not provide the exact same functionality. Check the generated code for potential precision and/or performance issues.")
 DEF_COMMENT(MATH_EMULATION_EXPRESSION, 1098, "The {1} expression is used instead of the {0} call. These two expressions do not provide the exact same functionality. Check the generated code for potential precision and/or performance issues.")
+DEF_WARNING(MACRO_EXPR_REPLACED, 1099, "'%0' expression was replaced with a value. Modify the code to use original expression, provided in comments, if it is correct.")
```
YuriPlyakhin: 1099 message ID is already reserved by other diagnostics (FFT-related). Please, update the ID.

YuriPlyakhin (changes requested):

YuriPlyakhin (changes requested):
Old ID is still in tests

clang/lib/DPCT/Diagnostics.inc
tangjj11: updated, and because the ID must be consecutive, so I need move the syclct IDs to here.

tangjj11:
updated the ID, and because the ID should be consecutive to avoid pull down conflict. I need add the syclct IDs to here.

YuriPlyakhin (approved):
Warning text looks good to me now.

PR #333, opened 2022-11-01
[SYCLomatic] Change queue_ptr to to select default queue on 0 initialization

jzc:


clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -1060,6 +1054,26 @@ static inline sycl::queue &get_default_queue() {
...
+// DPCT_CODE
+
+class queue_ptr {
+  sycl::queue* _q;
+public:
+  queue_ptr(sycl::queue* _p)
+    : _q(_p ? _p : &get_default_queue()) {}
+  queue_ptr()
+    : queue_ptr(0) {}
+  operator sycl::queue*  () const { return _q; }
```
ziranzha: Why we need this two conversions if we are going to hide sycl::queue pointer in all migration results?

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -1060,6 +1054,26 @@ static inline sycl::queue &get_default_queue() {
...
+// DPCT_DEPENDENCY_BEGIN
+// Device|get_default_queue
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+
+class queue_ptr {
+  sycl::queue* _q;
+public:
+  queue_ptr(sycl::queue* _p)
+    : _q(_p ? _p : &get_default_queue()) {}
```
ziranzha, line 1067: dpct::get_default_queue should also return queue_ptr since we should hide sycl::queue in all migration results.

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -1060,6 +1054,26 @@ static inline sycl::queue &get_default_queue() {
...
+  queue_ptr()
+    : queue_ptr(0) {}
+  operator sycl::queue*  () const { return _q; }
+  operator sycl::queue*& ()       { return _q; }
+  sycl::queue* operator->() const { return _q; }
```
ziranzha, line 1080: We may need add "operator bool" for this class.

jzc: With `operator sycl::queue*`, the following conversion sequence implicitly converts a `queue_ptr` to bool: 

`queue_ptr` to `sycl::queue*` (user-defined conversion)
`sycl::queue*` to `bool` (numeric conversion of pointers to bool)

clang/runtime/dpct-rt/include/device.hpp.inc
jzc: Two reasons: (1) for conversion to `sycl::queue*`, this allows us to use operations of a pointer. We can dereference a queue_ptr, use subscripting, add/subtract offset, etc. For conversion to `sycl::queue*&`, this allow use of `+=`, `-=`, `++`, etc. (2) End-to-end test caught the following issue: in other helper header files, there are API that use `sycl::queue*` instead of `queue_ptr`. Specifically `dpct::device_ext::destroy_queue` takes as parameter a reference to a `sycl::queue*`, and one test passed a `queue_ptr` to this. This conversion allows a `queue_ptr` to bind to this.

ziranzha: For reason 1:
`cudaStream_t`/`CUstream_t` is a pointer of an opaque class, which can not be dereferenced. Therefore, `queue_ptr` doesn't need to support such operators.
For reason 2:
I think we should replace `sycl::queue*` with `queue_ptr` in helper functions for consistency.

jzc: If `queue_ptr` is not dereferenceable, then there is no way to pass the `sycl::queue` is points to by reference to other functions. There are some helper functions that we take as argument a reference to `sycl::queue` - examples include `dpct::dpct_memcpy` and the BLAS utils.

clang/runtime/dpct-rt/include/device.hpp.inc
jzc: I'm not sure I why this change should be made/the reason behind it. Why should `get_default_queue` not return a reference to a `sycl::queue`? Also, this would incur many changes, as there are many places our migration expects `get_default_queue` to a be a reference to `sycl::queue` (e.g. member access through `.` : `q_ct1.parallel_for(...)`).

clang/runtime/dpct-rt/include/device.hpp.inc
ziranzha: I think all the helper function interface should use `queue_ptr` instead of `sycl::queue`. Considering backward compatibility, we need to add `operator*` for `queue_ptr` which returns `queue_ptr&` and a constructor which accept `sycl::queue&`.

PR #338, opened 2022-11-02
[SYCLomatic] Ensure wrappers are only generated under C language linkage and for __global__ functions

jzc:

### DIFF
--- a/clang/lib/DPCT/AnalysisInfo.cpp
+++ b/clang/lib/DPCT/AnalysisInfo.cpp
@@ -2448,7 +2448,7 @@ inline void DeviceFunctionDecl::emplaceReplacement() {
 
 inline void DeviceFunctionDeclInModule::emplaceReplacement() {
   DeviceFunctionDecl::emplaceReplacement();
-  if(GenerateWrapper) { insertWrapper(); }
+  insertWrapper();
 }
 
 void DeviceFunctionDeclInModule::buildParameterInfo(const FunctionDecl *FD) {
@@ -2459,9 +2459,6 @@ void DeviceFunctionDeclInModule::buildParameterInfo(const FunctionDecl *FD) {
 }
 
 void DeviceFunctionDeclInModule::buildWrapperInfo(const FunctionDecl *FD) {
-  GenerateWrapper =
-    (FD->getLanguageLinkage() == CLanguageLinkage
-     && FD->hasAttr<CUDAGlobalAttr>());
   auto &SM = DpctGlobalInfo::getSourceManager();
   const FunctionDecl *Def;
   HasBody = FD->hasBody(Def);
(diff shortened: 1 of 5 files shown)


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2459,6 +2459,9 @@ void DeviceFunctionDeclInModule::buildParameterInfo(const FunctionDecl *FD) {
 }
 
 void DeviceFunctionDeclInModule::buildWrapperInfo(const FunctionDecl *FD) {
+  GenerateWrapper =
```
0x12CC: This condition looks correct but can it be in the definition of `isModuleFunction`? I think that moving it there would eliminate the need for a new data member.

jzc: I think that makes sense, I will implement that.

0x12CC (approved):

clang/test/dpct/module_wrapper_gen.cu
```diff
@@ -0,0 +1,26 @@
+// RUN: dpct --extra-arg="--ptx" \
+// RUN:      --out-root=%T/module_wrapper_gen \
+// RUN:      --cuda-include-path="%cuda-path/include" %s
+// RUN: FileCheck %s --input-file=%T/module_wrapper_gen/module_wrapper_gen.dp.cpp
+
+// START
+
```
LU-JOHN, line 7: Can we add a functions:

__device__
void int_test() {  // check non-extern-C device non-operator is not wrapped
}

__global__
void int_wow() { // check non-extern-C global is not wrapped
}

to ensure that these are not wrapped.

jzc: I've added these cases in the test

clang/test/dpct/module_wrapper_gen.cu
```diff
@@ -0,0 +1,30 @@
...
+__device__ float2 operator+(float2 a, float2 b) {
+  return float2{a.x - b.x, a.y - b.y};
+}
+
+extern "C" __device__ void externCNonKernel() {}
+
+__device__ void deviceFun() {}
+
+__global__ void nonExternCKernel() {}
```
ziranzha, line 16: move to line:23, or it is not checked.

LU-JOHN (approved):
LGTM

clang/test/dpct/module_wrapper_gen.cu
jzc: I'm not sure that affects the behavior of the test -- the test should pass if there is exactly one `_wrapper` between the `// START` and `// END` comments, and moving this to line 23 should not affect that.

PR #344, opened 2022-11-07
[NFC][SYCLomatic] Clean llvm::Optional<T>::getValue() deprecated warning

yihwang-nv:


AndyCHHuang (changes requested):
should protect the code change with sycl_customization

PR #346, opened 2022-11-08
[SYCLomatic]Fix bug that the template parameters are not handled when the template is explicitly instantiated

wenhuiNi:

### DIFF
--- a/clang/lib/DPCT/AnalysisInfo.cpp
+++ b/clang/lib/DPCT/AnalysisInfo.cpp
@@ -2290,33 +2290,12 @@ void ExplicitInstantiationDecl::processFunctionTypeLoc(
 
 void ExplicitInstantiationDecl::processTemplateArgumentList(
     const TemplateArgumentListInfo &TAList) {
-  auto &SM = DpctGlobalInfo::getSourceManager();
   ExprAnalysis EA;
   for (const clang::TemplateArgumentLoc &ArgLoc : TAList.arguments()) {
     EA.analyze(ArgLoc);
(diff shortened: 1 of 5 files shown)


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2288,6 +2288,37 @@ void ExplicitInstantiationDecl::processFunctionTypeLoc(
...
+    if (EA.hasReplacement()) {
+      auto SR = getDefinitionRange(ArgLoc.getSourceRange().getBegin(),
+                                   ArgLoc.getSourceRange().getEnd());
+      auto BeginInfo = clang::dpct::DpctGlobalInfo::getLocInfo(SR.getBegin());
+      auto EndInfo = clang::dpct::DpctGlobalInfo::getLocInfo(SR.getEnd());
+      Token Tok2;
+      Lexer::getRawToken(ArgLoc.getSourceRange().getEnd(), Tok2, SM,
+                         DpctGlobalInfo::getContext().getLangOpts());
+      // if the last token is ">>" or ">>>",
+      // since DPCT does not support nested template type migration,
```
AndyCHHuang: since the raw token has not been splitted to multiple ">"s by the parser and no shift operator ">>" will used in template args, we simply assume the ">>" or ">>>" are the end token of nested template args. Therefore, the token length is 1.

AndyCHHuang (approved):
LGTM

ziranzha (approved):
LGTM

PR #352, opened 2022-11-10
[SYCLomatic] Use Kernel Library API

LU-JOHN:
Use dpct::load_sycl_lib*, dpct::get_kernel_function, dpct::invoke_kernel_function


clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +111,142 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+}
+
+static void write_vector_to_file(const std::string &name, const std::vector<char> &blob) {
+  auto file = std::fstream(name, std::ios::out | std::ios::binary);
+  file.write(blob.data(), blob.size());
+  file.close();
+}
+
+static uint64_t get_size_from_elf(char const *const blob) {
+  if (blob[0] != 0x7F || blob[1] != 'E' || blob[2] != 'L' || blob[3] != 'F')
```
ziranzha, line 299: The function can not work on Windows, is that expected?

LU-JOHN: Separate get_size_from_dll function was added for Windows.

clang/lib/DPCT/APINamesDriver.inc
```diff
@@ -54,8 +29,6 @@ FEATURE_REQUEST_FACTORY(HelperFeatureEnum::Image_image_wrapper_base_p_alias,
 
 CALL_FACTORY_ENTRY("cuModuleUnload", CALL("dlclose", ARG(0)))
```
0x12CC: Does this migration need to be updated?

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,231 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  sprintf(buffer,"%lld\n",(unsigned long long) rawtime);
+
+  std::string str(buffer);
+  std::hash<std::string> hasher;
+
+  int i = hasher(str);
+
+  sprintf(buffer,"%08X",i);
+
+  static char tempfile[200];
```
0x12CC: I think you should use a `std::string` instead of returning a pointer to static memory.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,231 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  ret |= static_cast<uint64_t>(ptr[4]) << 32;
+  ret |= static_cast<uint64_t>(ptr[5]) << 40;
+  ret |= static_cast<uint64_t>(ptr[6]) << 48;
+  ret |= static_cast<uint64_t>(ptr[7]) << 56;
+
+  return(ret);
+}
+
+#ifdef _WIN32
+static uint64_t get_size_from_dll(char const *const blob) {
```
0x12CC: You can remove the conditional compilation at the call site for these two functions if you make the Linux and Windows versions share the same declaration.

clang/test/dpct/helper_files_ref/include/kernel.hpp
```diff
@@ -36,5 +76,231 @@ static kernel_function_info get_kernel_function_info(const void *function) {
   return kernel_info;
 }
 
+static std::vector<char> read_file_to_vector(const std::string &name) {
```
0x12CC: I think that you should move all of your non-public helper functions into a namespace to distinguish them from those that you expect the user's code to call. Other runtime headers use the `detail` namespace for similar declarations.

clang/test/dpct/helper_files_ref/include/kernel.hpp
```diff
@@ -36,5 +76,231 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+#endif
+
+  std::vector<char> blob(size);
+  std::memcpy(blob.data(), image, size);
+  return load_dl_from_vector(blob);
+};
+
+static dpct::kernel_function get_kernel_function(module &module, const std::string &name) {
+#ifdef _WIN32
+  dpct::kernel_functor fn = (dpct::kernel_functor) GetProcAddress(static_cast<HMODULE>(module.get_ptr()), (name + std::string("_wrapper")).c_str());
```
0x12CC: You can use a `reinterpret_cast<>` expression to make it easier for developers to find all of the type conversions.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,231 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+
+  unsigned char const *const ublob = reinterpret_cast<unsigned char const *const>(blob);
+  uint64_t e_shoff     = extract64(ublob+0x28);
+  uint16_t e_shentsize = extract16(ublob+0x3A);
+  uint16_t e_shnum     = extract16(ublob+0x3C);
+
+  return e_shoff + (e_shentsize * e_shnum);
+}
+#endif
+class module {
```
0x12CC: Is this a valid name? I think `module` is a keyword in C++.

clang/lib/DPCT/Diagnostics.inc
```diff
@@ -217,3 +217,9 @@ DEF_WARNING(FFT_EXTERNAL_WORKSPACE, 1100, "Currently the DFT external workspace
 DEF_COMMENT(FFT_EXTERNAL_WORKSPACE, 1100, "Currently the DFT external workspace feature in the Intel(R) oneAPI Math Kernel Library (oneMKL) is only supported for GPU devices. Use the internal workspace if your code should run on non-GPU devices.")
 DEF_WARNING(MACRO_EXPR_REPLACED, 1101, "'%0' expression was replaced with a value. Modify the code to use original expression, provided in comments, if it is correct.")
 DEF_COMMENT(MACRO_EXPR_REPLACED, 1101, "'{0}' expression was replaced with a value. Modify the code to use original expression, provided in comments, if it is correct.")
+DEF_WARNING(MODULE_LOAD, 1102, "'%0' should be a dynamic library.  The dynamic-library should supply \"wrapped\" kernel-functions.")
```
0x12CC: There's an extraneous space between your two sentences. Also, I don't think you need to hyphenate "dynamic library" or "kernel function". Please make sure that all occurrences of these terms are consistent.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Updated to return std::string

clang/lib/DPCT/APINamesDriver.inc
LU-JOHN: Migration added for cuModuleUnload.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: done

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: done

clang/lib/DPCT/Diagnostics.inc
LU-JOHN: Fixed.

clang/test/dpct/helper_files_ref/include/kernel.hpp
LU-JOHN: Non-public helper functions moved to detail namespace.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -62,13 +71,44 @@ struct kernel_function_info {
...
+
 // DPCT_LABEL_BEGIN|get_kernel_function_info|dpct
 // DPCT_DEPENDENCY_BEGIN
 // Kernel|kernel_function_info
 // Device|dev_mgr
 // Device|dev_mgr_current_device
 // DPCT_DEPENDENCY_END
 // DPCT_CODE
+static void get_kernel_function_info(kernel_function_info *kernel_info,
+                                     const kernel_function &function) {
```
AndyCHHuang: What's the purpose of this overloading?
Since "operator void *()" of kernel_function is defined, we should be able to leverage
"static void get_kernel_function_info(kernel_function_info *kernel_info, const void *function)"

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
 }
 // DPCT_LABEL_END
 
+namespace detail {
+
+static std::vector<char> read_file_to_vector(const std::string &name) {
```
AndyCHHuang: Why can't we use std::string as the return type?

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  return(ret);
+}
+
+static uint64_t get_lib_size(char const *const blob) {
+#ifdef _WIN32
+  // See https://en.wikipedia.org/wiki/Portable_Executable
+  // and https://learn.microsoft.com/en-us/windows/win32/debug/pe-format
+  // to understand the layout of a DLL header
+
+  ///////////////////////////////////////////////////////////////////////
```
AndyCHHuang, line 241: Please align the comment format with other helper functions.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  ret |= static_cast<uint64_t>(ptr[5]) << 40;
+  ret |= static_cast<uint64_t>(ptr[6]) << 48;
+  ret |= static_cast<uint64_t>(ptr[7]) << 56;
+
+  return(ret);
+}
+
+static uint64_t get_lib_size(char const *const blob) {
+#ifdef _WIN32
+  // See https://en.wikipedia.org/wiki/Portable_Executable
```
AndyCHHuang: "Please reference https://en.wikipedia.org/wiki/Portable_Executable and https://learn.microsoft.com/en-us/windows/win32/debug/pe-format for the detail design of this function."

clang/runtime/dpct-rt/include/kernel.hpp.inc
AndyCHHuang: I think this class is identical to "void *".
I think we can use "void *" directly or using typedef if we really need to create a new name for "void *"

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  return(ptr);
+}
+private:
+void *ptr;
+};
+
+namespace detail {
+
+static std::string mkfilename()
+{
```
AndyCHHuang: 1. Why do we need to covert rawtime to string before hash? Can we hash directly with rawtime?
2. Please using c++ style string operation instead of sprintf() if needed.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  int i = hasher(str);
+
+  sprintf(buffer,"%08X",i);
+
+  return std::filesystem::temp_directory_path().string()
+#ifndef _WIN32
+         + "/"
+#endif
+         + "dpct"
+         + buffer
```
AndyCHHuang: Don't we need a ".so" for file ext in Linux?

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  write_vector_to_file(name, blob);
+#ifdef _WIN32
+  void *so = LoadLibraryA(name.c_str());
+#else
+  void *so = dlopen(name.c_str(), RTLD_LAZY);
+#endif
+  if (so == nullptr)
+    throw std::runtime_error("Failed to load module");
+
+  std::remove(name.c_str());
```
AndyCHHuang: Why do we need to explicitly remove "name" here?
I think it will be destruct after function return.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+
+  return so;
+}
+
+} // namespace detail
+
+static module load_sycl_lib(const std::string &name) {
```
AndyCHHuang: If the module file is exist, we should be able to load it directly.
Writing it to a new file(in load_dl_from_vector()) is not necessary here.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+#endif
+         + "dpct"
+         + buffer
+#ifdef _WIN32
+         + ".dll"
+#endif
+         ;
+}
+
+static module load_dl_from_vector(const std::vector<char> &blob) {
```
AndyCHHuang: Can we pass "const char *" and "size_t" into this function instead of "const std::vector<char>"?
Which will make the implementation of load_sycl_lib_mem() simpler.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+
+static module load_sycl_lib(const std::string &name) {
+  const auto blob = detail::read_file_to_vector(name);
+  return detail::load_dl_from_vector(blob);
+};
+
+static module load_sycl_lib_mem(char const *const image) {
+  const size_t size = detail::get_lib_size(image);
+
+  std::vector<char> blob(size);
```
AndyCHHuang: Redundant copy here, please consider change the function signature of load_dl_from_vector() or add a new overloading.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  std::memcpy(blob.data(), image, size);
+  return detail::load_dl_from_vector(blob);
+};
+
+static void unload_sycl_lib(module &module) {
+  dlclose(module.get_ptr());
```
AndyCHHuang: Please add "FreeLibrary" for windows.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +126,237 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+  if (fn == nullptr)
+    throw std::runtime_error("Failed to get function");
+  return fn;
+}
+
+static void invoke_kernel_function(dpct::kernel_function &function, sycl::queue &queue,
+                                   sycl::range<3> a,
+                                   sycl::range<3> b,
+                                   unsigned int localMemSize, void **kernelParams, void **extra) {
+  function(queue,
```
AndyCHHuang: Please trigger a code formatter for all the added code.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -62,13 +71,44 @@ struct kernel_function_info {
...
 // DPCT_CODE
+static void get_kernel_function_info(kernel_function_info *kernel_info,
+                                     const kernel_function &function) {
+  kernel_info->max_work_group_size =
+      dpct::dev_mgr::instance()
+          .current_device()
+          .get_info<sycl::info::device::max_work_group_size>();
+}
+static kernel_function_info get_kernel_function_info(const kernel_function &function) {
+  kernel_function_info kernel_info;
```
AndyCHHuang: Similar problem to above function.
"static kernel_function_info get_kernel_function_info(const void *function)" should work.

AndyCHHuang (changes requested):

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: done

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Removed unnecessary overloading.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: dlopen does not reload if the file contents change but the filename remains the same.  We write to a new filename to ensure that the file is loaded.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: The remove will delete the temporary file.  destructors will not delete the file.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: dlopen does not require a .so suffix.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: References to external web-sites have been deleted.   We do not want to refer to external sites and this information should be common knowledge.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: References to external web-sites have been deleted. We do not want to refer to external sites and this information should be common knowledge.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Extra overloading removed.  void *function variants are sufficient.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: 1. Using a string produces a more unpredictable hash.
2. sprintf has been removed

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Updated to use const char * and size.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Function is no longer needed.

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: done

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: memcpy removed

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +111,239 @@ static kernel_function_info get_kernel_function_info(const void *function) {
...
+} // namespace detail
+
+static module load_sycl_lib(const std::string &name) {
+  std::ifstream ifs;
+  ifs.open(name, std::ios::in | std::ios::binary);
+
+  std::stringstream buffer;
+  buffer << ifs.rdbuf();
+  return detail::load_dl_from_data(buffer.str().c_str(), buffer.str().size());
+};
```
jzc, line 405: `buffer.str()` returns a copy of the `buffer`'s underlying string, so the buffer is copied an extra time during `buffer.str().size()`. Also, since it's copied, care for dangling pointer from `.c_str()` (fine now since the temporary `buffer.str()` is destructed after `load_dl_from_data`).

LU-JOHN: Fixed.

YuriPlyakhin (approved):

jzc (approved):

AndyCHHuang (approved):

PR #356, opened 2022-11-14
[SYCLomatic] Fix kernel wrapper generation to use the correct types

jzc:


ziranzha:
First of all, for case like:
```
extern` "C" __global__ void foo(int a = 0) {
  __shared__ int b[10];
}
```
The generated code is incorrect since the extra arg of `int b` will be inserted before `int a`. The index passed to 'dpct::get_nth_parameter_t' is incorrect.
Second, we can get the type name easily from the prototype of the original kernel function. The helper function 'dpct::get_nth_parameter_t' is redundant.

jzc:
@ziranzha The types from `getParametersInfo` are unmigrated types, the template was added to cover cases like 
```
extern "C" void foo(float2 *x, ulong1 *y) {}
```

ziranzha:
I think it is a bug of `getParametersInfo`.
We should fix the implementation of `getParametersInfo` instead of adding new helper functions.

clang/test/dpct/module_kernel.cu
```diff
@@ -23,30 +23,42 @@ __constant__ unsigned int const_data[3] = {1, 2, 3};
...
-//CHECK-NEXT:      nr,
-//CHECK-NEXT:      [=](sycl::nd_item<3> item_ct1) {
-//CHECK-NEXT:        foo(k, y, item_ct1, dpct_local_acc_ct1.get_pointer(), const_data_ptr_ct1);
-//CHECK-NEXT:      });
-//CHECK-NEXT:  });
-//CHECK-NEXT: }
-//CHECK-NEXT: }
+// CHECK:      extern "C" {
+// CHECK-NEXT:   void foo_wrapper(sycl::queue &queue, const sycl::nd_range<3> &nr, unsigned int localMemSize, void **kernelParams, void **extra) {
+// CHECK-NEXT:     using foo_type = decltype(foo);
```
0x12CC: Will these generated names conflict if a kernel has the same name as one of its parameters?

AndyCHHuang: Do you mean the original kernel function name is queue, nr, localMemSize, kernelParams or extra?
It's possible, we might consider adding prefix or postfix to the parameter names.

YuriPlyakhin (approved):

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -86,5 +86,15 @@ static kernel_function_info get_kernel_function_info(const void *function) {
 }
 // DPCT_LABEL_END
 
+// DPCT_LABEL_BEGIN|dpct_export_define|dpct
+// DPCT_DEPENDENCY_EMPTY
+// DPCT_CODE
+#ifdef _WIN32
+#define DPCT_EXPORT __declspec(dllexport)
```
LU-JOHN: Could we move DPCT_EXPORT to util.hpp so that we only have one DPCT_LABEL_BEGIN in this commit?  I don't think having such fine-grained header analysis is beneficial (i.e. args_selector and dpct_export_define will always or almost always be used together).  Unifying these two will simplify this PR.

clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2366,45 +2366,36 @@ inline void DeviceFunctionDeclInModule::insertWrapper() {
...
-          std::string ExtraOffsetStr = "sizeof(void*)";
-          for (auto It = getParametersInfo().begin();
-               It != getParametersInfo().end(); It++) {
-            Printer.line(It->second + " = (" + It->first + ")(extra + " +
-                         ExtraOffsetStr + ");");
-            ExtraOffsetStr += " + sizeof(" + It->first + ")";
-          }
-        }
-        Printer.line("}");
+	requestFeature(HelperFeatureEnum::Util_args_selector, FilePath);
```
LU-JOHN: Should this requestFeature line be moved before line 2369 that generates DPCT_EXPORT?  Perhaps Util_args_selector should be renamed to Util_kernel_wrapper, since it is not just for argument selection.

jzc: That makes sense, I implemented that.

LU-JOHN (approved):
LGTM

clang/runtime/dpct-rt/include/util.hpp.inc
```diff
@@ -722,6 +722,106 @@ inline queue_ptr int_as_queue_ptr(uintptr_t x) {
 }
 // DPCT_LABEL_END
 
+// DPCT_LABEL_BEGIN|kernel_wrapper|dpct
+// DPCT_DEPENDENCY_EMPTY
+// DPCT_CODE
+template <int n_nondefault_params, int n_default_params, typename T>
```
AndyCHHuang, line 728: Please reference the designe of helper functions for kernel_wrapper

clang/test/dpct/module_kernel.cu
```diff
@@ -23,36 +22,51 @@ __constant__ unsigned int const_data[3] = {1, 2, 3};
...
-//CHECK-NEXT:      [=](sycl::nd_item<3> item_ct1) {
-//CHECK-NEXT:        foo(k, y, item_ct1, dpct_local_acc_ct1.get_pointer(), const_data_ptr_ct1);
-//CHECK-NEXT:      });
-//CHECK-NEXT:  });
-//CHECK-NEXT: }
-//CHECK-NEXT: }
+
+// CHECK:      extern "C" {
+// CHECK-NEXT:   DPCT_EXPORT void foo_wrapper(sycl::queue &queue, const sycl::nd_range<3> &nr, unsigned int localMemSize, void **kernelParams, void **extra) {
+// CHECK-NEXT:     dpct::args_selector<2, 0, decltype(foo)> selector(kernelParams, extra);
```
AndyCHHuang, line 29: `<2, 0, decltype(foo)>` is difficult to understand. Please reference the design of kernel wrapper.

AndyCHHuang (changes requested):

clang/test/dpct/helper_files_ref/include/util.hpp
```diff
@@ -534,6 +534,102 @@ inline queue_ptr int_as_queue_ptr(uintptr_t x) {
   : reinterpret_cast<queue_ptr>(x);
 }
 
+template <int n_nondefault_params, int n_default_params, typename T>
+class args_selector;
+
+/// args_selector is a helper class for extracting arguments from an
+/// array of pointers to arguments or buffer of arguments to pass to a
+/// kernel function.
+template <int n_nondefault_params, int n_default_params,
```
YuriPlyakhin, line 558: Please, add documentation for template parameters, what each parameter means

clang/test/dpct/module_kernel.cu
YuriPlyakhin: May be mitigate by adding a comment:

// 2 non-default parameters, 0 default parameters
dpct::args_selector<2, 0, decltype(foo)> selector(kernelParams, extra);

clang/test/dpct/helper_files_ref/include/util.hpp
```diff
@@ -534,6 +534,117 @@ inline queue_ptr int_as_queue_ptr(uintptr_t x) {
...
+template <int n_nondefault_params, int n_default_params, typename T>
+class args_selector;
+
+/// args_selector is a helper class for extracting arguments from an
+/// array of pointers to arguments or buffer of arguments to pass to a
+/// kernel function.
+///
+/// \param R(Ts...) The type of the kernel
+/// \param n_nondefault_params The number of nondefault parameters of the kernel
+/// (excluding parameters that like sycl::nd_item, etc.)
```
YuriPlyakhin, line 546: This statement is ambiguous. Could you define exactly which parameters should be excluded?

AndyCHHuang (approved):

PR #357, opened 2022-11-15
[WIP] Impl a framework to migrate math functions

zhiweij1:


clang/test/dpct/math_functions_test.cu
```diff
@@ -68,10 +68,10 @@ float test_sqrt(float in) { return sqrt(in); }
 // CHECK: float test_pow(float a, float b) { return pow(a, b); }
 float test_pow(float a, float b) { return pow(a, b); }
 
-// CHECK: float test_pow(float a, int b) { return sycl::pown(a, b); } 
+// CHECK: float test_pow(float a, int b) { return pow(a, b); }
```
zhimingwang36, line 71: if --optimize-migration is speicified, consider to migrate to sycl::pown()

zhiweij1: It is a pure host API and its declaration is not from cuda header, so I think keep the original code is better.

clang/lib/DPCT/APINamesMathRewrite.inc
```diff
@@ -6,28 +6,23 @@
...
+    MATH_API_REWRITER_HOST_2(
+        "rsqrt",
+        CALL_FACTORY_ENTRY(
+            "rsqrt", CALL(MapNames::getClNamespace(false, true) + "rsqrt",
+                          ARG(0)))),
+    MATH_API_REWRITER_DEVICE_2(
+        "rsqrt",
+        {
+            CALL_FACTORY_ENTRY(
+                "rsqrt", CALL(MapNames::getClNamespace(false, true) + "rsqrt",
```
zhiweij1, line 19: sycl::rsqrt<double> ?

PR #366, opened 2022-11-22
[SYCLomatic] Remove volatile from vector method

0x12CC:

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -3598,6 +3598,34 @@ void VectorTypeMemberAccessRule::runRule(
   }
 
   if (auto ME = getNodeAsType<MemberExpr>(Result, "DerivedVecMemberExpr")) {
+    const auto *MD = DpctGlobalInfo::findAncestor<CXXMethodDecl>(ME);
+    if (MD && MD->isVolatile()) {
+      const auto BaseType =
+          ME->getBase()->getBestDynamicClassTypeExpr()->getType();
+      const bool BaseAndThisSameType =
+          BaseType->getCanonicalTypeUnqualified() ==
+          MD->getThisType()->getCanonicalTypeUnqualified();
+      const bool BaseIsVolatile =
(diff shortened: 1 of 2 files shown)


LU-JOHN (approved):
LGTM

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -3598,6 +3598,20 @@ void VectorTypeMemberAccessRule::runRule(
   }
 
   if (auto ME = getNodeAsType<MemberExpr>(Result, "DerivedVecMemberExpr")) {
+    const auto *MD = DpctGlobalInfo::findAncestor<CXXMethodDecl>(ME);
+    if (MD && MD->isVolatile() && ME->isImplicitAccess()) {
```
ziranzha: Will case like `this->x` be handled?

0x12CC: I've updated the migration to handle `this->x` and added more test cases.

keryell:
Just curious: why this SYCL implementation does not support `volatile`?

0x12CC:
> Just curious: why this SYCL implementation does not support `volatile`?

This SYCL specification doesn't include volatile-qualifed declarations for these methods. The implementation does not include them to match the specification.

keryell:
Sure, the `volatile` keyword does not appear in the specification. But it does not prevent to ignore or implement it to do nothing more.
But I am curious also about the use case in some CUDA applications at the first place.

jzc:
If you wanted to access the components of a vector type in a volatile way in SYCL, how about 
```
*const_cast<volatile int*>(&const_cast<Derived*>(this)->x))
```
i.e. after removing volatility to be able to call the accessor, dereference a pointer to the component with volatility added? At least for CPU code, this seems to work (https://godbolt.org/z/axnfx65dd) - when `foo` is given a nonvolatile argument, the read is optimized out, while with a volatile argument, the load is still there.

jzc (approved):
LGTM though, I think its fine to remove the volatility if we emit a warning.

PR #367, opened 2022-11-23
[SYCLomatic] Add --change-filename-extension flag.

tangjj11:

### DIFF
--- a/clang/include/clang/DPCT/DPCTOptions.inc
+++ b/clang/include/clang/DPCT/DPCTOptions.inc
@@ -397,4 +397,12 @@ DPCT_NON_ENUM_OPTION(DPCT_OPT_TYPE(static llvm::cl::opt<std::string>), AnalysisS
              "Default: the value of --in-root."),
         llvm::cl::value_desc("dir"), llvm::cl::cat(DPCTCat))
 
+DPCT_NON_ENUM_OPTION(
+    DPCT_OPT_TYPE(static llvm::cl::opt<std::string>), ChangeExtension,
+    "change-filename-extension",
+    llvm::cl::desc("Specifies which extensions of migrated files will be changed, "
+                   "split by comma. Default:\n"
+                   ".cu, .cuh, .cpp, .cxx, .cc, .c, .C"),
+    llvm::cl::value_desc("file"), llvm::cl::cat(DPCTCat), llvm::cl::Optional)
+
 #endif // !DPCT_OPTIONS_IN_CLANG_DPCT
--- a/clang/lib/DPCT/DPCT.cpp
+++ b/clang/lib/DPCT/DPCT.cpp
@@ -656,6 +656,27 @@ int runDPCT(int argc, const char **argv) {
     ApiMappingEntry::printMappingDesc(llvm::outs(), QueryApiMapping);
     dpctExit(MigrationSucceeded);
   }
+
+  auto ExtensionStr = ChangeExtension.getValue();
+  ExtensionStr.erase(std::remove(ExtensionStr.begin(), ExtensionStr.end(), ' '),
+                     ExtensionStr.end());
+  auto Extensions = split(ExtensionStr, ',');
+  for (auto &Extension : Extensions) {
+    bool Legal = true;
+    const auto len = Extension.length();
+    if (len < 2 || len > 5 || Extension[0] != '.') {
+      ShowStatus(MigrationErrorInvalidChangeFilenameExtension);
+      dpctExit(MigrationErrorInvalidChangeFilenameExtension, false);
+    }
+    for (size_t i = 1; i < len; ++i) {
+      if (!std::isalpha(Extension[i])) {
+        ShowStatus(MigrationErrorInvalidChangeFilenameExtension);
+        dpctExit(MigrationErrorInvalidChangeFilenameExtension, false);
+      }
+    }
+    DpctGlobalInfo::addChangeExtensions(Extension);
+  }
+
   if (InRoot.empty() && ProcessAllFlag) {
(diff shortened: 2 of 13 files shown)


clang/include/clang/DPCT/DPCTOptions.inc
```diff
@@ -399,4 +399,12 @@ DPCT_NON_ENUM_OPTION(DPCT_OPT_TYPE(static llvm::cl::opt<std::string>), AnalysisS
              "Default: the value of --in-root."),
         llvm::cl::value_desc("dir"), llvm::cl::cat(DPCTCat))
 
+DPCT_NON_ENUM_OPTION(
+    DPCT_OPT_TYPE(static llvm::cl::opt<std::string>), ChangeExtension,
+    "change-filename-extension",
+    llvm::cl::desc("Specifies which extension of filename will be changed, "
```
YuriPlyakhin:
```suggestion
    llvm::cl::desc("Specifies which extensions of migrated files will be changed, "
```

YuriPlyakhin (changes requested):

clang/lib/DPCT/DPCT.cpp
```diff
@@ -656,6 +656,19 @@ int runDPCT(int argc, const char **argv) {
     ApiMappingEntry::printMappingDesc(llvm::outs(), QueryApiMapping);
     dpctExit(MigrationSucceeded);
   }
+
+  auto ExtensionStr = ChangeExtension.getValue();
+  ExtensionStr.erase(std::remove(ExtensionStr.begin(), ExtensionStr.end(), ' '),
+                     ExtensionStr.end());
+  auto Extensions = split(ExtensionStr, ',');
+  for (auto &Extension : Extensions)
```
zhimingwang36: here need enhance error handleing on illegal input:  -change-filename-extension=.*,  and so on.  
Please define which is legal, and treat all other as illeagal.

clang/lib/DPCT/DPCT.cpp
```diff
@@ -656,6 +656,19 @@ int runDPCT(int argc, const char **argv) {
...
+    DpctGlobalInfo::addChangeExtensions(Extension);
+  if (DpctGlobalInfo::getChangeExtensions().empty() &&
+      ChangeExtension.getNumOccurrences() > 0)
+    // If user set an empty value to this flag, just insert a blank extension to
+    // indicate the flag is set.
+    DpctGlobalInfo::addChangeExtensions(" ");
```
zhimingwang36:
```suggestion

```

clang/lib/DPCT/SaveNewFiles.cpp
```diff
@@ -174,15 +174,18 @@ bool rewriteDir(SmallString<512> &FilePath, const StringRef InRoot,
...
-    path::replace_extension(FileName, "dp.cpp");
-  } else if (FileType & SPT_CppSource) {
-    auto Extension = path::extension(FileName);
-    path::replace_extension(FileName, Extension + ".dp.cpp");
-  } else if (FileType & SPT_CudaHeader) {
-    path::replace_extension(FileName, "dp.hpp");
+  // If user does not specify which extension need be changed, we change all the
+  // SPT_CudaSource, SPT_CppSource and SPT_CudaHeader files.
+  if (DpctGlobalInfo::getChangeExtensions().empty() ||
+      DpctGlobalInfo::getChangeExtensions().count(Extension.str())) {
```
zhimingwang36, line 182: eg. when user want to change:  a.xyz ==> a.xyz.dp.cpp,  it will not happen accord to current logic,  right? (As .xyz is not in SPT_*  defines.

tangjj11: No, here is the logic in function `GetSourceFileType`:
```c++
    // clang-format off
    // For unknown file extensions, determine the file type according to:
    // A. If it shows up in the compilation database as single migration
    //    file, then treat it as a main source file.
    // B. If it is included by another source file, then treat it as a header
    //    file.
    // C. If both A and B hold, then default to A.
    // clang-format on
    auto &FileSetInDB = dpct::DpctGlobalInfo::getFileSetInCompiationDB();
    if (FileSetInDB.find(SourcePath.str()) != end(FileSetInDB)) {
      return SPT_CppSource;
    }
```
it will also return SPT_CppSource when the file is in database.

clang/include/clang/DPCT/DPCTOptions.inc
tangjj11: fixed

clang/lib/DPCT/DPCT.cpp
tangjj11: fixed

clang/lib/DPCT/DPCT.cpp
tangjj11: fixed

zhiweij1 (approved):

YuriPlyakhin (approved):

clang/lib/DPCT/Error.cpp
```diff
@@ -38,6 +38,12 @@ void ShowStatus(int Status, std::string Message) {
     StatusString = "Error: Path for CUDA header files specified by "
                    "--cuda-include-path is invalid.";
     break;
+  case MigrationErrorInvalidChangeFilenameExtension:
+    StatusString =
+        "Error: Extension specified by --change-filename-extension is "
+        "invalid.\n"
+        "Only a '.' followed by alphabetical characters and digits is allowed.";
```
zhimingwang36: a little update:  digits is not allowed according to DPCT.cpp line L672 code logic.

tangjj11: fixed

PR #373, opened 2022-11-24
[SYCLomatic] Support migration of 16 cusolver APIs

zhiweij1:


clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +93,240 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+/// \param [in,out] b The Hermitian matrix B.
+/// \param [in] ldb The leading dimension of matrix B.
+/// \param [in] w Eigenvalues.
+/// \param [in] scratchpad Scratchpad memory to be used by the routine
+/// for storing intermediate results.
+/// \param [in] scratchpad_size Size of scratchpad memory as a number of
+/// floating point elements of type T.
+/// \param [out] info If lapack synchronous exception is caught, the value
+/// returned from info() method of the exception is set to \p info. If other
+/// SYCL synchronous exception is caught, -1 is set to \p info.
```
zhimingwang36:
```suggestion
```

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +93,240 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+/// for complex data) positive-definite matrices.
+/// \return Returns 0 if no synchronous exception, otherwise returns 1.
+/// \param [in] queue Device queue where calculations will be performed.
+/// \param [in] uplo Must be uplo::upper or uplo::lower.
+/// \param [in] n The order of the matrix A.
+/// \param [in,out] a Array of pointers to matrix A.
+/// \param [in] lda The leading dimension of matrix A.
+/// \param [out] info If lapack synchronous exception is caught, the value
+/// returned from info() method of the exception is set to \p info. If other
+/// SYCL synchronous exception is caught, -1 is set to \p info.
```
zhimingwang36:
```suggestion
```

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +93,240 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+/// Computes the Cholesky factorizations of a batch of symmetric (or Hermitian,
+/// for complex data) positive-definite matrices.
+/// \return Returns 0 if no synchronous exception, otherwise returns 1.
+/// \param [in] queue Device queue where calculations will be performed.
+/// \param [in] uplo Must be uplo::upper or uplo::lower.
+/// \param [in] n The order of the matrix A.
+/// \param [in,out] a Array of pointers to matrix A.
+/// \param [in] lda The leading dimension of matrix A.
+/// \param [out] info If lapack synchronous exception is caught, the value
+/// returned from info() method of the exception is set to \p info. If other
```
zhimingwang36:
```suggestion
/// returned from info() method of the exception is set to \p info. 
```

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +93,240 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+  } catch (oneapi::mkl::lapack::exception const &e) {
+    std::cerr << "Unexpected exception caught during call to LAPACK API: "
+                 "potrf_batch_scratchpad_size/potrf_batch"
+              << std::endl
+              << "reason: " << e.what() << std::endl
+              << "info: " << e.info() << std::endl;
+    std::vector<int> info_vec(group_size, static_cast<int>(e.info()));
+    queue.memcpy(info, info_vec.data(), group_size * sizeof(int)).wait();
+    if (scratchpad)
+      sycl::free(scratchpad, queue);
```
zhimingwang36, line 237: In the two exception path,  it need call: 

>  std::free(matrix_info);

zhimingwang36 (approved):
LGTM

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+  queue.submit([&, info_val](sycl::handler &cgh) {
+    auto info_acc = info_buf.get_access<sycl::access_mode::write>(cgh);
+    cgh.single_task<dpct_kernel_name<class hegvd_set_info, T>>(
+        [=]() { info_acc[0] = info_val; });
+  });
+  return ret_val;
+#else
+  try {
+    oneapi::mkl::lapack::hegvd(queue, itype, jobz, uplo, n, (Ty *)a, lda, (Ty *)b,
+                               ldb, w, (Ty *)scratchpad, scratchpad_size);
```
Alexander-Kleymenov, line 199: here should be `wait_and_throw` called on the returned event.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+  } catch (oneapi::mkl::lapack::exception const& e) {
+    std::cerr << "Unexpected exception caught during call to LAPACK API: hegvd"
```
Alexander-Kleymenov, line 201: Note, this trap will catch only synchronous exceptions, risen during call to hegvd, in common case, calculation errors happened during GPU-side workload are risen as asynchronous exceptions, and delivered via asyncronous error_handler provided to the queue's constructor.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+    int info_val = static_cast<int>(e.info());
+    queue.memcpy(info, &info_val, sizeof(int)).wait();
+    return 1;
+  } catch (sycl::exception const& e) {
+    std::cerr << "Caught synchronous SYCL exception:" << std::endl
+              << "reason: " << e.what() << std::endl;
+    queue.memset(info, 0, sizeof(int)).wait();
+    return 1;
+  }
+  queue.memset(info, 0, sizeof(int));
```
Alexander-Kleymenov, line 214: `wait()` should be called on the event returned from memset.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+    scratchpad_size = oneapi::mkl::lapack::potrf_batch_scratchpad_size<Ty>(
+        queue, &(matrix_info->uplo_info), &(matrix_info->n_info),
+        &(matrix_info->lda_info), 1, &(matrix_info->group_size_info));
+    scratchpad = sycl::malloc_device<Ty>(scratchpad_size, queue);
+    e = oneapi::mkl::lapack::potrf_batch(
+        queue, &(matrix_info->uplo_info), &(matrix_info->n_info), (Ty **)a,
+        &(matrix_info->lda_info), 1, &(matrix_info->group_size_info),
+        scratchpad, scratchpad_size);
+  } catch (oneapi::mkl::lapack::exception const &e) {
+    std::cerr << "Unexpected exception caught during call to LAPACK API: "
```
Alexander-Kleymenov, line 267: Same comment as above applies here and to other synchronous exceptions' traps.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+                 "potrf_batch_scratchpad_size/potrf_batch"
+              << std::endl
+              << "reason: " << e.what() << std::endl
+              << "info: " << e.info() << std::endl;
+    std::vector<int> info_vec(group_size, static_cast<int>(e.info()));
```
Alexander-Kleymenov: Not all of the problems from the batch will get non-zero status info code. There is a `potrf_batch` documentation describing `batch_error` (mistakenly called batch_exceltion in the documentation): https://www.intel.com/content/www/us/en/develop/documentation/oneapi-mkl-dpcpp-developer-reference/top/lapack-routines/potrf_batch-group-version.html

> This exception is thrown when problems occur during calculations. You can obtain the info code of the problem using the info() method of the exception object:
> If info = -n, the n-th parameter had an illegal value.
> If info equals the value passed as scratchpad size, and detail() returns non-zero, then the passed scratchpad is of insufficient size, and the required size should be not less then value returned by the detail() method of the exception object.
> If info is zero, then the diagonal element of some of Ui is zero, and the solve could not be completed. The indexes of such matrices in the batch can be obtained with the ids() method of the exception object. You can obtain the indexes of the first zero diagonal elements in these Ui matrices using the infos() method of the exception object.

I wish we had an example demonstrating the way to handle this, but you can check how it is done in `potrf_batch_buffer.cpp` unit test: 
https://github.com/intel-innersource/libraries.performance.math.mkl/blob/develop/qa/testbase/tbmkl/tests/lapack/unit/dpcpp/potrf_batch_buffer/potrf_batch_buffer.cpp
Please note, as per docs, `batch_error` has `info()` equaling zero, to distinguish computation errors, and only `batch_error.exceptions()` will have non-zero `info()` for corresponding problem in the batch.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: Please note, `wait_and_throw()` should be called on the event object in order to deliver any async exceptions risen during execution of the GPU kernel.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
```diff
@@ -133,6 +132,260 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+  } catch (sycl::exception const &e) {
+    std::cerr << "Caught synchronous SYCL exception:" << std::endl
+              << "reason: " << e.what() << std::endl;
+    queue.memset(info, 0, group_size * sizeof(int)).wait();
+    std::free(matrix_info);
+    if (scratchpad)
+      sycl::free(scratchpad, queue);
+    return 1;
+  }
+  queue.submit([&](sycl::handler &cgh) {
```
Alexander-Kleymenov, line 300: This won't be necessary, if `e.wait_and_throw()` was called before.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: This is the catch example for the batch_error:
```
    } catch(oneapi::mkl::lapack::batch_error const& be) {
        // Handle LAPACK related exceptions happened during synchronous call
        std::cout << "Unexpected exception caught during synchronous call to LAPACK API:\nReason: " << be.what() << "\nNumber: " << be.info() << std::endl;
        int i = 0;
        auto &ids = be.ids();
        for (auto const& e : be.exceptions()) {
            try {
                std::rethrow_exception(e);
            } catch(oneapi::mkl::lapack::exception &e) {
               std::cout << "Exception " << ids[i++] << " in a batch says:\n" << e.what() << " (info code: " << e.info() << ")" << std::endl;
                info = e.info();
            }
        }
    }
```

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: > Same comment as above applies here and to other synchronous exceptions' traps.

But as for now, due to some issues in compiler, LAPACK does not throw asynchronous exception, only reports calculation status as synchronous exception, so code in this review should be fine.

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +92,241 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+    int info_val = static_cast<int>(e.info());
+    queue.memcpy(info, &info_val, sizeof(int)).wait();
+    return 1;
+  } catch (sycl::exception const& e) {
+    std::cerr << "Caught synchronous SYCL exception:" << std::endl
+              << "reason: " << e.what() << std::endl;
+    queue.memset(info, 0, sizeof(int)).wait();
+    return 1;
+  }
+  queue.memset(info, 0, sizeof(int));
```
Alexander-Kleymenov, line 168: should be done after event returned from hegvd finishes.

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
```diff
@@ -93,6 +92,241 @@ inline int sygvd(sycl::queue &queue, std::int64_t itype, oneapi::mkl::job jobz,
...
+    scratchpad_size = oneapi::mkl::lapack::potrf_batch_scratchpad_size<Ty>(
+        queue, &(matrix_info->uplo_info), &(matrix_info->n_info),
+        &(matrix_info->lda_info), 1, &(matrix_info->group_size_info));
+    scratchpad = sycl::malloc_device<Ty>(scratchpad_size, queue);
+    e = oneapi::mkl::lapack::potrf_batch(
+        queue, &(matrix_info->uplo_info), &(matrix_info->n_info), (Ty **)a,
+        &(matrix_info->lda_info), 1, &(matrix_info->group_size_info),
+        scratchpad, scratchpad_size);
+  } catch (oneapi::mkl::lapack::exception const &e) {
+    std::cerr << "Unexpected exception caught during call to LAPACK API: "
```
Alexander-Kleymenov, line 215: Same comment regarding batch_error applies.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: Below is a general template on how to implement oneMKL LAPACK wrapper catching both synchronous and asynchronous exceptions without imposing additional requirements to the user's queue object (like "queue should be constructed with async error handling supporting these types of exception: ..."), and avoiding code duplication by implementing single catch block for both types of possible exceptions (sync/async):
```c++
#include <CL/sycl.hpp>

// some LAPACK function submitting kernel throwing asynchronous exception
sycl::event lapack_func(sycl::queue &q)
{
    return q.submit([&](auto &h){
            h.host_task([=]{
                throw std::exception();
            });
        });
}

// helper function enhancing arbitrary user provided queue with async error handling ability
sycl::queue get_error_handling_queue(sycl::queue q, std::exception_ptr &exception) {
    auto error_handler = [&] (sycl::exception_list exceptions) {
            exception = *exceptions.begin();
        };

    sycl::device  dev { q.get_device() };
    sycl::context ctx { dev, error_handler };
    return sycl::queue { ctx, dev };
}

// some lapack wrapper concerned with handling both sync/async exceptions
int lapack_wrapper_missed_exception(sycl::queue &q)
{
    // lapack async exception, if thrown, will be stored in this object
    std::exception_ptr lapack_async_exception;
    // enhance user queue with error handler catching any async lapack exception
    sycl::queue que = get_error_handling_queue(q, lapack_async_exception);
    try {
        // sychronous exception, if any, will happen here, and caught by catch block below
        sycl:: event e = lapack_func(que);
        // asyncronous exception, if any, will happen here:
        e.wait_and_throw();
        // rethrow async exception, if any happened, so it will be caught by same catch block below
        if (lapack_async_exception) {
            std::rethrow_exception(lapack_async_exception);
        }
    } catch (const std::exception &e) {
        printf("Caught: %s\n", e.what());
        return 0; // exception was not missed, test passed
    }
    return 1; // exception missed, test failed
}

int main()
{
    sycl::queue q{ sycl::gpu_selector_v };

    if (lapack_wrapper_missed_exception(q)) {
        printf("FAILED\n");
    } else {
        printf("PASSED\n");
    }
}
```

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: Since cusolverDnChegvd is an async API, we do not want this mapping API is a sync API.
So there is no wait/wait_and_throw here.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: Actually, we know this issue. We create all queues with dpct interface in file https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/runtime/dpct-rt/include/device.hpp.inc . You can search word "exception_handler" in that file. And since it is a common interface, we do not want to introduce MKL-related exceptions in this exception_handler.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: We want thin function to be async (for non-error code path)

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: ok, I will fix it.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: We need consider again if it is reasonable to introduce MKL related code in the common exception_handler 
 (in device.hpp). @zhimingwang36 please comment.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: I think it is necessary if we want this API to be async.

clang/test/dpct/helper_files_ref/include/lapack_utils.hpp
zhiweij1: Since we want this API to be async, we cannot add wait in the impl.
So this "info" only contains sync exception related information.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: The way demonstrated here avoids changes to common `exception_handler`: it makes call-site specific handlers, and exception traps process only exceptions specific to any particular function.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: I've updated potrf_batch test. Please check if it helps:
https://github.com/intel-innersource/libraries.performance.math.mkl/blob/d962a39dc3a49c56a9ed5a1656b5fc9d01a07f41/qa/testbase/tbmkl/tests/lapack/unit/dpcpp/potrf_batch_buffer/potrf_batch_buffer.cpp

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: API can throw synchronous or asynchronous exceptions. This code handles synchronous, but if there was asynchronous exception, it will be left without proper updating incoming info device pointer.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
zhiweij1: Yes, but it introduced another queue and make this API to be synchronous.
If we need process the asynchronous exceptions, I think changing `exception_handler` is better than changing queue & synchronization.

zhimingwang36: The cost here to catch the Async exception is to call wait_and_throw(), we don't want to introduce this cost.  So I'm ok to not catch async exception in this help function.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: Ok, I see. If you don't care about particular non-zero status codes, then it's fine. If you need to properly update device side info code, then you will need to wait for the calculations to finish.

clang/runtime/dpct-rt/include/lapack_utils.hpp.inc
Alexander-Kleymenov: Ok, sounds good.

YuriPlyakhin (approved):

zhimingwang36 (approved):
LGTM

PR #374, opened 2022-11-24
[SYCLomatic] Enhance time measurement migration for on-device kernel timing by using SYCL property::queue::enable_profiling

tomflinda:


clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -463,12 +463,15 @@ public:
   }
   device_ext(const sycl::device &base)
       : sycl::device(base), _ctx(*this) {
+    auto property =
+        sycl::property_list{sycl::property::queue::enable_profiling(),
+                            sycl::property::queue::in_order()};
 #ifdef DPCT_USM_LEVEL_NONE
     _queues.push_back(
-        std::make_shared<sycl::queue>(_ctx, base, exception_handler));
+        std::make_shared<sycl::queue>(_ctx, base, exception_handler, property);
```
zhiweij1, line 471:
```suggestion
        std::make_shared<sycl::queue>(_ctx, base, exception_handler, property));
```

zhiweij1 (changes requested):

clang/runtime/dpct-rt/include/device.hpp.inc
zhiweij1:
```suggestion
        std::make_shared<sycl::queue>(_ctx, base, exception_handler, property));
```

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -8724,699 +8658,95 @@ const ValueDecl *getDecl(const Expr *E) {
...
-      CE->getCalleeDecl()->getAsFunction()->getNameAsString();
 
   auto StreamArg = CE->getArg(CE->getNumArgs() - 1);
   auto StreamName = getStmtSpelling(StreamArg);
   auto ArgName = getStmtSpelling(CE->getArg(0));
   bool IsDefaultStream = isDefaultStream(StreamArg);
-  auto IndentLoc = CE->getBeginLoc();
-  auto &Context = dpct::DpctGlobalInfo::getContext();
 
   if (IsAssigned) {
```
tangjj11, line 8676: The only different of the two branch of this `if` is this `StmtStr = "(" + StmtStr + ", 0)";`, so some code can be extract out of the `if.`

zhiweij1 (approved):

clang/test/dpct/driver-stream-and-event-enable-profiling.cu
```diff
@@ -0,0 +1,113 @@
+// RUN: dpct --enable-profiling  -out-root %T/driver-stream-and-event-enable-profiling %s --cuda-include-path="%cuda-path/include"
+// RUN: FileCheck --match-full-lines --input-file %T/driver-stream-and-event-enable-profiling/driver-stream-and-event-enable-profiling.dp.cpp %s
+
+#include<vector>
```
zhimingwang36, line 8: Could you add check to make sure "#define DPCT_PROFILING_ENABLED" is define before device.hpp or dpct.hppis included?

clang/test/dpct/driver-stream-and-event-enable-profiling.cu
```diff
@@ -0,0 +1,113 @@
...
+  size_t length;
+  CUdeviceptr  cuPtr;
+  // CHECK: std::async([&]() {hStream->wait(); callback<char>(hStream, 0, data); });
+  cuStreamAddCallback(hStream, callback<char>, data, flag);
+
+  // CHECK: int result = (std::async([&]() {hStream->wait(); callback<char>(hStream, 0, data); }), 0);
+  CUresult result = cuStreamAddCallback(hStream, callback<char>, data, flag);
+
+  //CHECK: /*
+  //CHECK-NEXT: DPCT1026:{{[0-9]+}}: The call to cuStreamAttachMemAsync was removed because SYCL currently does not support associating USM with a specific queue.
```
zhimingwang36: please make the functionality of the patch focus on the new feature implemented in this PR.

clang/test/dpct/driver-stream-and-event-enable-profiling.cu
tomflinda: OK, will update it in the next patch.

clang/test/dpct/driver-stream-and-event-enable-profiling.cu
tomflinda: OKay

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -463,12 +463,25 @@ public:
   }
   device_ext(const sycl::device &base)
       : sycl::device(base), _ctx(*this) {
+#ifdef DPCT_PROFILING_ENABLED
```
YuriPlyakhin: I think this code can be refactored to be more compact and less duplications.
For example you can define properties first with macro-related logic and then use one call without any macros, instead of repeating the call multiple times under different macros defined.
Also, consider moving it to some function, since there are 3 almost identical changes done - need to avoid duplication like that.

YuriPlyakhin (changes requested):
The option text looks good to me. But the changes in helpers may need improvement.

clang/runtime/dpct-rt/include/device.hpp.inc
tomflinda: OKay

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -724,20 +705,38 @@ public:
...
+        sycl::property_list{sycl::property::queue::enable_profiling()};
+#else
+    auto property =
+        sycl::property_list{sycl::property::queue::enable_profiling(),
+                            sycl::property::queue::in_order()};
+#endif
+#else
+#ifdef DPCT_USM_LEVEL_NONE
+    auto property =
+        sycl::property_list{sycl::property::queue::enable_profiling()};
```
zhimingwang36:
```suggestion
      
```

clang/runtime/dpct-rt/include/device.hpp.inc
```diff
@@ -724,20 +705,38 @@ public:
...
+#else
+    auto property =
+        sycl::property_list{sycl::property::queue::enable_profiling(),
```
zhimingwang36:
```suggestion

```

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -8209,7 +8210,10 @@ void EventAPICallRule::runRule(const MatchFinder::MatchResult &Result) {
     if (getEventQueryTraversal().startFromQuery(CE))
       return;
 
-    if (!isEventElapsedTimeFollowed(CE)) {
+    // Pattern-based solution for migration of time measurement code is enabled
+    // only when option '--enable-profiling' is enabled.
```
zhimingwang36: ### 
```suggestion
    // only when option '--enable-profiling' is disabled.
```

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -8380,149 +8396,340 @@ void EventAPICallRule::findEventAPI(const Stmt *Node, const CallExpr *&Call,
 void EventAPICallRule::handleEventRecord(const CallExpr *CE,
```
zhimingwang36: the function is quite big, how about we have two versoin of handleEventRecord,  one if for: event_profiling is enalbed,  other is the old one.

clang/runtime/dpct-rt/include/device.hpp.inc
tomflinda: Done

clang/runtime/dpct-rt/include/device.hpp.inc
tomflinda: Done

clang/lib/DPCT/ASTTraversal.cpp
tomflinda: Okay

clang/lib/DPCT/ASTTraversal.cpp
tomflinda: Okay

zhimingwang36 (approved):
LGTM

clang/test/dpct/helper_files_ref/include/device.hpp
```diff
@@ -340,24 +321,41 @@ class device_ext : public sycl::device {
     // Guard the destruct of current_queues to make sure the ref count is safe.
     lock.lock();
   }
+  sycl::property_list get_property_list() const {
```
YuriPlyakhin: I see that it is a public method of device_ext class. I suggest to make it private. I did not see any usage of it in the generated code.
Also, I suggest to rename to something like: "get_default_property_list_for_queue" to make the name more specific to what the method does actually.

YuriPlyakhin (changes requested):

clang/test/dpct/helper_files_ref/include/device.hpp
tomflinda: OKay

YuriPlyakhin (approved):
helper headers change LGTM

PR #378, opened 2022-11-28
[SYCLomatic] Refine cudaStream_t/CUstream migration

jzc:


0x12CC (approved):

LU-JOHN (commented):
LGTM except that I was also wondering about the test for x<=2 in int_as_queue_ptr.  If negative values and 1 really map to the default queue, we should add a LIT test to ensure that a call with a stream of -4, -3, or 1 get mapped to the default queue.

LU-JOHN (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #379, opened 2022-11-28
[SYCLomatic] Create pattern rewriting tool

0x12CC:


clang/tools/pattern-rewriter/src/driver.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  const auto RuleSetFile = readFile(RuleSetFilename.getValue());
+  llvm::yaml::Input RuleSetParser(RuleSetFile);
+  std::map<std::string, pattern::Rule> RuleSet;
+  RuleSetParser >> RuleSet;
+
+  const auto InputBuffer = readFile(InputFilename.getValue());
+  const std::string Input(InputBuffer.begin(), InputBuffer.end());
+
+  std::string Output = Input;
```
jzc: The input buffer is copied an extra time here. `std::string Output = std::move(input)` or elide copy with `auto Output = fixLineEndings(readFile(...))`

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  return Indentation;
+}
+
+static std::vector<std::string> split(const std::string &Input,
+                                      char Seperator) {
+  std::vector<std::string> Output;
+  const int Size = Input.size();
+  int Index = 0;
+  int Start = 0;
+  while (Index < Size) {
```
AndyCHHuang, line 68: can we use a string.find() function to enhance the "index++" in the while loop

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  while (End > (Index + 1) && isWhitespace(Input[End])) {
+    End--;
+  }
+
+  return Input.substr(Index, End + 1);
+}
+
+static std::string indent(const std::string Input, int Indentation) {
+  std::vector<std::string> Output;
+  std::stringstream IndentStream;
```
AndyCHHuang: Please consider using the following ctor of std::string:
string (size_t n, char c)

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+  int End = Size - 1;
+  while (End > (Index + 1) && isWhitespace(Input[End])) {
+    End--;
+  }
+
+  return Input.substr(Index, End + 1);
+}
+
+static std::string indent(const std::string Input, int Indentation) {
```
AndyCHHuang: const std::string "&"Input

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  std::vector<std::string> Output;
+  std::stringstream IndentStream;
+  for (int i = 0; i < Indentation; i++) {
+    IndentStream << " ";
+  }
+  const auto Indent = IndentStream.str();
```
AndyCHHuang: If we create a osstream, and keep insert indent and trimmed lines in to the stream in the for loop, we can improve the performance by returning std::string_view by trim() and split().

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+        Index++;
+      }
+      continue;
+    }
+
+    if (Index < (Size - 1) && Character == '$' && Pattern[Index + 1] == '{') {
+      Index += 2;
+
+      std::stringstream NameStream;
+      while (Index < Size && Pattern[Index] != '}') {
```
AndyCHHuang: Pos = Pattern.find("}", Index)
if(Pos != std::string::npos){
  Name = Pattern.substr(index, pos - index);
  index = pos + 1;
  }

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+      Skip = Indentation;
+    }
+    OutputStream << Character;
+    Index++;
+  }
+  return OutputStream.str();
+}
+
+static void adjustSuffixLengths(MatchPattern &Pattern) {
+  int SuffixTerminator = Pattern.size() - 1;
```
AndyCHHuang, line 167: Could you provide some comments to explain the design of this function?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+static int parseCodeElement(const MatchPattern &Suffix,
+                            const std::string &Input, const int Start);
+
+static int parseBlock(char LeftDelimeter, char RightDelimeter,
+                      const std::string &Input, const int Start) {
+  const int Size = Input.size();
+  int Index = Start;
+
+  if (Index >= Size || Input[Index] != LeftDelimeter) {
```
AndyCHHuang: Since the LeftDelimeter should be known while invoking this function, why we need this check here?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+    if (Character == '(') {
+      Index = parseBlock('(', ')', Input, Index);
+      continue;
+    }
+
+    if (isRightDelimeter(Input[Index])) {
+      break;
+    }
+
+    if (Character == '\'') {
```
AndyCHHuang, line 307: Please add comment to explain why we need to skip all the contents inside "" and ''

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+    if (Character == '"') {
+      Index++;
+      while (Index < Size &&
+             !(Input[Index - 1] != '\\' && Input[Index] == '"')) {
+        Index++;
+      }
+      if (Index >= Size) {
+        return -1;
+      }
+      Index++;
```
AndyCHHuang, line 329: Why is the process of "" different from '' at these 3 lines?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+  int Index = Start;
+  int PatternIndex = 0;
+  const int PatternSize = Pattern.size();
+  const int Size = Input.size();
+
+  while (PatternIndex < PatternSize && Index < Size) {
+    const auto &Element = Pattern[PatternIndex];
+
+    if (std::holds_alternative<SpacingElement>(Element)) {
```
AndyCHHuang, line 376: for a sequence of whitespaces in the input, will every whitespace checked once?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+      const auto &Code = std::get<CodeElement>(Element);
+      MatchPattern Suffix(Pattern.begin() + PatternIndex + 1,
+                          Pattern.begin() + PatternIndex + 1 +
+                              Code.SuffixLength);
+
+      int Next = parseCodeElement(Suffix, Input, Index);
+      if (Next == -1) {
+        return {};
+      }
+      const int Indentation = detectIndentation(Input, Index);
```
AndyCHHuang, line 407: why do we need to detect indent for a CodeElement?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+        }
+      } else {
+        Result.Bindings[Code.Name] = std::move(ElementContents);
+      }
+      Index = Next;
+      PatternIndex++;
+      continue;
+    }
+
+    PatternIndex++;
```
AndyCHHuang: In which condition will this line be executed?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+}
+
+static void instantiateTemplate(
+    const std::string &Template,
+    const std::unordered_map<std::string, std::string> &Bindings,
+    const int Indentation, std::ostream &OutputStream) {
+  const int Size = Template.size();
+  int Index = 0;
+
+  while (Index < Size && isWhitespace(Template[Index])) {
```
AndyCHHuang, line 438: Can we leverage trim()?

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  while (Index <= End) {
+    const auto Character = Template[Index];
+
+    if (Index < (Size - 1) && Character == '$' && Template[Index + 1] == '{') {
+      const int BindingStart = Index;
+      Index += 2;
+
+      std::stringstream NameStream;
+      while (Index < Size && Template[Index] != '}') {
+        NameStream << Template[Index];
```
AndyCHHuang: std::string::find()

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+      }
+
+      if (Index == Size) {
+        throw std::runtime_error("Invalid rewrite pattern expression");
+      }
+      Index++;
+
+      std::string Name = NameStream.str();
+      if (Bindings.count(Name) > 0) {
+        const int BindingIndentation =
```
AndyCHHuang, line 463: maybe we can store original lines as a vector to simplify the calculation here

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+            detectIndentation(Template, BindingStart) + Indentation;
+        const std::string Contents =
+            indent(Bindings.at(Name), BindingIndentation);
+        OutputStream << Contents;
+      }
+      continue;
+    }
+
+    OutputStream << Character;
+    if (Character == '\n') {
```
AndyCHHuang, line 473: consider prepare a std::string(" ", Indentation) in the function begining and call osstream.write(const char *, size_t)

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  const int Size = Input.size();
+
+  int Index = 0;
+  while (Index < Size) {
+    auto Result = findMatch(Pattern, Input, Index);
+
+    if (Result.has_value()) {
+      auto &Match = Result.value();
+      for (const auto &[Name, Value] : Match.Bindings) {
+        if (Rule.Subrules.count(Name)) {
```
AndyCHHuang: the combination of count() + at() can be replaced by find()
can save one of the search time

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+  }
+  const auto Indent = IndentStream.str();
+  const auto Lines = split(Input, '\n');
+  for (const auto &Line : Lines) {
+    Output.push_back((trim(Line).size() > 0) ? (Indent + Line) : "");
+  }
+  return trim(join(Output, "\n"));
+}
+
+static std::string dedent(const std::string &Input, int Indentation) {
```
AndyCHHuang, line 125: if we save all lines in a vector, indent() and dedent() could be much simpler.

AndyCHHuang (changes requested):

clang/tools/pattern-rewriter/src/pattern.cpp
```diff
@@ -0,0 +1,497 @@
...
+    if (Input[Index] == '\n' || Index == 0) {
+      Indentation = Start - Index - 1;
+      break;
+    }
+    Index--;
+  }
+  return Indentation;
+}
+
+static std::vector<std::string> split(const std::string &Input,
```
LU-JOHN, line 63: Seperator should be Separator

clang/tools/pattern-rewriter/src/pattern.cpp
LU-JOHN: Is there a clearer way to detect a line with no-whitespace than using trim(Line).size()>0?

0x12CC: I'm not sure what you mean. I wanted to avoid returning string views to ensure that the return values have value semantics. It's less efficient in certain cases but avoids lifetime issues that cause dangling string views.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: This parser checks the delimiter because it's the expected first character. I also check at the call sites of this function, but I want the parser to work correctly when the outer check is missing. This is useful when you want failure to imply backtracking:
```
// Parse any code block.
Index = std::max({
  parseBlock('{', '}', Input, Index),
  parseBlock('[', ']', Input, Index),
  parseBlock('(', ')', Input, Index)
});
```

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: A sequence of whitespace characters in the match pattern produces a single `SpacingElement`. This `SpacingElement` matches one or more spaces in the input.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: When storing the contents of a code element inside the binding map, we want it to be aligned to the left since we don't know where it will be used in the rewrite pattern. We detect the indentation where the element starts and then dedent it by that amount before storing it. The contents will get re-indented by the correct amount each time they're used during the rewrite template's instantiation.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: This line is statically dead. I initially included it to avoid an infinite loop when forgetting to handle a pattern type. I'll replace it with a throw statement.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: Yes, but I wanted to avoid allocating a new string.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: I prefer to compute the value each time to avoid using a large amount of memory.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: I agree, but I want to avoid copying fragments of the input into a vector.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: This wasn't intentional. I've fixed it.

clang/tools/pattern-rewriter/src/pattern.cpp
AndyCHHuang: @LU-JOHN I mean keep using trim() and split() but change the implementation of this 2 functions to return string_view to prevent extra copy in the return time.

@0x12CC  Point taken.
I think we might consider create a manager class to maintain the original string and all the string ref, which can guarantee the life time of all the elements and prevent memory fragment if we want to maintain a vector of all the lines.

clang/tools/pattern-rewriter/src/pattern.cpp
AndyCHHuang: @0x12CC I mean a sequence of whitespace in the input when the pattern index is not at SpacingElement.

clang/tools/pattern-rewriter/src/pattern.cpp
AndyCHHuang: @0x12CC 
What if the first line of codeElement has the deepest indent?
If we store all the lines in a vector<pair<string_view, size_t>> including the string_view and the indent size, we can perform dedent by passing a subset of the vector into a dedent function in which we can select the least indent size among all the lines.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: The current pattern element, `Pattern[PatternIndex]`, must be either a `CodeElement`, `SpacingElement`, or `LiteralElement`. The behavior for each of these when matching whitespace in the input is the following:

- `CodeElement`: Matches whitespace from the input and binds it to the element's name.
- `SpacingElement`: Matches whitespace from the input and skips over it.
- `LiteralElement`: This should fail to match. A literal will never expect a whitespace character since the pattern will include `SpacingElement`s for whitespace in the pattern.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: The current implementation does not handle cases where a `CodeElement`'s first line has a deeper indentation than any of the other lines. A `CodeElement` should bind contents with balanced delimiters, so I'm not sure there's a valid case where the indentation would be unbalanced.

For your `vector<pair<string_view, size_t>>` suggestion, I think this approach makes sense. I won't plan to make this change unless we find a test case where we need to find the minimum indentation.

clang/tools/pattern-rewriter/src/pattern.cpp
0x12CC: @LU-JOHN, I updated the condition to make the code more clear.

@AndyCHHuang, I'm not sure if a manager class would solve the problem. You would still need to ensure that the string view doesn't outlive the manager. Without using reference counting (or another garbage collection technique), I would need to change these functions to methods inside a class that owns the string. I'm hesitant to make this change since most lines of code are very short.

clang/tools/pattern-rewriter/src/driver.cpp
```diff
@@ -0,0 +1,99 @@
...
+  int Size = Input.size();
+  while (Index < Size) {
+    char Character = Input[Index];
+    if (Character != '\r') {
+      OutputStream << Character;
+    }
+    Index++;
+  }
+  return OutputStream.str();
+}
```
jzc, line 70: Could also use a function that modifies `Input` like so: `Input.erase(std::remove(Input.Begin(), Input.end(), '\r'), Input.end())` (erase-remove idiom, requires no new allocation)

jzc (approved):

LU-JOHN (approved):

AndyCHHuang (approved):

keryell (commented):
That looks interesting.
Is there a documentation somewhere?

PR #407, opened 2022-12-12
[SYCLomatic] Fix the atomic related functions un-migrated issue

zhiweij1:
Issue:
If two different CUDA header file paths are used in option --cuda-include-path
and option -I (or CPATH), the atomic functions will not be migrated.
We use path in option --cuda-include-path to determine whether an API is from
CUDA or not. But in above case, the atomic function's declaration may from
the path from -I but not --cuda-include-path, so the tool will skip the migration.

Even we prepend another `-I<cuda-include-path>` before ToolInvocation.run(),
it will not fix the problem. Because when generating cc1 args, the
CudaInstallationDetector appends `-internal-isystem=<detected path>` at the end of
cc1 args, and the logic of removing duplicated path will decide to remove
 `<cuda-include-path>` because it is not system path.

Fix:
1. Prepend another -I<cuda-include-path> before ToolInvocation.run()
2. Do not append `<detected path>` in CudaInstallationDetector


tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #411, opened 2022-12-13
[SYCLomatic] Merge branch 'sycl' into 'SYCLomatic'

ziranzha:


zhimingwang36:
Two test cases on bt-version check failed,  and dnn test cases fail as expected.  And we decided to merge directly, and fix the two failed case later.

PR #475, opened 2023-01-04
[SYCLomatic] Fix migration of cudaStreamDefault and cudaStreamNonBlocking

jzc:
Signed-off-by: Cai, Justin justin.cai@intel.com

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -289,8 +289,8 @@ void IncludesCallbacks::MacroExpands(const Token &MacroNameTok,
     std::string MacroNameStr;
     if (auto Identifier = MacroNameTok.getIdentifierInfo())
       MacroNameStr = Identifier->getName().str();
-    if (MapNames::PredefinedStreamName.find(MacroNameStr) !=
-        MapNames::PredefinedStreamName.end()) {
+    if (MacroNameStr == "cudaStreamDefault"
+        || MacroNameStr == "cudaStreamNonBlocking") {
       // Currently, only support examples like,
       // #define CONCATE(name) cuda##name
       // which contains 3 tokens, and the 2nd token is ##.
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -14600,30 +14600,6 @@ void RemoveBaseClassRule::runRule(const MatchFinder::MatchResult &Result) {
 
 REGISTER_RULE(RemoveBaseClassRule, PassKind::PK_Migration)
 
-void PreDefinedStreamHandleRule::registerMatcher(MatchFinder &MF) {
-  MF.addMatcher(integerLiteral(equals(0)).bind("stream"), this);
-  MF.addMatcher(parenExpr(has(cStyleCastExpr(has(
-                              integerLiteral(anyOf(equals(1), equals(2)))))))
-                    .bind("stream"),
-                this);
-}
-
-void PreDefinedStreamHandleRule::runRule(
-    const MatchFinder::MatchResult &Result) {
-  if (auto E = getNodeAsType<Expr>(Result, "stream")) {
-    std::string Str = getStmtSpelling(E);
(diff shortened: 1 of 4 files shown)


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15531,6 +15502,16 @@ void CudaStreamCastRule::runRule(const ast_matchers::MatchFinder::MatchResult &R
...
+  else if (auto E = getNodeAsType<Expr>(Result, "stream")) {
+    std::string Str = getStmtSpelling(E);
+    if (Str == "cudaStreamDefault") {
+      auto &SM = DpctGlobalInfo::getSourceManager();
+      auto Begin = getStmtExpansionSourceRange(E).getBegin();
+      unsigned int Length = Lexer::MeasureTokenLength(
+          Begin, SM, DpctGlobalInfo::getContext().getLangOpts());
+      emplaceTransformation(new ReplaceText(Begin, Length, "0"));
+    }
+  }
```
jzc: `cudaStreamDefault` is migrated to 0 if there has not been a replacement migrating it to `&dpct::get_default_queue()`.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -14699,36 +14699,6 @@ void RemoveBaseClassRule::runRule(const MatchFinder::MatchResult &Result) {
...
-      int Index = DpctGlobalInfo::getHelperFuncReplInfoIndexThenInc();
-      buildTempVariableMap(Index, E, HelperFuncType::HFT_DefaultQueue);
-      emplaceTransformation(new ReplaceText(
-          Begin, Length, "&{{NEEDREPLACEQ" + std::to_string(Index) + "}}"));
-    }
-  }
-}
-
-REGISTER_RULE(PreDefinedStreamHandleRule, PassKind::PK_Migration)
-
```
jzc, line 14632: This migration rule was migrating the macros `cudaStreamDefault`, `cudaStreamLegacy`, and `cudaStreamPerThread` to `dpct::get_default_queue()`. This is correct for `cudaStreamLegacy` and `cudaStreamPerThread`, as these macros' definition expand to a value of `cudaStream_t`. This is incorrect for `cudaStreamDefault` in a edge case where you assign `cudaStreamDefault` to an int. (See the old `macro_test.cu` - the checked output is not valid C++ because you cannot assign a `sycl::queue *` to an int). For the migration of `cudaStreamLegacy` and `cudaStreamPerThread`, this migration behavior is capture by `CudaStreamCastRule`, so it can be removed. 

(Note: the intended purpose of the `cudaStreamDefault` macro is as a flag used with `cudaStreamCreateWithFlags `. The name of `cudaStreamDefault` makes it seem like you can use it as a the default `cudaStream_t`, which perhaps coincidently, you can, because `cudaStreamDefault` expands to `0`,  which is the same value used to denote the default stream with the default stream semantics of the cuda api.)

0x12CC (approved):

LU-JOHN (approved):
LGTM

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15476,10 +15446,11 @@ REGISTER_RULE(TemplateSpecializationTypeLocRule, PassKind::PK_Migration)
...
   MF.addMatcher(
-    castExpr(
-      hasType(typedefDecl(hasAnyName("CUstream", "cudaStream_t"))))
-    .bind("cast"),
-    this);
+     castExpr(hasType(qualType(hasCanonicalType(
+        qualType(pointsTo(namedDecl(hasName("CUstream_st"))))))))
+     .bind("cast"),
+     this);
+  MF.addMatcher(integerLiteral(equals(0)).bind("stream"), this);
```
ziranzha: Suggest do the replacing in IncludesCallbacks::MacroExpands.

ziranzha (approved):

ShengchenJ (commented):
The comment is out of date 
    // If PredefinedStreamName is used with concatenated macro token,
    // detect the previous macro expansion
please also clean it.

PR #477, opened 2023-01-04
[SYCLomatic] If building with old gcc (pre GCC 8) use experimental/filesystem

LU-JOHN:
GCC 7 does not support filesystem interface, but it does have experimental/filesystem.  Update dpct kernel.hpp runtime header
to use experimental/filesystem with pre GCC 7 and earlier  compilers.

### DIFF
--- a/clang/runtime/dpct-rt/include/kernel.hpp.inc
+++ b/clang/runtime/dpct-rt/include/kernel.hpp.inc
@@ -46,7 +46,13 @@
 #include <dlfcn.h>
 #endif
 
+#if (__GNUC__ < 8) && !defined(__llvm__)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
 #include <filesystem>
+namespace fs = std::filesystem;
+#endif
 #include <random>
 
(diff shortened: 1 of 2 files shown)


LU-JOHN:
Use experimental/filesystem instead of C++17 standard's filesystem.

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -46,7 +46,13 @@
 #include <dlfcn.h>
 #endif
 
+#if (__GNUC__ < 8) && !defined(__llvm__)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
```
0x12CC: Can this namespace alias be moved to a scope within the `dpct` namespace? We want to avoid possible name conflicts in files that include this header.

LU-JOHN: Updated to put filesystem alias in detail scope.

jzc (approved):

0x12CC (approved):

PR #481, opened 2023-01-06
[SYCLomatic] Fix dpct::device_vector move semantics bug

Nuullll:
The ownership of the raw pointer `_storage` was not properly transferred when move constructor or move assignment operator is called.

Signed-off-by: Yilong Guo <yilong.guo@intel.com>

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
@@ -243,9 +243,8 @@ public:
   }
   device_vector &operator=(device_vector &&other) {
     // Move assignment operator:
-    // Swap resource with `other`, so that the old resource owned by `this` will
-    // be deallocated by destructor of `other`.
-    this->swap(other);
+    device_vector dummy(std::move(other));
+    this->swap(dummy);
     return *this;
   }
   size_type size() const { return _size; }
@@ -263,6 +262,7 @@ public:
     std::swap(_size, v._size);
     std::swap(_capacity, v._capacity);
     std::swap(_storage, v._storage);
+    std::swap(_alloc, v._alloc);
   }
   reference operator[](size_type n) { return _storage[n]; }
   const_reference operator[](size_type n) const { return _storage[n]; }
--- a/clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
@@ -151,7 +151,11 @@ public:
   }
   device_vector(device_vector &&other)
       : _alloc(get_default_queue()), _size(other.size()),
-        _capacity(other.capacity()) {}
+        _capacity(other.capacity()), _storage(other._storage) {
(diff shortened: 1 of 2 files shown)


Nuullll:
Test: https://github.com/oneapi-src/SYCLomatic-test/pull/190

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -239,9 +243,9 @@ public:
   }
   device_vector &operator=(device_vector &&other) {
     // Move assignment operator:
-    this->_size = std::move(other._size);
-    this->_capacity = std::move(other._capacity);
-    this->_storage = std::move(other._storage);
+    // Swap resource with `other`, so that the old resource owned by `this` will
+    // be deallocated by destructor of `other`.
+    this->swap(other);
```
ziranzha: The old resource should be deallocated inside the assignment, not depends on the destructor of `other`.
Like:
   ```
   device_vector dummy;
   this->swap(dummy);
   ```

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -256,11 +260,9 @@ public:
   T *real_begin() { return _storage; }
   const T *real_begin() const { return _storage; }
   void swap(device_vector &v) {
-    auto temp = std::move(v._storage);
-    v._storage = std::move(this->_storage);
-    this->_storage = std::move(temp);
     std::swap(_size, v._size);
     std::swap(_capacity, v._capacity);
+    std::swap(_storage, v._storage);
```
ziranzha, line 264: `_alloc` also need to be swapped for safety.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
Nuullll: Directly swapping makes move assignment operator exception-free: `std::swap` doesn't throw. The drawback is that the lifetime of old resource may be unexpectedly extended, however, it's still a valid implementation: https://en.cppreference.com/w/cpp/language/move_assignment

> Move assignment operators typically "steal" the resources held by the argument (e.g. pointers to dynamically-allocated objects, file descriptors, TCP sockets, I/O streams, running threads, etc.), rather than make copies of them, and **leave the argument in some valid but otherwise indeterminate state**.

Deallocating resources in the move assignment operator may throw, but the lifetime of old resource is released in-time (intuitively).

There're debates over the move assignment implementations:
- https://stackoverflow.com/questions/6687388/why-do-some-people-use-swap-for-move-assignments
- https://stackoverflow.com/questions/32234623/using-swap-to-implement-move-assignment
- https://stackoverflow.com/questions/9847860/questions-about-the-move-assignment-operator

So I'm okay with both.

ziranzha (approved):

Nuullll:
Hi, could anyone help merge this? Thanks.

PR #468, opened 2023-01-03
[SYCLomatic][DPCT] io_iterator_pair and associated sort APIs

danhoeflinger:
This add the struct io_iterator_pair to the dpct header, and and adds associated APIs for sort_keys, sort_pairs, and segmented_sort_pairs.  

Signed-off-by: Dan Hoeflinger <dan.hoeflinger@intel.com>

PR #483, opened 2023-01-06
[SYCLomatic] [DPCT] Fix for in/out types segmented sort

danhoeflinger:
This resolves the compiler issues with providing different input and output types for keys and values into segmented_sort_pairs.  For example, one may want to have `const int*` as input key iterator type and `int*` as output key iterator type.  This is now possible with this fix.

Signed-off-by: Dan Hoeflinger <dan.hoeflinger@intel.com>

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
@@ -1540,29 +1540,26 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
     // coordinate to mark segments
     policy.queue()
         .submit([&](sycl::handler &h) {
-          h.parallel_for(sycl::nd_range<1>{work_group_size, work_group_size},
-                         ([=](sycl::nd_item<1> item) {
-                           auto sub_group = item.get_sub_group();
-                           ::std::size_t num_subgroups =
-                               sub_group.get_group_range().size();
-                           ::std::size_t local_size =
-                               sub_group.get_local_range().size();
+          h.parallel_for(
+              sycl::nd_range<1>{work_group_size, work_group_size},
+              ([=](sycl::nd_item<1> item) {
+                auto sub_group = item.get_sub_group();
+                ::std::size_t num_subgroups =
+                    sub_group.get_group_range().size();
+                ::std::size_t local_size = sub_group.get_local_range().size();
 
-                           ::std::size_t sub_group_id =
-                               sub_group.get_group_id();
-                           while (sub_group_id < nsegments) {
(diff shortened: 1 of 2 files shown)


danhoeflinger:
Similar to #468, I believe my changes to the dependencies in the .inc file caused some changes in the api_tests.  However, resolving this is beyond the scope of my expertise.  Can someone help adjust these tests? @yihanwg perhaps?

danhoeflinger:
> Similar to #468, I believe my changes to the dependencies in the .inc file caused some changes in the api_tests. However, resolving this is beyond the scope of my expertise. Can someone help adjust these tests? @yihanwg perhaps?

I've resolved the issue in the api_tests.

jzc (approved):

yihwang-nv:
Please ignore `Jenkins/Windows_CI` failed, it caused by a known issue.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -1531,29 +1532,26 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+              sycl::nd_range<1>{work_group_size, work_group_size},
+              ([=](sycl::nd_item<1> item) {
+                auto sub_group = item.get_sub_group();
+                ::std::size_t num_subgroups =
+                    sub_group.get_group_range().size();
+                ::std::size_t local_size = sub_group.get_local_range().size();
+
+                ::std::size_t sub_group_id = sub_group.get_group_id();
+                while (sub_group_id < nsegments) {
+                  ::std::size_t subgroup_local_id = sub_group.get_local_id();
```
yihwang-nv, line 1553: I think here don't need `::` before std?

yihwang-nv (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: We use it to ensure that we are picking up the `std` from the global namespace, `::std` is more explicit in meaning than `std`.  You are correct that it very likely wont many any difference here in practice here, but it should not hurt us.

danhoeflinger:
Thanks for the approvals.
Just in case you are waiting on me to merge, I do not have permissions to do so.

yihwang-nv:
@danhoeflinger can you please resolve the conflict

danhoeflinger:
> @danhoeflinger can you please resolve the conflict

Done.

yihwang-nv (approved):
Thanks, still LGTM!

zhimingwang36 (approved):
LGTM

PR #486, opened 2023-01-09
[SYCLomatic] check for filesystem precisely

LU-JOHN:
Check for filesystem precisely.  Previous check did not work for gcc 7.5.0 invoked through icpx.  Precise check tested with
icpx , icpx-gcc 9.4.0, MSVC which uses filesystem.
icpx-gcc 7.5.0 which uses experimental/filesystem.

### DIFF
--- a/clang/runtime/dpct-rt/include/kernel.hpp.inc
+++ b/clang/runtime/dpct-rt/include/kernel.hpp.inc
@@ -46,10 +46,10 @@
 #include <dlfcn.h>
 #endif
 
-#if (__GNUC__ < 8) && !defined(__llvm__)
-#include <experimental/filesystem>
-#else
+#if defined(__has_include) && __has_include(<filesystem>)
 #include <filesystem>
+#else
+#include <experimental/filesystem>
 #endif
 
 #include <random>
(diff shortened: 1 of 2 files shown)


zhiweij1 (approved):

clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -46,10 +46,10 @@
 #include <dlfcn.h>
 #endif
 
-#if (__GNUC__ < 8) && !defined(__llvm__)
-#include <experimental/filesystem>
-#else
+#if defined(__has_include) && __has_include(<filesystem>)
 #include <filesystem>
+#else
+#include <experimental/filesystem>
```
ziranzha, line 52: Should we support elder version compilers that don't have <experimental/filesystem>?

LU-JOHN: experimental/filesystem was added in gcc 5.3.0 which was released in December 2015.

LU-JOHN: The GCC compiler requirements for building migrated code is:

GCC* 7.5 or later for SUSE, 9.3 or later for Ubuntu, 8.2.1 or later for Red Hat, 12.0.1 or later for Fedora

ziranzha (approved):

PR #487, opened 2023-01-09
[SYCLomatic] Find real path in getInstallPath()

0x12CC:
This PR fixes an issue where the detected installation path is incorrect when the `dpct` binary is linked into another directory.

### DIFF
--- a/clang/lib/DPCT/DPCT.cpp
+++ b/clang/lib/DPCT/DPCT.cpp
@@ -267,6 +267,7 @@ std::string getInstallPath(clang::tooling::ClangTool &Tool,
   }
 
   makeCanonical(InstalledPath);
+  llvm::sys::fs::real_path(InstalledPath, InstalledPath, true);
   StringRef InstalledPathParent(llvm::sys::path::parent_path(InstalledPath));
   // Move up to parent directory of bin directory
   StringRef InstallPath = llvm::sys::path::parent_path(InstalledPathParent);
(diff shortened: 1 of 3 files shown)


clang/lib/DPCT/DPCT.cpp
```diff
@@ -267,6 +267,7 @@ std::string getInstallPath(clang::tooling::ClangTool &Tool,
   }
 
   makeCanonical(InstalledPath);
+  llvm::sys::fs::real_path(InstalledPath, InstalledPath, true);
```
LU-JOHN, line 270: Can you give an example InstalledPath input that is not sufficiently modified by makeCanonical that needs the call to llvm:;sys::fs::real_path.

clang/test/dpct/symlink/symlink.cu
```diff
@@ -0,0 +1,13 @@
+// UNSUPPORTED: -windows-
```
LU-JOHN, line 1: Can this test be run on Windows?

clang/lib/DPCT/DPCT.cpp
0x12CC: The test case provides one example. When the `dpct` binary is symbolically linked, `makeCannonical()` won't find the real path. We need the real path since that's where the headers and other installation files are. The test case fails without this change.

clang/test/dpct/symlink/symlink.cu
0x12CC: I don't think so since the way I've created the test workspace is Linux-specific.

LU-JOHN (approved):

jzc (approved):

PR #492, opened 2023-01-11
[SYCLomatic] Report error if filesystem not supported

LU-JOHN:
Report error if filesystem header not found.

### DIFF
--- a/clang/runtime/dpct-rt/include/kernel.hpp.inc
+++ b/clang/runtime/dpct-rt/include/kernel.hpp.inc
@@ -48,8 +48,10 @@
 
 #if defined(__has_include) && __has_include(<filesystem>)
 #include <filesystem>
-#else
+#elif defined(__has_include) && __has_include(<experimental/filesystem>)
 #include <experimental/filesystem>
+#else
+#error "SYCLomatic runtime requires C++ filesystem support"
 #endif
 
 #include <random>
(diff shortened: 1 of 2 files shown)


clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -48,8 +48,10 @@
 
 #if defined(__has_include) && __has_include(<filesystem>)
 #include <filesystem>
-#else
+#elif defined(__has_include) && __has_include(<experimental/filesystem>)
```
ziranzha, line 51: Is the case that the host compiler has <experimental/filesystem> but don't have `__has_include` defined exist?

LU-JOHN: __has_include and filesystem are both part of C++17.    If a toolchain supports experimental/filesystem but not __has_include it would be very unusual, since  __has_include is simpler to add than filesystem.

ziranzha: I think <experimental/filesystem> is added before <filesystem>, so __has_include may be invalid when the compiler has <experimental/filesystem>.

0x12CC (approved):

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Our requirements specify that GCC 7.5 is the minimum GCC that we support:  https://www.intel.com/content/www/us/en/developer/articles/system-requirements/intel-dpc-compatibility-tool-system-requirements.html, and GCC 7.5 has experimental/filesystem and __has_include.

ziranzha (approved):

PR #528, opened 2023-01-30
[SYCLomatic] null_type

danhoeflinger:
Implementation for dpct::null_type for to indicate the lack of argument.

### DIFF
--- a/clang/test/dpct/helper_files_ref/include/dpl_extras/functional.h
+++ b/clang/test/dpct/helper_files_ref/include/dpl_extras/functional.h
@@ -22,6 +22,8 @@
 
 namespace dpct {
 
+struct null_type {};
+
 namespace internal {
 
 template <class _ExecPolicy, class _T>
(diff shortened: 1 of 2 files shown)


clang/test/dpct/helper_files_ref/include/dpl_extras/functional.h
```diff
@@ -22,6 +22,8 @@
 
 namespace dpct {
 
+struct null_type {};
```
zhimingwang36, line 25: this PR doesn't show the use of null_type{}.  So not sure how it is used and why it is necessary to add null_type.

danhoeflinger: I've just added a new PR into SYCLomatic-test which shows the intended usage of `dpct::null_type`.  

https://github.com/oneapi-src/SYCLomatic-test/pull/219

It will also likely be used in future dpct helper header APIs to a similar effect.

zhimingwang36 (approved):
LGTM

PR #529, opened 2023-01-31
[SYCLomatic] Fixes for DPCT1013 diagnostics

vladimir-polin:
then->than typo fixed and "precision" changed to "accuracy" which is more relevant to this particular diagnostics.

Signed-off-by: Vladimir Polin Vladimir.Polin@intel.com


clang/lib/DPCT/Diagnostics.inc
```diff
@@ -39,8 +39,8 @@ DEF_WARNING(TRNA_WARNING_OVERLOADED_API_FOUND, 1011, "The tool detected overload
...
 DEF_COMMENT(TIME_MEASUREMENT_FOUND, 1012, "Detected kernel execution time measurement pattern and generated an initial code for time measurements in SYCL. You can change the way time is measured depending on your goals.")
-DEF_WARNING(ROUNDING_MODE_UNSUPPORTED, 1013, "The rounding mode could not be specified and the generated code may have different precision then the original code. Verify the correctness. SYCL math built-ins rounding mode is aligned with OpenCL C 1.2 standard.")
-DEF_COMMENT(ROUNDING_MODE_UNSUPPORTED, 1013, "The rounding mode could not be specified and the generated code may have different precision then the original code. Verify the correctness. SYCL math built-ins rounding mode is aligned with OpenCL C 1.2 standard.")
+DEF_WARNING(ROUNDING_MODE_UNSUPPORTED, 1013, "The rounding mode could not be specified and the generated code may have different accuracy than the original code. Verify the correctness. SYCL math built-ins rounding mode is aligned with OpenCL C 1.2 standard.")
```
zhiweij1, line 42: I still prefer "precision" to "accuracy". @zhimingwang36 what's your opinion?

vladimir-polin: Look at https://en.wikipedia.org/wiki/Rounding article
"Precision" term is mostly for double precision, single precision, integer, etc. 
"Accuracy" term or "Accurate rounding" is used for rounding modes themselves 
So unless there is unexpected DP->SP precision loss covered by this diagnostics I suggest using "accuracy".

zhimingwang36: LGTM

zhimingwang36 (approved):
LGTM

PR #537, opened 2023-02-01
[SYCLomatic] Refine the migration of thrust::device_malloc_allocator

tomflinda:


zhiweij1 (approved):

clang/test/dpct/helper_files_ref/include/memory.hpp
```diff
@@ -1304,6 +1304,8 @@ namespace deprecated {
 template <typename T>
 using usm_host_allocator = detail::deprecated::usm_allocator<T, sycl::usm::alloc::host>;
 
+template <typename T>
+using usm_device_allocator = detail::deprecated::usm_allocator<T, sycl::usm::alloc::shared>;
```
zhimingwang36, line 1313: it 's not totally correct to tell sycl::usm::alloc::shared is a device allocator.

clang/test/dpct/thrust-for-h2o4gpu-specific-case.cu
```diff
@@ -11,24 +11,29 @@
...
+//CHECK-NEXT:  foo_cpy(*tt[0], *dd[0]);
 //CHECK-NEXT:}
 void foo_1() {
   thrust::device_vector<int, thrust::device_malloc_allocator<int>> **tt=NULL;
   thrust::device_vector<int, thrust::device_malloc_allocator<int>> *dd[10];
   foo_cpy(*tt[0], *dd[0]);
 }
 
+//CHECK:namespace foo_struct { template <typename T, typename MemorySpace> struct default_memory_allocator : dpct::deprecated::usm_device_allocator<T>{};}
+namespace foo_struct { template <typename T, typename MemorySpace> struct default_memory_allocator : thrust::device_malloc_allocator<T>{};}
```
zhimingwang36, line 39: is the mapping right:  using shared alloctor to migrate thrust::device_malloc_allocator?

clang/test/dpct/helper_files_ref/include/memory.hpp
tomflinda: From the description of [thrust::device_malloc_allocator](https://thrust.github.io/doc/classthrust_1_1device__malloc__allocator.html), thrust::device_malloc_allocator uses device memory, so detail::deprecated::usm_allocator inherited from sycl::usm_allocator should be specified with "usm::alloc::device" instead.  While currently sycl::usm_allocator does not support "usm::alloc::device"  according to https://github.com/intel/llvm/blob/3df87e20569ea63d0a74de525b6d19788dd8afca/sycl/include/sycl/usm/usm_allocator.hpp#L39.
I will sync with compiler team to request sycl::usm_allocator to support "usm::alloc::device".

clang/test/dpct/thrust-for-h2o4gpu-specific-case.cu
tomflinda: Yes, there is semantically difference, same as implementation of dpct::device_vector, which is implemented by using usm_allocator<T, sycl::usm::alloc::shared>. 
And from feedback of oneDPL team, current usage of usm shared memory is more in line with usage patterns one would choose for this type of functionality if writing code from scratch with sycl.

clang/test/dpct/helper_files_ref/include/memory.hpp
tomflinda: Synced with compiler team, the reason that sycl::usm_allocator does not support "usm::alloc::device" is :
there's a fundamental C++ problem here with a tradeoff. If device allocations is supported, it breaks a lot of C++ - std::allocate_shared for example will not work.

timmiesmith (approved):
LGTM.

PR #543, opened 2023-02-02
[SYCLomatic] Prevent indirection on void* for async_convolution_forward

0x12CC:

### DIFF
--- a/clang/test/dpct/dnn/convolution.cu
+++ b/clang/test/dpct/dnn/convolution.cu
@@ -70,6 +70,10 @@ int main() {
     float alpha = 1.0f, beta = 0.0f;
     // CHECK: handle.async_convolution_forward(covdes, dnnl::algorithm::convolution_direct, alpha, dataTensor, data, filterTensor, filter, beta, outTensor, out);
     cudnnConvolutionForward(handle, &alpha, dataTensor, data, filterTensor, filter, covdes, CUDNN_CONVOLUTION_FWD_ALGO_DIRECT, workspacesize, size, &beta, outTensor, out);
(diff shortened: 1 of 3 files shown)


clang/test/dpct/dnn/convolution.cu
```diff
@@ -70,6 +70,10 @@ int main() {
     float alpha = 1.0f, beta = 0.0f;
     // CHECK: handle.async_convolution_forward(covdes, dnnl::algorithm::convolution_direct, alpha, dataTensor, data, filterTensor, filter, beta, outTensor, out);
     cudnnConvolutionForward(handle, &alpha, dataTensor, data, filterTensor, filter, covdes, CUDNN_CONVOLUTION_FWD_ALGO_DIRECT, workspacesize, size, &beta, outTensor, out);
+    // CHECK: handle.async_convolution_forward(covdes, dnnl::algorithm::convolution_direct, *(float *)(void *)&alpha, dataTensor, data, filterTensor, filter, *(float *)(void *)&beta, outTensor, out);
```
ziranzha, line 73: Maybe `*(float *)&alpha` is better?

0x12CC: I think that it looks simpler but may change the semantics of the code being migrated. Consider the following program:
```cpp
#include <iostream>

static float A = 4.4f;
static float B = 7.7f;

class Type {
public:
  operator void *() const { return reinterpret_cast<void *>(&A); }
  operator float *() const { return &B; }
};

void printFloat(float x) { std::cout << x << std::endl; }

int main() {
  Type x;
  printFloat(*x);
  printFloat(*(float *)x);
  printFloat(*(float *)(void *)x);
}
```
The last two statements will print different numbers. I want to avoid removing any part of the expression that's written in the original source.

LU-JOHN (approved):
LGTM

ziranzha (approved):

PR #539, opened 2023-02-02
[SYCLomatic] Fix libcu lit failed due to wrong cuda version support

wenhuiNi:
Signed-off-by: Ni, Wenhui <wenhui.ni@intel.com>

PR #544, opened 2023-02-02
[SYCLomatic] cuda/std/array requires CUDA 11.7+

LU-JOHN:
Disable libcuarray.cu test if CUDA version does not support cuda/std/array.
cuda/std/array was added in CUDA 11.7.


0x12CC (approved):

jzc (approved):

clang/test/dpct/LibCU/libcu_array.cu
```diff
@@ -1,5 +1,5 @@
 // UNSUPPORTED: v7.0, v7.5, v8.0, v9.0, v9.2, v10.0, v10.1, v10.2, v11.0, v11.1
```
wenhuiNi, line 1: for disable lit  if CUDA version does not support， we need disable **windows** and linux.
like #539

LU-JOHN:
Fixed in PR 539.

PR #545, opened 2023-02-02
[SYCLomatic] Make kernel library more secure

LU-JOHN:
Make kernel library more secure.

### DIFF
--- a/clang/runtime/dpct-rt/include/kernel.hpp.inc
+++ b/clang/runtime/dpct-rt/include/kernel.hpp.inc
@@ -126,6 +126,10 @@ static inline fs::path write_data_to_file(char const *const data,
                                    size_t size) {
   std::error_code ec;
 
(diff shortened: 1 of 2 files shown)


clang/runtime/dpct-rt/include/kernel.hpp.inc
```diff
@@ -125,6 +125,10 @@ static fs::path write_data_to_file(char const *const data,
                                    size_t size) {
   std::error_code ec;
 
+  if (sizeof(size_t)>=sizeof(std::streamsize) &&
```
ziranzha: Need format

ziranzha (approved):

clang/runtime/dpct-rt/include/kernel.hpp.inc
LU-JOHN: Updated for clang-format.

0x12CC (approved):

PR #546, opened 2023-02-02
[SYCLomatic] Map result types to preserve overload resolution

0x12CC:


clang/runtime/dpct-rt/include/util.hpp.inc
```diff
@@ -51,6 +51,26 @@
...
+// DPCT_CODE
+namespace detail {
+
+template <typename tag, typename T> class new_type {
+public:
+  new_type(T value) : value{value} {}
+  operator T() { return value; }
+
+private:
+  T value;
```
ziranzha, line 65: Maybe we don't need `value`, just return `0` for `operator int`.

clang/runtime/dpct-rt/include/util.hpp.inc
```diff
@@ -51,6 +51,26 @@
 
 namespace dpct {
 
+// DPCT_LABEL_BEGIN|err_tmp_types|dpct
+// DPCT_DEPENDENCY_EMPTY
+// DPCT_CODE
+namespace detail {
+
+template <typename tag, typename T> class new_type {
```
ziranzha: `new_type` is confusing. We need a better name such as `generic_error_type` or `error_type_base`.

clang/lib/DPCT/MapNames.cpp
```diff
@@ -93,8 +93,8 @@ void MapNames::setExplicitNamespaceMap() {
       {"cudaDeviceProp",
        std::make_shared<TypeNameRule>(getDpctNamespace() + "device_info",
                                       HelperFeatureEnum::Device_device_info)},
-      {"cudaError_t", std::make_shared<TypeNameRule>("int")},
-      {"cudaError", std::make_shared<TypeNameRule>("int")},
+      {"cudaError_t", std::make_shared<TypeNameRule>(getDpctNamespace() + "err_tmp0")},
```
LU-JOHN: The "tmp" in err_tmp0 and err_tmp1 in the migrated code looks unsettling to me as part of a typename.  I would prefer something like err_t/status_t or err0_t/err1_t.

jzc (approved):

clang/runtime/dpct-rt/include/util.hpp.inc
0x12CC: I've changed it to `generic_error_type`.

clang/runtime/dpct-rt/include/util.hpp.inc
0x12CC: I don't think this change makes sense since the semantics would differ. The original type is not stateless, so even though we always map to 0, there are other ways to get a non-zero value of type `err0`.

clang/lib/DPCT/MapNames.cpp
0x12CC: Removed `tmp` from the type names.

ziranzha (approved):

PR #553, opened 2023-02-03
[SYCLomatic] Fix migration of cv-qualified cg::thread_block in function parameter

jzc:

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -2673,8 +2673,9 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
       EndLoc = SM->getExpansionRange(TL->getBeginLoc()).getEnd();
     }
 
-    std::string CanonicalTypeStr = DpctGlobalInfo::getTypeName(
-        TL->getUnqualifiedLoc().getType().getCanonicalType());
+    std::string CanonicalTypeStr =
+      DpctGlobalInfo::getUnqualifiedTypeName(
+        TL->getType().getCanonicalType());
     StringRef CanonicalTypeStrRef(CanonicalTypeStr);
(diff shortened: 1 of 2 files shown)


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -2702,8 +2702,8 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
       }
     }
 
-    if (DpctGlobalInfo::getTypeName(TL->getType().getCanonicalType()) ==
-        "cooperative_groups::__v1::thread_block") {
+    if (CanonicalTypeStr.find("cooperative_groups::__v1::thread_block")
```
0x12CC: This condition will be true for `"cooperative_groups::__v1::thread_block_tile<"` from line 2680 when `ETL` is `nullptr`. Would it be better to explicitly remove the qualifiers and compare for equality?

jzc: Yea, good point. I'll make this fix.

0x12CC (approved):

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -2673,8 +2673,8 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
       EndLoc = SM->getExpansionRange(TL->getBeginLoc()).getEnd();
     }
 
-    std::string CanonicalTypeStr =
-        DpctGlobalInfo::getTypeName(TL->getType().getCanonicalType());
+    std::string CanonicalTypeStr = DpctGlobalInfo::getTypeName(
```
ziranzha: How about use `DpctGlobalInfo::getUnqualifiedTypeName`?

ziranzha (approved):

zhimingwang36 (approved):
LGTM

PR #563, opened 2023-02-07
[SYCLomatic] Migrate thrust::zip_iterator to dpct::zip_iterator

zhiweij1:


ziranzha (approved):

clang/test/dpct/helper_files_ref/include/dpl_extras/iterators.h
```diff
@@ -146,8 +146,16 @@ template <typename KeyTp, typename _ValueTp> struct make_key_value_pair {
...
+template <class T> struct __zip_iterator_impl;
+template <class... Ts> struct __zip_iterator_impl<std::tuple<Ts...>> {
+  using type = oneapi::dpl::zip_iterator<Ts...>;
+};
+
 } // end namespace detail
 
+template <typename... Ts>
+using zip_iterator = typename detail::__zip_iterator_impl<Ts...>::type;
+
```
danhoeflinger, line 161: I wonder if it will be confusing to users to have `oneapi::dpl::zip_iterator` and `dpct::zip_iterator` named the same but with a different way to specify their contained types.  Would `dpct::legacy_zip_iterator` be a better name?
On the other hand, we aren't creating a new type here but rather are using this to more easily specify the correct type, perhaps naming it something different will result in more confusion.  I'm torn on this.

danhoeflinger (commented):
I can only speak to the changes in the dpct header and tests, but this functionally looks good to me other than the naming question.

clang/test/dpct/helper_files_ref/include/dpl_extras/iterators.h
zhiweij1: @danhoeflinger @zhimingwang36 
How about using `dpct::zip_iterator_compat` and adding comments for this this type?
E.g.,
```
// dpct::zip_iterator_compat can only accept std::tuple type as template
// argument for compatibility purpose. Please use oneapi::dpl::zip_iterator if
// you want to pass iterator's types directly.
template <typename... Ts>
using zip_iterator_compat = typename detail::__zip_iterator_impl<Ts...>::type;
```

zhimingwang36: I'd perfer using dpct::zip_iterator,  as dpct:: namespace itself has telling the interface has some speicific behaviour (on tuple type), while zip_iterator tell it's usage are almost same as dpl::zip_iterator.  And the iters can be assigned from/to each other.

danhoeflinger: I'm OK with either option.  Perhaps the best option is to keep the name, as we are in fact not creating a new type, just an alternative way to specify a `oneapi::dpl::zip_iterator` type.  I do think that a comment would be useful, the one above works for me.

zhiweij1: @danhoeflinger I'm still using `dpct::zip_iterator`, but add some comments in the updated commit.

zhimingwang36 (approved):
LGTM

PR #554, opened 2023-02-06
[SYCLomatic] Skip unsupported cuda version for test cub/devicelevel/device_unique_by_key.cu

zhiweij1:
Signed-off-by: Jiang, Zhiwei <zhiwei.jiang@intel.com>

PR #565, opened 2023-02-07
[SYCLomatic]Fix cub lit failed due to wrong cuda version support

wenhuiNi:


wenhuiNi:
duplicated #554

PR #584, opened 2023-02-09
[SYCLomatic] Fix replace_if/replace_copy_if for corner case

LU-JOHN:
replace_if/replace_copy_if must be done based on type of iterator value, not on type of new value.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
@@ -91,7 +91,7 @@ Iter3 replace_copy_if(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask,
                        std::random_access_iterator_tag>::value,
       "Iterators passed to algorithms must be random-access iterators.");
   return std::transform(std::forward<Policy>(policy), first, last, mask, result,
-                        internal::replace_if_fun<typename std::iterator_traits<Iter1>::value_type, Pred>(p, new_value));
+                        internal::replace_if_fun<typename std::iterator_traits<Iter3>::value_type, Pred>(p, new_value));
 }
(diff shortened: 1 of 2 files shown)


jzc (approved):
LGTM. From what I understood, the example corner case is when 
* `T = int`, 
* `std::iterator_traits<Iter1>::value_type = long long`, 
* `pred(*first)` is false,

then the call of `std::transform` in `replace_if` will essentially do `*first = int(*first)` (`*result = int(*first)` for `replace_copy_if`)

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -91,7 +91,7 @@ Iter3 replace_copy_if(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask,
                        std::random_access_iterator_tag>::value,
       "Iterators passed to algorithms must be random-access iterators.");
   return std::transform(std::forward<Policy>(policy), first, last, mask, result,
-                        internal::replace_if_fun<T, Pred>(p, new_value));
+                        internal::replace_if_fun<typename std::iterator_traits<Iter1>::value_type, Pred>(p, new_value));
```
ziranzha: Traits of Iter3 should be used here.
For case sizeof(TraitsOf(Iter3)) == sizeof(T) > sizeof(TraitsOf(Iter1)), the `new_value` will first be casted to TraitsOf(Iter1), then TraitsOf(Iter3). Precision will be lost.

LU-JOHN: Fixed.  SYCLomatic-test will be updated to hit this corner.

PR #589, opened 2023-02-13
[NFC][SYCLomatic] Use std::optional instead of llvm::Optional

yihwang-nv:
Currently, `llvm::Optional` is an alias of `std::optional`, and it will be remove after `llvm-16` .

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -126,7 +126,7 @@ int IncludesCallbacks::findPoundSign(SourceLocation DirectiveStart) {
   if (CharDataInvalid || !Entry.isFile()) {
     return -1;
   }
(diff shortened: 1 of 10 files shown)


wenhuiNi (approved):

jzc:
It seems there are some unqualified uses of `llvm::Optional` in, for example, `CallExprRewriter.h`, `MemberExprRewriter.h` and `TypeLocRewriter.h`. And should the include of `llvm/ADT/Optional.h` be removed?

yihwang-nv:
> It seems there are some unqualified uses of `llvm::Optional` in, for example, `CallExprRewriter.h`, `MemberExprRewriter.h` and `TypeLocRewriter.h`. And should the include of `llvm/ADT/Optional.h` be removed?

Thanks, removed.

jzc (approved):

PR #596, opened 2023-02-14
[SYCLomatic][DPCT] Decouple sycl version check from oneDPL 

danhoeflinger:
Decoupling version check from oneDPL internal preprocessor variable so that this header works with any version of oneDPL.  `_ONEDPL_LIBSYCL_VERSION` had its name changed recently in oneDPL, causing co-dependency with change of the name here.  It is easy enough to break that dependence by relying on the source.

Also, correcting a version number for `sycl::buffer_allocator<>` to 60000, as was done in oneDPL here: https://github.com/oneapi-src/oneDPL/pull/707.


timmiesmith (approved):
LGTM. Thank you.

yihwang-nv (approved):

PR #607, opened 2023-02-17
[DOC] Update HTML for doc publication

krisdale:
Updated HTML that includes new migration status table. Corresponding content update is PR 606.


jiejanezhang (approved):

dev_guide/api-mapping-status.html
```diff
@@ -0,0 +1,18300 @@
+
```
zhimingwang36, line 1: There are some updates on API mapping status, please check with Wenhui. Thanks

zhimingwang36 (approved):
LGTM

PR #619, opened 2023-02-17
[SYCLomatic] dpct::device_vector standardizing and protecting allocation

danhoeflinger:
This PR consolidates and protects the allocation of `dpct::device vector`.  We use `_min_capacity()` to avoid allocating size zero allocations.  

We avoid calls to `std::copy` and `std::fill` when the size is zero.

We change the rvalue constructor, `device_vector(device_vector &&other)` .
 We set `other._capacity = 0`, rather than using `_min_capacity()`, as this conflicted with the actual allocated storage.  

While we are touching this code, we fix  `std::` -> `::std::`  for `dpct::device_vector`.


zhimingwang36 (approved):
LGTM

ziranzha (approved):

PR #620, opened 2023-02-17
[SYCLomatic] dpct::device_vectors::shrink_to_fit() bugfix for non USM

danhoeflinger:
Bugfix for non-USM shrink_to_fit implementation to avoid uninitialized memory reads and out of bounds writes.


zhimingwang36 (approved):
LGTM, please also add behavior test for the updated code.

ziranzha (approved):

PR #652, opened 2023-02-22
[SYCLomatic] Migrate CU_LAUNCH_PARAM_* macros

LU-JOHN:
Signed-off-by: Lu, John [john.lu@intel.com](mailto:john.lu@intel.com)

### DIFF
--- a/clang/lib/DPCT/MapNames.cpp
+++ b/clang/lib/DPCT/MapNames.cpp
@@ -4129,15 +4129,15 @@ std::unordered_map<std::string, MacroMigrationRule> MapNames::MacroRuleMap{
 
     {"CU_LAUNCH_PARAM_BUFFER_SIZE",
      MacroMigrationRule("kernel_param_rule", RulePriority::Fallback,
-                        "CU_LAUNCH_PARAM_BUFFER_SIZE", "DPCT_KERNEL_PARAM_SIZE",
+                        "CU_LAUNCH_PARAM_BUFFER_SIZE", "((void *) 2)",
                         HelperFeatureEnum::Kernel_kernel_library)},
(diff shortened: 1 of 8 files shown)


jzc (approved):

AndyCHHuang (approved):
LGTM

clang/lib/DPCT/MapNames.cpp
```diff
@@ -4122,6 +4122,19 @@ std::unordered_map<std::string, MacroMigrationRule> MapNames::MacroRuleMap{
     {"cudaStreamDefault",
      MacroMigrationRule("cudaStreamDefault_rule", RulePriority::Fallback,
                         "cudaStreamDefault", "0")},
+
+    {"CU_LAUNCH_PARAM_BUFFER_SIZE",
+     MacroMigrationRule("kernel_param_rule", RulePriority::Fallback,
+                        "CU_LAUNCH_PARAM_BUFFER_SIZE", "DPCT_KERNEL_PARAM_SIZE",
```
zhimingwang36: Could u migrate those macros to literal values (like 0,1,2)?

LU-JOHN: Isn't it preferable to use a symbolic value?   If were writing SYCL code to use this interface we wouldn't want them to use magic numbers in their source.

zhimingwang36: It's quite API specific things, it's not common. From SYCL view, we don't want to introduce this kind new macros in dpct:: namespace.

LU-JOHN: Updated code to use literal values

zhimingwang36 (approved):
LGTM

PR #653, opened 2023-02-22
[SYCLomatic] Fix convolution_desc bool conversion

jzc:
After calling `convolution_desc::operator=(std::nullptr_t)`, `convolution_desc::operator bool()` should return false. This is the opposite of what it should be right now (it returns true).


0x12CC (approved):

intwanghao (approved):
LGTM

PR #661, opened 2023-02-23
[SYCLomatic] Migrate CUdeviceptr to dpct::device_ptr

jzc:
Previously `CUdeviceptr` was migrated to `char *`. If there was a declaration like
```c++
CUdeviceptr a = (CUdeviceptr)0, b = (CUdeviceptr)0;
```
the migration would have a type error due to declarator syntax:
```c++
char * a = (char *)0, b = (char *)0;
```
(`b` would have type char, but cannot initialize `char` with `char *`)

Now, `CUdeviceptr` is migrated to `dpct::device_ptr`, which is an alias for `char *`, avoiding syntax issues.


0x12CC (approved):

AndyCHHuang (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #666, opened 2023-02-24
[SYCLomatic] Fix a bug of filtering mode when fetching 1D image data

ziranzha:
Should use nearest mode instead of linear mode when migrating tex1Dfetch


zhimingwang36 (approved):
LGTM

PR #669, opened 2023-02-24
[DOC] Remove unecessary link from footer

krisdale:
Remove cookies link from footer as it is not needed/used.


zhimingwang36 (approved):
LGTM

PR #674, opened 2023-02-27
[SYCLomatic] Add help function for Dropout layer

intwanghao:


clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4939,82 @@ sycl::event engine_ext::async_rnn_backward(
...
+  float scale_factor = 1.f / (1.f - p);
+  void *cache = workspace;
+
+  memory_desc_ext rng_data_desc(
+      ::dnnl::memory::desc(src_desc.get_dims(), ::dnnl::memory::data_type::s32,
+                           src_desc.get_strides()));
+  if (src_desc.get_desc().get_data_type() != ::dnnl::memory::data_type::s32) {
+    cache = allocate(rng_data_desc);
+  }
+  oneapi::mkl::rng::generate(*desc.get_distr(), *desc.get_rng_engine(),
```
ziranzha: How about using dropout_desc::generate for these two functions? And remove dropout_desc::get_rng_engine, dropout_desc::get_distr and dropout_desc::get_state.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -742,6 +745,79 @@ public:
...
+// DPCT_DEPENDENCY_BEGIN
+// DnnlUtils|dropout_desc_def
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+/// A class holding description for a Dropout operation.
+class dropout_desc {
+  float _p;
+  unsigned long long _seed;
+  oneapi::mkl::rng::philox4x32x10 *_rng_engine;
+  oneapi::mkl::rng::bernoulli<std::int32_t, oneapi::mkl::rng::bernoulli_method::icdf> *_distr;
```
ziranzha: distribution can be built only in used function.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -742,6 +745,79 @@ public:
...
+// DPCT_LABEL_BEGIN|dropout_desc|dpct::dnnl
+// DPCT_DEPENDENCY_BEGIN
+// DnnlUtils|dropout_desc_def
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+/// A class holding description for a Dropout operation.
+class dropout_desc {
+  float _p;
+  unsigned long long _seed;
+  oneapi::mkl::rng::philox4x32x10 *_rng_engine;
```
ziranzha: use shared_ptr for engine

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -2147,6 +2277,44 @@ public:
...
+                       size_t state_size, unsigned long long seed) {
+  _p = p;
+  _seed = seed;
+  if (state) {
+    _state = state;
+    sycl::queue *q = engine.get_queue();
+    _rng_engine = new oneapi::mkl::rng::philox4x32x10(*q, seed);
+    _distr = new oneapi::mkl::rng::bernoulli<
+        std::int32_t, oneapi::mkl::rng::bernoulli_method::icdf>(p);
+    if (state_size < oneapi::mkl::rng::get_state_size(*_rng_engine)) {
```
ziranzha: Maybe we can store the state_size as a `static const` variable in `dropout_desc`?

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4939,82 @@ sycl::event engine_ext::async_rnn_backward(
...
 // DPCT_LABEL_END
+
+// DPCT_LABEL_BEGIN|dropout_get_state_size|dpct::dnnl
+// DPCT_DEPENDENCY_BEGIN
+// DnnlUtils|dropout_get_state_size_decl
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+inline
+size_t engine_ext::dropout_get_state_size(){
+  auto rng_engine = oneapi::mkl::rng::philox4x32x10(*_q, 0ull);
```
ziranzha: same as above

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4939,82 @@ sycl::event engine_ext::async_rnn_backward(
...
 }
 // DPCT_LABEL_END
+
+// DPCT_LABEL_BEGIN|dropout_get_state_size|dpct::dnnl
+// DPCT_DEPENDENCY_BEGIN
+// DnnlUtils|dropout_get_state_size_decl
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+inline
+size_t engine_ext::dropout_get_state_size(){
```
ziranzha: How about use get_dropout_state_size?

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4939,82 @@ sycl::event engine_ext::async_rnn_backward(
...
+  if (src_desc.get_desc().get_data_type() != ::dnnl::memory::data_type::s32) {
+    cache = allocate(rng_data_desc);
+  }
+  oneapi::mkl::rng::generate(*desc.get_distr(), *desc.get_rng_engine(),
+                             rng_data_desc.get_element_num(), (std::int32_t*)cache);
+  oneapi::mkl::rng::save_state(*desc.get_rng_engine(), (std::uint8_t *)desc.get_state());
+
+  async_reorder(scale_factor, rng_data_desc, cache, 0.f, src_desc, workspace);
+
+  return async_binary(binary_op::mul, 1.f, src_desc, src, 1.f, src_desc,
```
ziranzha: Need free cache before function finished.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -742,6 +745,66 @@ public:
...
+// DPCT_LABEL_BEGIN|dropout_desc|dpct::dnnl
+// DPCT_DEPENDENCY_BEGIN
+// DnnlUtils|dropout_desc_def
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+/// A class holding description for a Dropout operation.
+class dropout_desc {
+  float _p = 0.5f;
+  unsigned long long _seed = 0;
+  std::shared_ptr<oneapi::mkl::rng::philox4x32x10> _rng_engine;
```
ziranzha: Use a *_impl class to wrap all data members that need consistent between different desc (like sycl::queue).

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -742,6 +745,66 @@ public:
...
+  std::shared_ptr<oneapi::mkl::rng::bernoulli<std::int32_t>> _distr;
+  std::shared_ptr<std::vector<std::uint8_t>> _state_host = nullptr;
+  void *_state = nullptr;
+  void generate(sycl::queue *q, std::int64_t state_size, std::int64_t num, void *buffer) {
```
ziranzha: How about rename `state_size` to `required_state_size`

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -2147,6 +2266,54 @@ public:
...
 // DPCT_LABEL_END
 
+// DPCT_LABEL_BEGIN|dropout_desc_def|dpct::dnnl
+// DPCT_DEPENDENCY_EMPTY
+// DPCT_CODE
+inline
+void dropout_desc::restore(engine_ext &engine, float p, void *state,
+                                  size_t state_size, unsigned long long seed) {
+  _p = p;
+  _seed = seed;
```
ziranzha: Check whether `p` and `seed` will be updated or not in cudnn when `state` is nullptr.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -742,6 +745,78 @@ public:
...
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+/// A class holding description for a Dropout operation.
+class dropout_desc {
+  struct dropout_desc_imp {
+    float _p = 0.5f;
+    unsigned long long _seed = 1;
+    void *_state = nullptr;
+    std::vector<std::uint8_t> _host_state;
+    std::shared_ptr<oneapi::mkl::rng::philox4x32x10> _rng_engine = nullptr;
```
ziranzha: How about use a object instead of a shared_ptr?

ziranzha (approved):

clang/test/dpct/dnn/dropout.cu
mgouicem, line 1: Why the .cu extension? IIUC, the convention is for files containing cuda kernels to have a .cu extension.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4962,122 @@ sycl::event engine_ext::async_rnn_backward(
...
+      dst_desc.get_desc());
+
+  auto e = execute_primitive(primitive, execution_args, {});
+
+  if (cache != workspace) {
+    _q->submit([&](sycl::handler &cgh) {
+      cgh.depends_on(e);
+      cgh.host_task([=] { sycl::free(cache, *_q); });
+    });
+  }
```
mgouicem, line 5045: I am a bit confused: do you assume that the queue is in-order or not?
Asking because between desc.generate (L5020) and execute_primitive (L5038), you did not have any sync on event, whereas you have one between execute_primitive (L5038) and the host_task to free the cache (L5042)

mgouicem (approved):

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -2147,6 +2295,49 @@ public:
...
+  _imp->_p = p;
+  if (state) {
+    std::int64_t required_state_size = engine.get_dropout_state_size();
+    if (state_size < required_state_size) {
+      throw std::runtime_error("set: no sufficient memory to save states.");
+    }
+    sycl::queue *q = engine.get_queue();
+    _imp->_seed = seed;
+    _imp->_state = state;
+    _imp->_host_state = std::vector<std::uint8_t>(required_state_size);
```
mgouicem, line 2333: could you clarify why the host_state is needed? From cuDNN API, it seems that all pointers to state are on device.
And from onemkl doc, it seems that if you create the rng_engine for a gpu device, the state pointer should be a device pointer as well (not sure though).

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
```diff
@@ -4771,6 +4962,122 @@ sycl::event engine_ext::async_rnn_backward(
...
+                                              size_t workspace_size) {
+  if (workspace_size < src_desc.get_size()) {
+    throw std::runtime_error("async_dropout_forward: no sufficient workspace.");
+  }
+  float p = desc.get_probability();
+  if (p == 1.f) {
+    return _q->memset(dst, 0, dst_desc.get_size());
+  } else if (p == 0.f) {
+    return async_reorder(1.f, src_desc, src, 0.f, dst_desc, dst);
+  }
```
mgouicem, line 5008: Are you sure of the order here? I would expect the if conditions to be swapped as
in general a Bernoulli distribution returns 1 with probability p. (see [onemkl doc](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/oneapi-mkl-rng-bernoulli.html#GUID-7E457D80-1FEA-4E4E-B4C9-C3AFA7490A99)).
So if p == 1, the mask is all ones so the src data should be returned.
if p == 0, the mask is all zeros so dst is set to 0.

clang/test/dpct/dnn/dropout.cu
intwanghao: Based on [nvcc doc](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#supported-input-file-suffixes) , the .cu extension used for CUDA source file, containing host code and device functions. So, we will use .cu extension, if the file containing CUDA semantics.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
intwanghao: Yes, In order to align with CUDA, we use in-order queue.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
intwanghao: Confirmed with oneMKL team, the state pointer should be a host pointer even the random generator engine is created for gpu device.

clang/runtime/dpct-rt/include/dnnl_utils.hpp.inc
intwanghao: Yes, I'm sure. For cuDNN, the value of dropout means the probability with which the value from input is set to zero during the dropout layer. So, the p in cuDNN is equivalent to (1 - p) in cuDNN.

zhimingwang36 (approved):
LGTM

PR #679, opened 2023-02-27
[SYCLomatic] Ensure semicolon is not removed when removing a call not surrounded by curly braces

jzc:
Currently, code removed using the `REMOVE_API_FACTORY_ENTRY` rewriter have the same behavior of removing a semicolon after the expression. Example:
```c++
if(x) cudnnDestroy(x);
std::cout << "cleanup complete.\n"
```
The migrated code will be 
```c++
if(x) std::cout << "cleanup complete.\n"
```
Now, the semicolon is not removed, giving the following migration:
```c++
if(x) ;
std::cout << "cleanup complete.\n"
```
---
The original replacement generated by ExprAnalysis does not remove the semicolon, but the call of `ReplaceStmt::getReplacement` from 

https://github.com/oneapi-src/SYCLomatic/blob/77ef8ab0126e6d0ff737f2ae1f4cb27c89aae7b7/clang/lib/DPCT/MigrationAction.cpp#L189

modifies the original replacement to remove the semicolon. This is because the returned replacement is from 

https://github.com/oneapi-src/SYCLomatic/blob/77ef8ab0126e6d0ff737f2ae1f4cb27c89aae7b7/clang/lib/DPCT/TextModification.cpp#L75-L76 

where `removeStmtWithCleanups` has the logic for removing the semicolon. The PR changes the intialization of `IsCleanup` so that when the statement being removed is not in a compound statement, the semicolon is not removed.

### DIFF
--- a/clang/lib/DPCT/TextModification.cpp
+++ b/clang/lib/DPCT/TextModification.cpp
@@ -24,6 +24,14 @@ using namespace clang;
 using namespace clang::dpct;
 using namespace clang::tooling;
 
+bool ReplaceStmt::inCompoundStmt(const Stmt *E) {
+  auto &context = DpctGlobalInfo::getContext();
+  const auto parents = context.getParents(*E);
+  return std::find_if(parents.begin(), parents.end(), [](const DynTypedNode &n) {
+    return bool(n.get<CompoundStmt>());
+  }) != parents.end();
+}
+
 std::shared_ptr<ExtReplacement>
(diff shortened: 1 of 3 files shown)


ziranzha:
For code like
```
auto ret = CALL(...);
```
Will the semicolon will also be removed? If so, I think it is not expected.

ziranzha (approved):

zhimingwang36 (approved):
LGTM

PR #724, opened 2023-03-07
[SYCLomatic] dpct::group::reduce_over_partial_group

danhoeflinger:
New function `dpct::group::reduce_over_partial_group` which provides the functionality to reduce over part of a subgroup within a sycl kernel.  
Returns to each work item in the subgroup the reduction of the first `items_to_reduce` elements of the subgroup using `binary_op`.  Requires the specification of the `identity` for the type `T` and binary operation, `binary_op`.


clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h.inc
```diff
@@ -554,6 +554,31 @@ reduce(Item item, T (&inputs)[VALUES_PER_THREAD], BinaryOperation binary_op) {
...
+/// Perform a reduction on a limited number of the work items in a subgroup
+///
+/// \param item A work-item in a group.
+/// \param value value per work item which is to be reduced
+/// \param items_to_reduce num work items at the start of the subgroup to reduce
+/// \param identity identity for type and binary_op
+/// \param binary_op functor that implements the binary operation used to
+/// perform the scan. \returns value of the reduction using binary_op
+template <typename Item, typename T, class BinaryOperation>
+__attribute__((always_inline)) T
```
yihwang-nv: Can we use `inline` here? `_attribute__((always_inline))` is a GCC/Clang extension

danhoeflinger: Sure,  I was just following the convention of the surrounding functions.  Should we adjust those as well?

yihwang-nv: Ah, let's still following the convention of the surrounding functions.

yihwang-nv (approved):
LGTM

clang/test/dpct/helper_files_ref/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -501,6 +501,25 @@ reduce(Item item, T (&inputs)[VALUES_PER_THREAD], BinaryOperation binary_op) {
...
+///
+/// \param item A work-item in a group.
+/// \param value value per work item which is to be reduced
+/// \param items_to_reduce num work items at the start of the subgroup to reduce
+/// \param identity identity for type and binary_op
+/// \param binary_op functor that implements the binary operation used to
+/// perform the scan. \returns value of the reduction using binary_op
+template <typename Item, typename T, class BinaryOperation>
+__attribute__((always_inline)) T
+reduce_over_partial_group(const Item &item, const T &value,
```
zhimingwang36, line 511: Please give more info on how to use identity,  give example like for min/max/sum op.
the limitation:  how to set idnetity if the caller don't know the detail of BinaryOperation.

clang/test/dpct/helper_files_ref/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -501,6 +501,25 @@ reduce(Item item, T (&inputs)[VALUES_PER_THREAD], BinaryOperation binary_op) {
...
+/// Perform a reduction on a limited number of the work items in a subgroup
+///
+/// \param item A work-item in a group.
+/// \param value value per work item which is to be reduced
+/// \param items_to_reduce num work items at the start of the subgroup to reduce
+/// \param identity identity for type and binary_op
+/// \param binary_op functor that implements the binary operation used to
+/// perform the scan. \returns value of the reduction using binary_op
+template <typename Item, typename T, class BinaryOperation>
+__attribute__((always_inline)) T
```
zhimingwang36: maybe can use __dpct_inline__  which currently defined dpct.hpp.

clang/test/dpct/helper_files_ref/include/dpl_extras/dpcpp_extensions.h
danhoeflinger: I've updated the function in a way which should clear this up.  The sycl spec only supports `reduce_over_group` for the same set of functions for which `sycl::has_known_identity` is true and `sycl::known_identity` exists. Specifically from the table below.  

I removed the identity argument and instead use the identity which is supplied by sycl and only enable the function when `sycl::has_known_identity` is true.  If we need support for reductions for types / BinaryOperation combinations not in the table below, that would require a different function and it would require local memory to be provided for the reduction operation.


<html>
<body>


Operator | Available Only When | Identity
-- | -- | --
sycl::plus | std::is_arithmetic_v<AccumulatorT> | AccumulatorT{}
sycl::multiplies | std::is_arithmetic_v<AccumulatorT> | AccumulatorT{1}
sycl::bit_and | std::is_integral_v<AccumulatorT> | ~AccumulatorT{}
sycl::bit_or | std::is_integral_v<AccumulatorT> | AccumulatorT{}
sycl::bit_xor | std::is_integral_v<AccumulatorT> | AccumulatorT{}
sycl::logical_and | std::is_same_v<std::remove_v_t<AccumulatorT>, bool> | true
sycl::logical_or | std::is_same_v<std::remove_v_t<AccumulatorT>, bool> | false
sycl::minimum | std::is_integral_v<AccumulatorT> | std::numeric_limits<AccumulatorT>::max()
sycl::minimum | std::is_floating_point_v<AccumulatorT> \|\|     std::is_same_v<std::remove_cv_t<AccumulatorT>, sycl::half> | std::numeric_limits<AccumulatorT>::infinity()
sycl::maximum | std::is_integral_v<AccumulatorT> | std::numeric_limits<AccumulatorT>::lowest()
sycl::maximum | std::is_floating_point_v<AccumulatorT> \|\|     std::is_same_v<std::remove_cv_t<AccumulatorT>, sycl::half> | -std::numeric_limits<AccumulatorT>::infinity()


</body>
</html>

clang/test/dpct/helper_files_ref/include/dpl_extras/dpcpp_extensions.h
danhoeflinger: I've updated this whole file to now use `__dpct_inline__` instead of `__attribute__((always_inline))` or `inline`.

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h.inc
```diff
@@ -80,7 +80,7 @@ constexpr auto __inclusive_scan_over_group(_Args... __args) {
 /// perform the scan.
 template <typename Item, typename T, class BinaryOperation,
           int VALUES_PER_THREAD>
-__attribute__((always_inline)) void
+__dpct_inline__ void
```
yihwang-nv, line 89: `__dpct_inline` defined in `dpct/dpct.hpp`, so ,I think we should include `dpct/dpct.hpp` in this file, WDYT?

clang/test/dpct/helper_files_ref/include/dpl_extras/dpcpp_extensions.h
yihwang-nv: This looks good! In my opinion, if we can't find a correct identifiy for the binop, we can just emit a warning in migration progress, to tell users that they should manually specify an identify. So, I think, we can remove the `sycl::has_known_identity` constraint, WDYT? @danhoeflinger @zhimingwang36

danhoeflinger: @yihanwg The `sycl::reduce_over_group` function itself requires that the binary op / type be from this set.  Even if the user specified an identity, we would not be able to use it.  
To handle other cases, we would need to write our own reduction code and it would require local memory to be passed in.

danhoeflinger: I suppose (as used below in this same file) there is a `sycl::ext::oneapi::experimental::reduce_over_group_impl()` we can use which allows for user defined reductions, but we would still need to create local memory and pass it in in that scenario via the grouphelper scratchpad.

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h.inc
danhoeflinger: updated to include `../dpct.hpp`

zhimingwang36 (approved):
LGTM

PR #755, opened 2023-03-20
[SYCLomatic] Migrate cudaCpuDeviceId in cudaMemPrefetchAsync

### LINKED ISSUE #569: cudaCpuDeviceId isn't translated
When the symbolic constant `cudaCpuDeviceId` is used as device identifier in CUDA calls, it isn't always translated into the right device selector.

Inside `cudaMemAdvise()`, using `cudaCpuDeviceId` will trigger the use of `cpu_device()`  as device selector. However, inside another function like `cudaMemPrefetchAsync()`, using `cudaCpuDeviceId` won't trigger any specific behavior and generate 
`get_device(cudaCpuDeviceId)` as device selector. Subsequently, this SYCL code will fail to compile as `cudaCpuDeviceId` won't be recognized by the compiler.

For reproducing, please use the following code:

```
#include <cuda.h>
void foo(void *dst, int  size) {
    cudaMemPrefetchAsync(dst, size, cudaCpuDeviceId, 0);
}
void bar(const void *devPtr, int size) {
    cudaMemAdvise(devPtr, size, cudaMemAdviseSetPreferredLocation, cudaCpuDeviceId);
}
void baz(const void *devPtr, int size) {
    cudaMemAdvise(devPtr, size, cudaMemAdviseSetPreferredLocation, 1);
}
```

This will generate:

```
#include <sycl/sycl.hpp>
#include <dpct/dpct.hpp>
void foo(void *dst, int  size) {
    dpct::dev_mgr::instance()
        .get_device(cudaCpuDeviceId) //<== this should probably be ".cpu_device()" instead 
        .default_queue()
        .prefetch(dst, size);
}
void bar(const void *devPtr, int size) {
    dpct::cpu_device().default_queue().mem_advise(devPtr, size, 0);
}
void baz(const void *devPtr, int size) {
    dpct::get_device(1).default_queue().mem_advise(devPtr, size, 0);
}
```

I guess that the problem is larger than that and that `cudaCpuDeviceId` should somehow be recognized globally for what it is and translated so that a call to `get_device(translation_of_cudaCpuDeviceId)` would be correctly handled.

0x12CC:

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -11024,11 +11024,14 @@ void MemoryMigrationRule::prefetchMigration(
 
     // In clang "define NULL __null"
     if (StmtStrArg3 == "0" || StmtStrArg3 == "") {
+      const auto Prefix =
+          MapNames::getDpctNamespace() +
+          (StmtStrArg2 == "cudaCpuDeviceId"
+               ? +"cpu_device()"
+               : "dev_mgr::instance().get_device(" + StmtStrArg2 + ")");
       requestFeature(HelperFeatureEnum::Device_dev_mgr_get_device, C);
       requestFeature(HelperFeatureEnum::Device_device_ext_default_queue, C);
-      Replacement = MapNames::getDpctNamespace() +
-                    "dev_mgr::instance().get_device(" + StmtStrArg2 +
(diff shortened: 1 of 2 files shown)


tomflinda (approved):
LGTM

jzc (approved):

PR #807, opened 2023-04-07
[DOC] Update notices & disclaimers

krisdale:
Move notices/disclaimers to footer.
Update notice with 0BSD license info, remove non-applicable notices.


zhimingwang36 (approved):
LGTM

PR #819, opened 2023-04-11
[SYCLomatic] moving bfe() to public namespace, enabling unsigned types 

danhoeflinger:
Moving `bfe()` from internal namepace to public `dpct::bfe()` namespace, and enabling all unsigned integers as template input.

Tests found in (https://github.com/oneapi-src/SYCLomatic-test/pull/289)

*Rebased after .inc removal*

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
@@ -18,6 +18,7 @@
 
 #include "../dpct.hpp"
 #include "functional.h"
+
 namespace dpct {
 namespace group {
 namespace detail {
--- a/clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
@@ -17,20 +17,9 @@
 #endif
 
 #include "../dpct.hpp"
-
+#include "functional.h"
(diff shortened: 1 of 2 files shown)


tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -162,12 +162,6 @@ template <int N, int COUNT> struct log2<N, 0, COUNT> {
   enum { VALUE = (1 << (COUNT - 1) < N) ? COUNT : COUNT - 1 };
 };
 
-__dpct_inline__ uint32_t bfe(uint32_t source, uint32_t bit_start,
```
zhiweij1, line 165: Do we need keep backward compatibility here?
Can we keep the interface, call the ::bfe in its function body and emit a deprecated message to hint user to use ::bfe?

yihwang-nv: This `bfe` defined in the `detail` namespace, I'm not sure should we keep the backward compatibility in `detail` namespace.

zhiweij1: Oh, you are right. I did not notice that `bfe` is in the `detail` namespace.
I will +1 for this PR.

zhiweij1 (approved):

PR #823, opened 2023-04-12
[SYCLomatic]Fix an inappropriate migration that uses 'auto' as the type of cast

wenhuiNi:


zhimingwang36 (approved):
LGTM

wenhuiNi:
Using isDependentType() will result in misjudging normal template types.

ziranzha (approved):

PR #844, opened 2023-04-21
[SYCLomatic] Fix a compilation error when using gcc 11.x.

tangjj11:

### DIFF
--- a/clang/lib/DPCT/CallExprRewriter.h
+++ b/clang/lib/DPCT/CallExprRewriter.h
@@ -1280,7 +1280,7 @@ public:
   DerefExprRewriter(
       const CallExpr *C, StringRef Source,
       const std::function<ArgValueT(const CallExpr *)> &ArgCreator)
-      : PrinterRewriter<DerefExpr>(C, Source, ArgCreator(C)) {}
+      : PrinterRewriter<dpct::DerefExpr>(C, Source, ArgCreator(C)) {}
 };
 
 class SubGroupPrinter {


tangjj11:
The minimum reproduce case (will build failed on newest version of gcc/clang/msvc):
```c++
template <class T = int *>
class A1 {
public:
  A1(const int *E) {}
};

class A2 {
public:
  A2(const int *E) {}
};

template <class T> class B : T {
public:
  template <class... ArgsT>
  B(ArgsT &&...Args) : T(std::forward<ArgsT>(Args)...) {}
};

template <class T> class C : public B<A1<T>>, public B<A2/*first*/> { // public B<A2/*first*/> pass.
public:
  C(const T Arg) : B<A1<T>>(Arg), B<A2/*second*/>(Arg) {} // B<A1<T>>(Arg) pass, B<A2/*second*/>(Arg) error.
};

void f() {
  int *x;
  C<int *> E(x);
}
```
out put:
```
error: 'class A2 A2::A2' is private within this context
   22 |   C(const T Arg) : B<A1<T>>(Arg), B<A2>(Arg) {} // B<A1<T>>(Arg) pass, B<A2>(Arg) error.
      |          
```
Reason (analyze the code in clang):
The Name look up order: self class body -->base class body --> outer scope...
1. For the second `A2`, the `Sema::LookupQualifiedName()` in the `Sema::CPPLookupName` will find the target (A2::A2) in base class body A2, and will stop searching. Then the access check will failed because the class B inherited class A2 in private mode, and second `A2` in the class C body cannot see the symbol in class A2 body.
2. For the first `A2`, the `Sema::LookupQualifiedName()` will find nothing, because the expr is out of the class C body. And the `Sema::CPPLookupName` will continue searching outer scope and find the public class A2. So it is accessible.

wenhuiNi (approved):

ziranzha (approved):

PR #877, opened 2023-04-28
[SYCLomatic] segmented reduce argmin/max

danhoeflinger:
Functional implementation of `dpct::segmented_reduce_argmin` and `dpct::segmented_reduce_argmax`. 
Tested by https://github.com/oneapi-src/SYCLomatic-test/pull/310

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
@@ -2066,6 +2066,80 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
 }
 // DPCT_LABEL_END
 
+// DPCT_LABEL_BEGIN|segmented_reduce_argmin|dpct
+// DPCT_DEPENDENCY_BEGIN
+// DplExtrasIterators|arg_index_input_iterator
+// DplExtrasIterators|key_value_pair
+// DplExtrasVector|is_iterator
+// DPCT_DEPENDENCY_END
+// DPCT_CODE
+template <typename _ExecutionPolicy, typename key_t, typename key_out_t, 
+          typename OffsetIteratorT>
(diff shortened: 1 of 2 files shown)


clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2066,6 +2066,80 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
+      if (segment_length <= 0) {
+        *(keys_out + i) = dpct::key_value_pair(
+            ptrdiff_t(1),
+            ::std::numeric_limits<
+                typename ::std::iterator_traits<key_t>::value_type>::max());
+      } else {
+        dpct::arg_index_input_iterator arg_index(keys_in + segment_begin);
+        *(keys_out + i) = *::std::min_element(
+            arg_index, arg_index + segment_length,
+            [](const auto &a, const auto &b) { return a.value < b.value; });
```
yihwang-nv, line 2098: Can we use std::less, what do you think?

danhoeflinger: We can't just use `::std::less` or omit a comparator for the default because we are comparing dereferenced `dpct::arg_index_input_iterator`  (`dpct::key_value_pair`) by their `value` member variable. 

I suppose we could instead do a `::std::min_element` on `keys_in` using the defaulted `std::less`, and then construct a key value pair using `std::distance`.  I don't have a strong preference either way, but the current implementation seems simpler to me.

yihwang-nv: Sorry, I didn't notice this. Current implementation looks good to me.

yihwang-nv (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #878, opened 2023-05-02
[SYCLomatic] equal_range via binary_search

danhoeflinger:
Switching implementation to use serial (device) binary search rather than parallel transform reduce.
This takes advantage of the ordered input, and should have better performance.


zhimingwang36 (approved):
LGTM

PR #884, opened 2023-05-04
[SYCLomatic] Make device_vector AllocatorAwareContainer 

danhoeflinger:
Changes to make `dpct::device_vector` allocator aware, while still relying upon a USM shared memory allocator.

This should satisfy the requirements of [AllocatorAwareContainer](https://en.cppreference.com/w/cpp/named_req/AllocatorAwareContainer).  

With the underlying data being host accessible, we can fully satisfy AllocatorAwareContainer.

Tested by https://github.com/oneapi-src/SYCLomatic-test/pull/314.

*Rebased after .inc removal*

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
@@ -73,7 +73,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static
       typename ::std::enable_if_t<__has_construct<_Allocator, T *>::value, void>
       uninitialized_value_construct_n(_Allocator alloc, T *p, Size n) {
-    assert( p != nullptr && "value constructing null data");
+    assert(p != nullptr && "value constructing null data");
     for (Size i = 0; i < n; i++) {
       ::std::allocator_traits<_Allocator>::construct(alloc, p + i);
     }
@@ -83,7 +83,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static typename ::std::enable_if_t<!__has_construct<_Allocator, T *>::value,
                                      void>
   uninitialized_value_construct_n(_Allocator alloc, T *p, Size n) {
-    assert( p != nullptr && "value constructing null data");
+    assert(p != nullptr && "value constructing null data");
     ::std::uninitialized_value_construct_n(
         oneapi::dpl::execution::make_device_policy(get_default_queue()), p, n);
   }
@@ -92,7 +92,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static typename ::std::enable_if_t<
       __has_construct<_Allocator, T *, const Value &>::value, void>
   uninitialized_fill_n(_Allocator alloc, T *first, Size n, const Value &value) {
-    assert( first != nullptr && "filling null data");
+    assert(first != nullptr && "filling null data");
     for (Size i = 0; i < n; i++) {
       ::std::allocator_traits<_Allocator>::construct(alloc, first + i, value);
     }
@@ -102,7 +102,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static typename ::std::enable_if_t<
       !__has_construct<_Allocator, T *, const Value &>::value, void>
   uninitialized_fill_n(_Allocator alloc, T *first, Size n, const Value &value) {
-    assert( first != nullptr && "filling null data");
+    assert(first != nullptr && "filling null data");
     ::std::uninitialized_fill_n(
         oneapi::dpl::execution::make_device_policy(get_default_queue()), first,
         n, value);
@@ -111,7 +111,7 @@ template <typename _Allocator> struct device_allocator_traits {
   template <typename Iter1, typename Size, typename T>
   static void uninitialized_custom_copy_n(_Allocator alloc, Iter1 first, Size n,
                                           T *d_first) {
-    assert( d_first != nullptr && "copying into null data");
+    assert(d_first != nullptr && "copying into null data");
     for (Size i = 0; i < n; i++) {
       ::std::allocator_traits<_Allocator>::construct(alloc, d_first + i,
                                                      *(first + i));
@@ -126,7 +126,7 @@ template <typename _Allocator> struct device_allocator_traits {
       void>
   uninitialized_device_copy_n(_Allocator alloc, Iter1 first, Size n,
                               T *d_first) {
-    assert( d_first != nullptr && "copying into null data");
+    assert(d_first != nullptr && "copying into null data");
     uninitialized_custom_copy_n(alloc, first, n, d_first);
   }
 
@@ -138,7 +138,7 @@ template <typename _Allocator> struct device_allocator_traits {
       void>
   uninitialized_device_copy_n(_Allocator alloc, Iter1 first, Size n,
                               T *d_first) {
-    assert( d_first != nullptr && "copying into null data");
+    assert(d_first != nullptr && "copying into null data");
     ::std::uninitialized_copy_n(
         oneapi::dpl::execution::make_device_policy(get_default_queue()), first,
         n, d_first);
@@ -151,7 +151,7 @@ template <typename _Allocator> struct device_allocator_traits {
           typename ::std::iterator_traits<Iter1>::value_type>::value,
       void>
   uninitialized_host_copy_n(_Allocator alloc, Iter1 first, Size n, T *d_first) {
-    assert( d_first != nullptr && "copying into null data");
+    assert(d_first != nullptr && "copying into null data");
     uninitialized_custom_copy_n(alloc, first, n, d_first);
   }
 
@@ -162,7 +162,7 @@ template <typename _Allocator> struct device_allocator_traits {
           typename ::std::iterator_traits<Iter1>::value_type>::value,
       void>
   uninitialized_host_copy_n(_Allocator alloc, Iter1 first, Size n, T *d_first) {
-    assert( d_first != nullptr && "copying into null data");
+    assert(d_first != nullptr && "copying into null data");
     ::std::uninitialized_copy_n(first, n, d_first);
   }
 
@@ -170,7 +170,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static
       typename ::std::enable_if_t<__has_destroy<_Allocator, T *>::value, void>
       destroy_n(_Allocator alloc, T *p, Size n) {
-    assert( p != nullptr && "destroying null data");
+    assert(p != nullptr && "destroying null data");
     for (Size i = 0; i < n; i++) {
       ::std::allocator_traits<_Allocator>::destroy(alloc, p + i);
     }
@@ -180,7 +180,7 @@ template <typename _Allocator> struct device_allocator_traits {
   static
       typename ::std::enable_if_t<!__has_destroy<_Allocator, T *>::value, void>
       destroy_n(_Allocator alloc, T *p, Size n) {
-    assert( p != nullptr && "destroying null data");
+    assert(p != nullptr && "destroying null data");
     ::std::destroy_n(
         oneapi::dpl::execution::make_device_policy(get_default_queue()), p, n);
   }
@@ -231,20 +231,18 @@ private:
   }
 
   template <typename Iter>
-  void _construct(Iter first, Iter last, size_type start_idx = 0) {
-    size_type num_eles = ::std::distance(first, last);
-    if (num_eles > 0) {
+  void _construct_iter(Iter first, size_type n, size_type start_idx = 0) {
+    if (n > 0) {
       device_allocator_traits<Allocator>::uninitialized_device_copy_n(
-          _alloc, first, num_eles, _storage + start_idx);
+          _alloc, first, n, _storage + start_idx);
     }
   }
 
   template <typename Iter>
-  void _construct_host(Iter first, Iter last, size_type start_idx = 0) {
-    size_type num_eles = ::std::distance(first, last);
-    if (num_eles > 0) {
+  void _construct_iter_host(Iter first, size_type n, size_type start_idx = 0) {
+    if (n > 0) {
       device_allocator_traits<Allocator>::uninitialized_host_copy_n(
-          _alloc, first, num_eles, _storage + start_idx);
+          _alloc, first, n, _storage + start_idx);
     }
   }
 
@@ -252,6 +250,38 @@ private:
     if (n > 0) {
       device_allocator_traits<Allocator>::destroy_n(_alloc,
                                                     _storage + start_idx, n);
+    }
+  }
+
+  void _assign_elements(const device_vector &other) {
+    if (other.size() <= _size) {
+      // if incoming elements fit within existing elements, copy then destroy
+      // excess
+      ::std::copy(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()),
+          other.begin(), other.end(), begin());
+      resize(other.size());
+    } else if (other.size() < _capacity) {
+      // if incoming elements don't fit within existing elements but do fit
+      // within total capacity
+      // copy elements that fit, then use uninitialized copy to ge the rest and
+      // adjust size
+      std::copy_n(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()),
+          other.begin(), _size, begin());
+      device_allocator_traits<Allocator>::uninitialized_device_copy_n(
+          _alloc, other.begin() + _size, other.size() - _size,
+          _storage + _size);
+      _size = other.size();
+    } else {
+      // If incoming elements exceed current capacity, destroy all existing
+      // elements, then allocate incoming vectors capacity, then use
+      // uninitialized copy
+      clear();
+      reserve(other.capacity());
+      device_allocator_traits<Allocator>::uninitialized_device_copy_n(
+          _alloc, other.begin(), other.size(), _storage);
+      _size = other.size();
     }
   }
 
@@ -299,7 +329,7 @@ public:
   device_vector(device_vector &&other, const Allocator &alloc)
       : _alloc(alloc), _size(other.size()), _capacity(other.capacity()) {
     _storage = alloc_traits::allocate(_alloc, _capacity);
-    _construct(other.begin(), other.end()); // ok to parallelize
+    _construct_iter(other.begin(), _size); // ok to parallelize
     other._size = 0;
     other._capacity = 0;
     other._storage = nullptr;
@@ -317,7 +347,7 @@ public:
     _set_capacity_and_alloc();
     // unsafe to parallelize on device as we dont know if InputIterator is valid
     // oneDPL input type
-    _construct_host(first, last);
+    _construct_iter_host(first, _size);
   }
 
   device_vector(const device_vector &other, const Allocator &alloc)
@@ -325,7 +355,7 @@ public:
     _size = other.size();
     _capacity = other.capacity();
     _storage = alloc_traits::allocate(_alloc, _capacity);
-    _construct(other.begin(), other.end());
+    _construct_iter(other.begin(), _size);
   }
 
   device_vector(const device_vector &other)
@@ -341,7 +371,7 @@ public:
     _size = other.size();
     _capacity = other.capacity();
     _storage = alloc_traits::allocate(_alloc, _capacity);
-    _construct(other.begin(), other.end());
+    _construct_iter(other.begin(), _size);
   }
 
   template <typename OtherAllocator>
@@ -369,10 +399,7 @@ public:
       _capacity = 0;
       _alloc = other._alloc;
     }
-    reserve(other.capacity());
-    resize(other.size());
-    ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                other.begin(), other.end(), begin());
+    _assign_elements(other);
     return *this;
   }
   device_vector &operator=(device_vector &&other) {
@@ -385,11 +412,8 @@ public:
       _capacity = ::std::move(other._capacity);
       _size = ::std::move(other._size);
     } else {
-      reserve(other.capacity());
-      resize(other.size());
-      ::std::copy(
-          oneapi::dpl::execution::make_device_policy(get_default_queue()),
-          other.begin(), other.end(), begin());
+      _assign_elements(other);
+      // destroy and deallocate other vector
       other.clear();
       alloc_traits::deallocate(other._alloc, other._storage, other._capacity);
     }
@@ -419,24 +443,25 @@ public:
     } else {
       // swap all elements up to the minimum size between the two vectors
       size_type min_size = ::std::min(size(), v.size());
-      for (size_type i = 0; i < min_size; i++) {
-        ::std::swap(_storage[i], v._storage[i]);
-      }
+      auto zip = oneapi::dpl::make_zip_iterator(begin(), v.begin());
+      ::std::for_each(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()), zip,
+          zip + min_size, [](auto zip_ele) {
+            std::swap(::std::get<0>(zip_ele), ::std::get<1>(zip_ele));
+          });
       // then copy the elements beyond the end of the smaller list, and resize
       if (size() > v.size()) {
-        size_type tmp_v_size = v.size();
-        v.resize(size());
-        ::std::copy(
-            oneapi::dpl::execution::make_device_policy(get_default_queue()),
-            begin() + tmp_v_size, end(), v.begin() + tmp_v_size);
-        resize(tmp_v_size);
-      } else {
-        size_type tmp_size = v.size();
-        resize(v.size());
-        ::std::copy(
-            oneapi::dpl::execution::make_device_policy(get_default_queue()),
-            v.begin() + tmp_size, v.end(), begin() + tmp_size);
-        v.resize(tmp_size);
+        v.reserve(capacity());
+        device_allocator_traits<Allocator>::uninitialized_device_copy_n(
+            _alloc, begin() + min_size, begin() + size(), size() - min_size,
+            v.begin() + min_size);
+        v._size = size();
+      } else if (size() < v.size()) {
+        reserve(v.capacity());
+        device_allocator_traits<Allocator>::uninitialized_device_copy_n(
+            _alloc, v.begin() + min_size, v.begin() + v.size(),
+            v.size() - min_size, begin() + min_size);
+        _size = v.size();
       }
     }
   }
@@ -559,10 +584,10 @@ public:
   }
   void insert(iterator position, size_type n, const T &x) {
     if (position == end()) {
-      resize(size() + n);
-      ::std::fill(
-          oneapi::dpl::execution::make_device_policy(get_default_queue()),
-          end() - n, end(), x);
+      reserve(size() + n);
+      device_allocator_traits<Allocator>::uninitialized_fill_n(
+          _alloc, _storage + size(), n, x);
+      _size += n;
     } else {
       auto i_n = ::std::distance(begin(), position);
       // allocate temporary storage
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
@@ -581,9 +581,10 @@ public:
                                    InputIterator>::type last) {
     auto n = ::std::distance(first, last);
     if (position == end()) {
-      resize(size() + n);
+      reserve(size() + n);
       // unsafe to call on device as we dont know the InputIterator type
-      ::std::copy(first, last, end() - n);
+      ::std::uninitialized_copy(first, last, end());
+      _size += n;
     } else {
       auto m = ::std::distance(position, end());
       // will throw if position is not inside active vector
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
@@ -11,6 +11,7 @@
 
 #include <oneapi/dpl/algorithm>
 #include <oneapi/dpl/execution>
+#include <oneapi/dpl/memory>
 
 #include <sycl/sycl.hpp>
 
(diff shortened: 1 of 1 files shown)


tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -121,143 +122,142 @@ private:
...
     // Move assignment operator:
-    device_vector dummy(::std::move(other));
-    this->swap(dummy);
+    if constexpr (::std::allocator_traits<Allocator>::
+                      propagate_on_container_move_assignment::value) {
+      _alloc = std::move(other._alloc);
+    }
+    _size = std::move(other._size);
+    _capacity = std::move(other._capacity);
+    _storage = std::move(other._storage);
```
ziranzha: If `std::allocator_traits<Allocator>::propagate_on_container_move_assignment::value` is false, we need use current `alloc` to allocate storage and deallocate `other`'s storage.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -275,28 +275,32 @@ public:
     ::std::swap(_size, v._size);
     ::std::swap(_capacity, v._capacity);
     ::std::swap(_storage, v._storage);
-    ::std::swap(_alloc, v._alloc);
+    if constexpr (::std::allocator_traits<
+                      Allocator>::propagate_on_container_swap::value) {
```
ziranzha: Similar as above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -121,143 +122,142 @@ private:
...
   }
   device_vector &operator=(const device_vector &other) {
     // Copy assignment operator:
-    resize(other.size());
-    if (_size > 0) {
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                other.begin(), other.end(), begin());
+    if constexpr (::std::allocator_traits<Allocator>::
+                      propagate_on_container_copy_assignment::value) {
+      _alloc = other._alloc;
```
ziranzha: Should we deallocate current storage here?

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -121,143 +122,142 @@ private:
...
+    _construct(other.begin(), other.end());
     return *this;
   }
   device_vector &operator=(device_vector &&other) {
     // Move assignment operator:
-    device_vector dummy(::std::move(other));
-    this->swap(dummy);
+    if constexpr (::std::allocator_traits<Allocator>::
+                      propagate_on_container_move_assignment::value) {
+      _alloc = std::move(other._alloc);
```
ziranzha: Should we deallocate current storage here?

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
```diff
@@ -121,143 +122,142 @@ private:
 
   void _set_capacity_and_alloc() {
     _capacity = ::std::max(_size * 2, _min_capacity());
-    _storage = _alloc.allocate(_capacity);
+    _storage = ::std::allocator_traits<Allocator>::allocate(_alloc, _capacity);
+  }
+
+  void _construct(size_type n, size_type start_idx = 0) {
+    if (n > 0) {
+      for (int i = 0; i < n; i++) {
```
ziranzha: Can we speed up the construction with dpl functions?

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
danhoeflinger: Yes, you are correct.  I'll make this and the other related changes requested below.  I have missed some aspects of destruction / deallocation, and will update accordingly.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
danhoeflinger: I'm not sure we can assume the thread safety of construct or destroy calls.  

If we could assume this thread safety, from a technical perspective we cant send these construct calls to a SYCL kernel because of the use of variadic functions.  However, if we could be sure that the construction are thread safe we could either use a `par_unseq` policy or create our own version of `std::allocator_traits` with a fixed number of possible arguments to the construct call to avoid the variadic function call allow us to use them in SYCL kernels.  I have a prototype of something like this from a different implementation of this code.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
danhoeflinger: Yes, done.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
danhoeflinger: Yes, done.

clang/runtime/dpct-rt/include/dpl_extras/vector.h.inc
danhoeflinger: I've addressed this section as well.

danhoeflinger (commented):
Thanks for the comments.  I've addressed them here in this PR, but I've marked this as a draft for now while I upgrade the testing to cover more of the code.  
Constructors and assignment operators handling of allocators needs to be tested better (especially the different scenarios of allocator settings for propagation).  Once I am finished upgrading the testing, I will convert this back to be a ready to merge PR.
In the meantime, I welcome more comments.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,206 @@ class device_vector {
...
     auto __tmp = ::std::vector<T, OtherA>(this->size());
     ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-              this->begin(), this->end(), __tmp.begin());
+                this->begin(), this->end(), __tmp.begin());
     return __tmp;
   }
-  device_vector()
-      : _alloc(get_default_queue()), _size(0), _capacity(_min_capacity()) {
+
+  device_vector(const Allocator &alloc)
```
mmichel11: Can we give this a default argument, so we don't need the additional constructor in line 125?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,206 @@ class device_vector {
...
-  device_vector(const device_vector &other) : _alloc(get_default_queue()) {
-    _size = other.size();
-    _capacity = other.capacity();
-    _storage = _alloc.allocate(_capacity);
-    ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-              other.begin(), other.end(), begin());
+
+  explicit device_vector(size_type n) : device_vector(n, get_default_queue()) {}
+
+  explicit device_vector(size_type n, const T &value, const Allocator &alloc)
```
mmichel11: Can we remove the need for the separate constructor in line 146 by having the constructor's signature be `explicit device_vector(size_type n, const T &value, const Allocator &alloc = get_default_queue())`

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,206 @@ class device_vector {
...
-    _set_capacity_and_alloc();
-    if (_size > 0) {
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                v.begin(), v.end(), this->begin());
-    }
-  }
+      : device_vector(v.begin(), v.end()) {}
+
+  template <typename OtherAllocator>
+  device_vector(::std::vector<T, OtherAllocator> &v, const Allocator &alloc)
```
mmichel11: Can `v` be taken as const reference here?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,206 @@ class device_vector {
...
   device_vector &operator=(device_vector &&other) {
     // Move assignment operator:
-    device_vector dummy(::std::move(other));
-    this->swap(dummy);
+    if constexpr (::std::allocator_traits<Allocator>::
+                      propagate_on_container_move_assignment::value) {
+      clear();
+      ::std::allocator_traits<Allocator>::deallocate(_alloc, _storage,
+                                                     _capacity);
+      _alloc = std::move(other._alloc);
```
mmichel11: `std::` -> `::std::` if we want to be consistent with the convention we use.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Yes,  fixed.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Thanks, fixed this and others in the PR.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Yes, addressed in a few places.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Yes, I fixed these and a few others, using `const Allocator &alloc = Allocator(get_default_queue())`.

mmichel11 (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,191 @@ class device_vector {
 
   void _set_capacity_and_alloc() {
     _capacity = ::std::max(_size * 2, _min_capacity());
-    _storage = _alloc.allocate(_capacity);
+    _storage = ::std::allocator_traits<Allocator>::allocate(_alloc, _capacity);
```
ziranzha: Using an alias for `::std::allocator_traits<Allocator>::` may simplify the code.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,191 @@ class device_vector {
...
-                    InputIterator>::type last)
-      : _alloc(get_default_queue()) {
-    _size = ::std::distance(first, last);
-    _set_capacity_and_alloc();
-    if (_size > 0) {
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                first, last, begin());
-    }
+  device_vector(device_vector &&other, const Allocator &alloc)
+      : _alloc(alloc), _size(other.size()), _capacity(other.capacity()),
```
ziranzha: If `other.alloc` is unequal to `alloc`, the `other.storage` can not reuse.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -258,61 +337,75 @@ class device_vector {
   void shrink_to_fit(void) {
     if (_size != capacity()) {
       size_type tmp_capacity = ::std::max(_size, _min_capacity());
```
ziranzha: If current capacity equal to `_min_capaticy()`, the following code can be jumped.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,191 @@ class device_vector {
...
+  device_vector(const ::std::vector<T, OtherAllocator> &v,
+                const Allocator &alloc)
+      : device_vector(v.begin(), v.end(), alloc) {}
+
   template <typename OtherAllocator>
   device_vector &operator=(const ::std::vector<T, OtherAllocator> &v) {
     resize(v.size());
-    if (_size > 0) {
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
+    ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: We may use the queue of allocator for the dpl algorithm.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -258,61 +337,75 @@ class device_vector {
...
       _storage = tmp;
       _capacity = tmp_capacity;
     }
   }
   void assign(size_type n, const T &x) {
     resize(n);
     if (_size > 0) {
-      ::std::fill(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                begin(), begin() + n, x);
+      ::std::fill(
```
ziranzha: Should we do `clear()` to destroy the objects at first?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -258,61 +337,75 @@ class device_vector {
...
          typename ::std::enable_if<internal::is_iterator<InputIterator>::value,
-                                 InputIterator>::type last) {
+                                   InputIterator>::type last) {
     auto n = ::std::distance(first, last);
     resize(n);
     if (_size > 0) {
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                first, last, begin());
+      ::std::copy(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: Same as above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,143 +67,191 @@ class device_vector {
...
+        ::std::allocator_traits<Allocator>::destroy(_alloc,
+                                                    _storage + start_idx + i);
+      }
+    }
   }
 
 public:
   template <typename OtherA> operator ::std::vector<T, OtherA>() const {
     auto __tmp = ::std::vector<T, OtherA>(this->size());
     ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: Here the objects construction is speeded up by dpl algorithm, but you comments said that is not make sense. So which is expected?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -258,61 +337,75 @@ class device_vector {
...
       return end();
     }
     auto m = ::std::distance(last, end());
     if (m <= 0) {
       return end();
     }
-    auto tmp = _alloc.allocate(m);
+    auto tmp = ::std::allocator_traits<Allocator>::allocate(_alloc, m);
     // copy remainder to temporary buffer.
     ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: Why copy to a temporary buffer? Just move ahead is much simpler.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -325,57 +418,62 @@ class device_vector {
...
       // allocate temporary storage
       auto m = ::std::distance(position, end());
       // will throw if position is not inside active vector
-      auto tmp = _alloc.allocate(m);
+      auto tmp = ::std::allocator_traits<Allocator>::allocate(_alloc, m);
       // copy remainder
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                position, end(), tmp);
+      ::std::copy(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: Similar as above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -325,57 +418,62 @@ class device_vector {
...
+      ::std::fill(
+          oneapi::dpl::execution::make_device_policy(get_default_queue()),
+          end() - n, end(), x);
     } else {
       auto i_n = ::std::distance(begin(), position);
       // allocate temporary storage
       auto m = ::std::distance(position, end());
       // will throw if position is not inside active vector
-      auto tmp = _alloc.allocate(m);
+      auto tmp = ::std::allocator_traits<Allocator>::allocate(_alloc, m);
```
ziranzha: Why it will throw?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Thanks, done.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: You are correct, fixed.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: I don't believe there is a way to extract the `queue` information from the allocator.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Thanks, fixed.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: We are following the convention of `std::vector`, which I believe `assign` does not destroy the previous elements, it merely copies in the value or incoming elements into the vector.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: We are following the convention of `std::vector`, which I believe `assign` does not destroy the previous elements, it merely copies in the value or incoming elements into the vector.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: This is to take advantage of the parallel algorithms, as input and output sequences may overlap if we just move ahead.

I suppose we can check if they overlap and choose which method we use accordingly.  I've done this.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: It will throw because it would be attempting to allocate a negative number of bits for the temporary.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: See above comment.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: This instance is copying elements, not constructing them.  It does not have the issues described below.

The issue with construction being sped up by a SYCL kernel or oneDPL algorithm is two-fold: 
1) Conceptually, allocators (which may be custom user-supplied overrides of `usm_allocator`) can't be assumed to have thread safe `construct` calls.  For instance, if you had a constructor which counted the number of element constructions and did not provide synchronization around the count variable, this could have issues.
2) Technically, `std::allocator_traits<T>::construct()` is a variadic function, and those cannot be used SYCL device functions.  Therefore, it cannot be called within a SYCL kernel.  As I mention before, if parallel accellerated construction is a feature which is highly requested, we can create our own version of `allocator_traits` which does not rely on variadic functions for use in this context.  However, with (1) in mind, it does not seem worthwhile to add this complexity.  

If we add this, we would need to state some rules about what is allowed in custom `construct` calls.

danhoeflinger (commented):
Thanks for the feedback, I've replied and made some changes in the code

clang/runtime/dpct-rt/include/dpl_extras/vector.h
ziranzha: So what will the dpl algorithm do when it tries to replace existed element? Do the inplace construct or assignment?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -217,31 +259,60 @@ class device_vector {
...
-      auto tmp = _alloc.allocate(2 * n);
+      auto tmp = alloc_traits::allocate(_alloc, n);
       // copy content (old buffer to new buffer)
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                begin(), end(), tmp);
-      // deallocate old memory
-      _alloc.deallocate(_storage, _capacity);
+      if (capacity() > 0) {
+        ::std::copy(
+            oneapi::dpl::execution::make_device_policy(get_default_queue()),
```
ziranzha: I think here we can not use dpl to speed up, as construction is doing here.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
ziranzha: Is it possible that the the distance is a positive number and not to large, so the allocation can be done?

danhoeflinger: Yes, in normal operation where this position *is* inside the vector, it the distance will return a reasonable positive number and the allocation will succeed.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: We are performing a copy to move existing (previously constructed) data to a new memory location, not calling the constructor here.  I don't believe any constructor is called here, and thus, using oneDPL to accelerate the copy is appropriate.

ziranzha: The new memory is not initialized, for some complex type, it may cause unexpected errors when doing copy assignment (which I think will be done inside `std::copy`).

AndyCHHuang (approved):
LGTM
According to the test result of small test example of std copy, I think std copy does not trigger copy constructor.

AndyCHHuang:
@danhoeflinger 
Please rebase the PR,

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Sorry for the delayed response here.  After doing some more experimentation with std::vector and a custom allocator, you are correct, and this use of oneDPL for speeding up this operation is dubious.  

However, I'm worried that fixing this will result in punishing the common case to account for the rare case of a custom allocator's custom constructor.  

I'm working on a idea which will use parallel `std::uninitialized_copy` when no custom constructor is specified by the allocator, but will use serial calls to the allocators custom constructor when one is specified.  (the default `usm_allocator` does not define one).  If I can get it to work, I'll take a sweep of the PR here to make sure all similar instances are handled appropriately.  

Thanks for bringing this up.

danhoeflinger:
@ziranzha  I've made some changes which should hopefully address your concerns about copying to uninitialized memory, and using oneDPL acceleration to construct memory.  

I've added the struct `device_allocator_traits<Allocator>` which acts somewhat similarly to `std::iterator_traits`, but at a slightly higher level.  If a matching custom constructor is specified by the allocator, then we sequentially call that custom constructor on all the data we are constructing, filling, copying to, etc.  This is because we have no guarantees on thread safety / device safety of such custom constructors.  

However, we do not want to slow down the normal case (without custom constructor from allocator, likely TriviallyCopyable types), for the rare case with a custom constructor.  Therefore, If no matching custom constructor is provided by the supplied allocator, then we send it to the relevant `uninitialized_fill`, `uninitialized_copy`, `uninitialized_default_construct_n` routines with a parallel execution policy.  These routines should properly handle uninitialized data.  

To pair with this I've also fixed in a few places where the constructor was incorrectly being skipped (in favor of a std::copy).  These places now correctly use the `_construct` routines, which utilize the new `device_allocator_traits<Allocator>` struct.

Please take a look and see if this resolves your concerns.  

*edit* I see a couple failures in the tests,  I'll need to look at these tomorrow.

tomflinda:
@danhoeflinger  the latest CI test failure below is introduced by this PR.
```
thrust_scatter_if RUNFAIL 
```
Pls fix this regression test before merging this PR.

danhoeflinger:
@tomflinda @mmichel11 @ziranzha I believe all issues have been addressed.  Tests are passing (most recent change is formatting only but retriggered tests).  

Take a look and merge if you are happy.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -41,6 +42,139 @@ struct is_iterator<T *> : std::true_type {};
...
+          typename ::std::iterator_traits<Iter1>::value_type>::value,
+      void>
+  uninitialized_host_copy_n(_Allocator alloc, Iter1 first, Size n, T *d_first) {
+    ::std::uninitialized_copy_n(first, n, d_first);
+  }
+
+  template <typename T, typename Size>
+  static
+      typename ::std::enable_if_t<__has_destroy<_Allocator, T *>::value, void>
+      destroy_n(_Allocator alloc, T *p, Size n) {
```
mmichel11: Do we want some assert to check that `p != nullptr` throughout these utilities or is that unnecessary here?

mmichel11 (commented):
One minor question from me about the added device allocator traits utility.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Good idea, added these asserts.

mmichel11 (approved):
The recent changes LGTM. Reapproving.

danhoeflinger:
@zhimingwang36 @tomflinda I think this is ready to merge.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -41,6 +42,150 @@ struct is_iterator<T *> : std::true_type {};
...
+//  while enabling oneDPL pstl accelleration when such custom constructors are
+//  not provided.
+template <typename _Allocator> struct device_allocator_traits {
+
+  // taken from libc++
+  template <class, class _Alloc, class... _Args>
+  struct __has_construct_impl : ::std::false_type {};
+
+  template <class _Alloc, class... _Args>
+  struct __has_construct_impl<decltype((void)std::declval<_Alloc>().construct(
```
ziranzha: I think we can use `__has_construct` directly here, like `__has_destroy`.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -41,6 +42,150 @@ struct is_iterator<T *> : std::true_type {};
...
+  template <class _Alloc, class _Pointer>
+  struct __has_destroy<_Alloc, _Pointer,
+                       decltype((void)std::declval<_Alloc>().destroy(
+                           std::declval<_Pointer>()))> : ::std::true_type {};
+  // end of taken from libc++
+
+  template <typename T, typename Size>
+  static
+      typename ::std::enable_if_t<__has_construct<_Allocator, T *>::value, void>
+      uninitialized_value_construct_n(_Allocator alloc, T *p, Size n) {
```
ziranzha: How about implement `uninitialized_value_construct_n`, `uninitialized_fill_n`, `uninitialized_device_copy_n` inside `__has_construct` to make code cleaner and simpler?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,145 +213,189 @@ class device_vector {
...
   device_vector &operator=(const device_vector &other) {
     // Copy assignment operator:
+    if constexpr (alloc_traits::propagate_on_container_copy_assignment::value) {
+      clear();
+      alloc_traits::deallocate(_alloc, _storage, _capacity);
+      _capacity = 0;
+      _alloc = other._alloc;
+    }
+    reserve(other.capacity());
     resize(other.size());
```
ziranzha: How about use `uninitialized_copy` instead of `resize` and `copy`?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,145 +213,189 @@ class device_vector {
...
+    if constexpr (alloc_traits::propagate_on_container_move_assignment::value) {
+      clear();
+      alloc_traits::deallocate(_alloc, _storage, _capacity);
+      _alloc = ::std::move(other._alloc);
+      _storage = ::std::move(other._storage);
+      _capacity = ::std::move(other._capacity);
+      _size = ::std::move(other._size);
+    } else {
+      reserve(other.capacity());
+      resize(other.size());
```
ziranzha: Same as above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -219,31 +410,58 @@ class device_vector {
...
+    } else {
+      // swap all elements up to the minimum size between the two vectors
+      size_type min_size = ::std::min(size(), v.size());
+      for (size_type i = 0; i < min_size; i++) {
+        ::std::swap(_storage[i], v._storage[i]);
+      }
+      // then copy the elements beyond the end of the smaller list, and resize
+      if (size() > v.size()) {
+        size_type tmp_v_size = v.size();
+        v.resize(size());
```
ziranzha: Same as above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -327,57 +560,59 @@ class device_vector {
   void insert(iterator position, size_type n, const T &x) {
     if (position == end()) {
       resize(size() + n);
-      ::std::fill(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                end() - n, end(), x);
+      ::std::fill(
```
ziranzha: How about use `uninitialized_fill` instead of `std::fill` and `resize`?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,145 +213,189 @@ class device_vector {
...
-        auto buf_first = oneapi::dpl::begin(buf);
-        auto buf_last = oneapi::dpl::end(buf);
-        ::std::copy(
-            oneapi::dpl::execution::make_device_policy(get_default_queue()),
-            buf_first, buf_last, begin());
-      }
-    }
+    // unsafe to parallelize on device as we dont know if InputIterator is valid
+    // oneDPL input type
+    _construct_host(first, last);
```
ziranzha: If user defined allocator is used and the allocated memory can only be accessed by device, it will not workable. Should we consider such cases?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -66,145 +213,189 @@ class device_vector {
...
+  void _construct(size_type n, const T &value, size_type start_idx = 0) {
+    if (n > 0) {
+      device_allocator_traits<Allocator>::uninitialized_fill_n(
+          _alloc, _storage + start_idx, n, value);
+    }
+  }
+
+  template <typename Iter>
+  void _construct(Iter first, Iter last, size_type start_idx = 0) {
+    size_type num_eles = ::std::distance(first, last);
```
ziranzha: I think we can accept size as the input argument to avoid call of `std::distance` since the size is known in use positions.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: I don't believe its possible to use directly because of the variable number of arguments `construct()` can accept.  The reason this works for `__has_destroy` gets to be simpler, is that we know that it only accepts a single pointer as an argument.  

They are effectively using the same trick, but because of the template variable pack, we have to utilize the additional template argument at the beginning and use the `_impl` layer rather than at the end of the template parameters with a defaulted argument.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: I'm not sure I understand this request.  How would one implement these inside of `__has_construct`?

One thing I could think that _might_ make things simpler is to convert the `::std::enable_if_t<__has_construct<_Allocator, T *>::value, void>` with different functions to `if constexpr(__has_construct<_Allocator, T *>::value)` within each of these `uninitialized_...` functions.  This would consolidate each function to a single instance, and could consolidate currently repeated assertions.  Is this what you might be asking for?

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: edit: my original response was wrong.  

Sure, we can make this change.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: We don't support those cases.  The underlying premise which enables `device_vector` to work is that we are relying on shared USM memory.  If we move away from that, it no longer works generally.

The USM_none case relies on `sycl::buffers`, (and we don't support the same functionality in that case).

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Yes, I think this can be improved, but we have a few different scenarios to cover.  
We need special cases for 
1) when the incoming size is <= existing size
`std::copy`, then destroy excess

2) when it isn't but the incoming size is <= _capacity
`std::copy_n` the already constructed elements, `unintialized_device_copy_n` the rest, then set new size

3) when incoming elements exceed current capacity
destroy all existing elements, `uninitialized_device_copy_n` all elements, set size

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Basically the same can be done as described above.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Thanks for catching this, it needs a bit of work.  We probably need special casing for different capacity vs size situations.  We also may want a special swapping parallel for in order to avoid bringing the data to the host to swap it, and to extra temporary allocation requirements.
Uninitialized fill makes sense for the excess.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: Yes, this should be fixed.  Thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: fixed, thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: fixed, thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: fixed, thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: fixed, thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: fixed, thanks.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: I did what I mentioned here above and I think I prefer it, so I kept it.  Let me know if you had something else in mind.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
```diff
@@ -326,58 +548,60 @@ class device_vector {
...
   insert(iterator position, InputIterator first,
          typename ::std::enable_if<internal::is_iterator<InputIterator>::value,
-                                 InputIterator>::type last) {
+                                   InputIterator>::type last) {
     auto n = ::std::distance(first, last);
     if (position == end()) {
       resize(size() + n);
-      ::std::copy(oneapi::dpl::execution::make_device_policy(get_default_queue()),
-                first, last, end());
+      // unsafe to call on device as we dont know the InputIterator type
```
ziranzha: I think we can also use `uninitialized_copy_n` here.

clang/runtime/dpct-rt/include/dpl_extras/vector.h
ziranzha: LGTM

clang/runtime/dpct-rt/include/dpl_extras/vector.h
danhoeflinger: done.  I looked at changing some of the other `resize` to `reserve` and `uninitialized_x`, but the others are not clear wins like this one seems to be.

ziranzha (approved):
LGTM

PR #896, opened 2023-05-09
[SYCLomatic] segmented_reduce_argmin remove n param, change return type

danhoeflinger:
Removing unnecessary parameter n from function signature.
Changing return type to use int rather than ptrdiff_t.

Fixing naming conventions and formatting. 
Minor misc. fixes.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
@@ -2053,22 +2053,21 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
 // DplExtrasVector|is_iterator
 // DPCT_DEPENDENCY_END
 // DPCT_CODE
-template <typename Policy, typename KeyT, typename KeyOutT,
-          typename OffsetIteratorT>
-inline ::std::enable_if_t<dpct::internal::is_iterator<KeyT>::value &&
-                          dpct::internal::is_iterator<KeyOutT>::value>
-segmented_reduce_argmin(Policy &&policy, KeyT keys_in, KeyOutT keys_out,
-                        ::std::int64_t nsegments, OffsetIteratorT begin_offsets,
-                        OffsetIteratorT end_offsets) {
+template <typename Policy, typename Iter1, typename Iter2, typename Iter3>
+inline ::std::enable_if_t<dpct::internal::is_iterator<Iter1>::value &&
+                          dpct::internal::is_iterator<Iter2>::value>
+segmented_reduce_argmin(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
+                        ::std::int64_t nsegments, Iter3 begin_offsets,
+                        Iter3 end_offsets) {
   policy.queue().submit([&](sycl::handler &cgh) {
     cgh.parallel_for(nsegments, [=](sycl::id<1> i) {
       if (end_offsets[i] <= begin_offsets[i]) {
         keys_out[i] = dpct::key_value_pair(
             1, ::std::numeric_limits<
-                   typename ::std::iterator_traits<KeyT>::value_type>::max());
+                   typename ::std::iterator_traits<Iter1>::value_type>::max());
       } else {
-        dpct::arg_index_input_iterator<KeyT, int> arg_index(keys_in +
-                                                            begin_offsets[i]);
+        dpct::arg_index_input_iterator<Iter1, int> arg_index(keys_in +
+                                                             begin_offsets[i]);
         keys_out[i] = *::std::min_element(
             arg_index, arg_index + +end_offsets[i] - begin_offsets[i],
             [](const auto &a, const auto &b) { return a.value < b.value; });
@@ -2086,23 +2085,22 @@ segmented_reduce_argmin(Policy &&policy, KeyT keys_in, KeyOutT keys_out,
 // DplExtrasVector|is_iterator
 // DPCT_DEPENDENCY_END
 // DPCT_CODE
-template <typename Policy, typename KeyT, typename KeyOutT,
-          typename OffsetIteratorT>
-inline ::std::enable_if_t<dpct::internal::is_iterator<KeyT>::value &&
-                          dpct::internal::is_iterator<KeyOutT>::value>
-segmented_reduce_argmax(Policy &&policy, KeyT keys_in, KeyOutT keys_out,
-                        ::std::int64_t nsegments, OffsetIteratorT begin_offsets,
-                        OffsetIteratorT end_offsets) {
+template <typename Policy, typename Iter1, typename Iter2, typename Iter3>
+inline ::std::enable_if_t<dpct::internal::is_iterator<Iter1>::value &&
+                          dpct::internal::is_iterator<Iter2>::value>
+segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
+                        ::std::int64_t nsegments, Iter3 begin_offsets,
+                        Iter3 end_offsets) {
   policy.queue().submit([&](sycl::handler &cgh) {
     cgh.parallel_for(nsegments, [=](sycl::id<1> i) {
       if (end_offsets[i] <= begin_offsets[i]) {
         keys_out[i] = dpct::key_value_pair(
             1,
             ::std::numeric_limits<
-                typename ::std::iterator_traits<KeyT>::value_type>::lowest());
+                typename ::std::iterator_traits<Iter1>::value_type>::lowest());
       } else {
-        dpct::arg_index_input_iterator<KeyT, int> arg_index(keys_in +
-                                                            begin_offsets[i]);
+        dpct::arg_index_input_iterator<Iter1, int> arg_index(keys_in +
+                                                             begin_offsets[i]);
         keys_out[i] = *::std::max_element(
             arg_index, arg_index + end_offsets[i] - begin_offsets[i],
             [](const auto &a, const auto &b) { return a.value < b.value; });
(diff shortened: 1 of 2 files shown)


clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,26 +2053,26 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
       if (segment_length <= 0) {
         *(keys_out + i) = dpct::key_value_pair(
-            ptrdiff_t(1),
-            ::std::numeric_limits<
-                typename ::std::iterator_traits<key_t>::value_type>::max());
+            1, ::std::numeric_limits<
+                   typename ::std::iterator_traits<key_t>::value_type>::max());
       } else {
-        dpct::arg_index_input_iterator arg_index(keys_in + segment_begin);
+        dpct::arg_index_input_iterator<decltype(keys_in), int> arg_index(
```
yihwang-nv: Can we use `key_t` here but not `decltype(keys_in)`, What do you think?
```suggestion
        dpct::arg_index_input_iterator<keys_t, int> arg_index(
```

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,26 +2053,26 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
 // DplExtrasVector|is_iterator
 // DPCT_DEPENDENCY_END
 // DPCT_CODE
-template <typename _ExecutionPolicy, typename key_t, typename key_out_t, 
+template <typename _ExecutionPolicy, typename key_t, typename key_out_t,
```
yihwang-nv: I think we should keep the same style and use upper camel case type name, like `_KeyT`, can we rename this type name?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,26 +2053,26 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
           typename OffsetIteratorT>
 inline ::std::enable_if_t<dpct::internal::is_iterator<key_t>::value &&
                           dpct::internal::is_iterator<key_out_t>::value>
 segmented_reduce_argmin(_ExecutionPolicy &&policy, key_t keys_in,
-                        key_out_t keys_out, ::std::int64_t n,
-                        ::std::int64_t nsegments, OffsetIteratorT begin_offsets,
+                        key_out_t keys_out, ::std::int64_t nsegments,
```
yihwang-nv: Seems qualifier for `std` isn't required, can we remove extra `::`, What do you think?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,26 +2053,26 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
-            ptrdiff_t(1),
-            ::std::numeric_limits<
-                typename ::std::iterator_traits<key_t>::value_type>::max());
+            1, ::std::numeric_limits<
+                   typename ::std::iterator_traits<key_t>::value_type>::max());
       } else {
-        dpct::arg_index_input_iterator arg_index(keys_in + segment_begin);
+        dpct::arg_index_input_iterator<decltype(keys_in), int> arg_index(
+            keys_in + segment_begin);
         *(keys_out + i) = *::std::min_element(
```
yihwang-nv: Seems we are missing an array subscript operator in `dpct::arg_index_input_iterator`, can we add `operator[]`? And then here may become more simply. WDYT?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,26 +2053,26 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
                           dpct::internal::is_iterator<key_out_t>::value>
 segmented_reduce_argmin(_ExecutionPolicy &&policy, key_t keys_in,
-                        key_out_t keys_out, ::std::int64_t n,
-                        ::std::int64_t nsegments, OffsetIteratorT begin_offsets,
+                        key_out_t keys_out, ::std::int64_t nsegments,
+                        OffsetIteratorT begin_offsets,
                         OffsetIteratorT end_offsets) {
   policy.queue().submit([&](sycl::handler &cgh) {
     cgh.parallel_for(nsegments, [=](sycl::id<1> i) {
       ::std::int64_t segment_begin = begin_offsets[i];
```
yihwang-nv: I think we can use `std::iterator_traits<OffsetIteratorT>::value_type` instead of `std::int64_t`, what do you think?
```
using OffsetT = std::iterator_traits<OffsetIteratorT>::value_type;
OffsetT segment_begin = begin_offsets[i];
OffsetT segment_end = end_offsets[i];

if (segmented_end <= segmented_begin) {
...
}
```

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
yihwang-nv: Seems other template parameter name in this file does not have a `_` prefix.

danhoeflinger: I've addressed the styling of template parameters relevant to `segmented_reduce_argmin`.  I plan to introduce a new PR to address styling on the rest of the file.

danhoeflinger: Template styling addressed.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: oneDPL's convention is to use `::` before `std::`, and also to use `::std::` before types like `int64_t` or `uint64_t`.  However, I see that this is not uniform across this file.  I will address this in a separate PR (along with template styling mentioned above).

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: As previously written, I believe we need signed types because of the subtraction and comparison with zero.  However, I have rewritten this to avoid these temporary variables altogether.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: Yes, fixed.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: Actually the `operator[]` exists, I've updated the code to use it.

danhoeflinger (commented):
Thanks for the comments.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: https://github.com/oneapi-src/SYCLomatic/pull/902 should resolve inconsistencies in the file.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: https://github.com/oneapi-src/SYCLomatic/pull/902 should resolve styling / naming inconsistencies

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,28 +2053,23 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
+            1, ::std::numeric_limits<
+                   typename ::std::iterator_traits<Iter1>::value_type>::max());
       } else {
-        dpct::arg_index_input_iterator arg_index(keys_in + segment_begin);
-        *(keys_out + i) = *::std::min_element(
-            arg_index, arg_index + segment_length,
+        dpct::arg_index_input_iterator<Iter1, int> arg_index(keys_in +
+                                                             begin_offsets[i]);
+        keys_out[i] = *::std::min_element(
+            arg_index, arg_index + +end_offsets[i] - begin_offsets[i],
```
yihwang-nv: Seems there's a redundant '+' here. And can we add a `()` around `end_offsets[i] - begin_offsets[i]`, to avoid call `arg_index_input_iterator::operator-(...)`? https://godbolt.org/z/a8v4naaTo
```suggestion
            arg_index, arg_index + (end_offsets[i] - begin_offsets[i]),
```

yihwang-nv:
Thanks for your fix, almost LGTM, but have a very little issue.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
```diff
@@ -2053,28 +2053,23 @@ inline ::std::pair<Iter1, Iter1> equal_range(_ExecutionPolicy &&policy,
...
+        keys_out[i] = dpct::key_value_pair(
+            1, ::std::numeric_limits<
+                   typename ::std::iterator_traits<Iter1>::value_type>::max());
       } else {
-        dpct::arg_index_input_iterator arg_index(keys_in + segment_begin);
-        *(keys_out + i) = *::std::min_element(
-            arg_index, arg_index + segment_length,
+        dpct::arg_index_input_iterator<Iter1, int> arg_index(keys_in +
+                                                             begin_offsets[i]);
+        keys_out[i] = *::std::min_element(
```
yihwang-nv, line 2074: I'm not sure should we pass the execution policy to `std::min_element`

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: fixed.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h.inc
danhoeflinger: No, we are just using a parallel for over segments of sequential versions for this functional mapping.

yihwang-nv (approved):
Thanks, LGTM!

zhimingwang36 (approved):
LGTM

PR #913, opened 2023-05-17
Fix for default constructor of device_pointer_base

danhoeflinger:
`sycl::buffer` has no default constructor, so we must call with an empty `sycl::range<1>(0)`.


tomflinda (approved):
LGTM

danhoeflinger:
Is there a timeout for compilation in the test suite?  `onedpl_test_segmented_sort_pairs` is failing but I see no error logs and haven't been able to reproduce myself.  I know it has a long compilation time.  

Either way, I don't believe it is related to this PR's change.

zhimingwang36 (approved):
LGTM

PR #942, opened 2023-05-25
[SYCLomatic] Remove deprecated content

intwanghao:


clang/lib/DPCT/MapNames.cpp
```diff
@@ -161,11 +161,11 @@ void MapNames::setExplicitNamespaceMap() {
       {"long4", std::make_shared<TypeNameRule>(getClNamespace() + "long4")},
       {"longlong1", std::make_shared<TypeNameRule>("int64_t")},
       {"longlong2",
-       std::make_shared<TypeNameRule>(getClNamespace() + "longlong2")},
+       std::make_shared<TypeNameRule>(getClNamespace() + "vec<std::int64_t, 2>")},
```
ziranzha: How about use `long2`? It is an aliases of `vec<std::int64_t, 2>`.

clang/lib/DPCT/MapNames.cpp
```diff
@@ -184,11 +184,11 @@ void MapNames::setExplicitNamespaceMap() {
       {"ulong4", std::make_shared<TypeNameRule>(getClNamespace() + "ulong4")},
       {"ulonglong1", std::make_shared<TypeNameRule>("uint64_t")},
       {"ulonglong2",
-       std::make_shared<TypeNameRule>(getClNamespace() + "ulonglong2")},
+       std::make_shared<TypeNameRule>(getClNamespace() + "vec<std::uint64_t, 2>")},
```
ziranzha: How about use `ulong2`? It is an aliases of `vec<std::uint64_t, 2>`.

zhimingwang36 (approved):
LGTM

PR #951, opened 2023-05-30
[SYCLomatic] Remove the using declaration if it is not std namespace

zhiweij1:

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -13873,8 +13873,22 @@ void NamespaceRule::runRule(const MatchFinder::MatchResult &Result) {
         Repl += ";";
       }
       emplaceTransformation(new ReplaceText(Beg, Len, std::move(Repl)));
-      if (UD->getNameAsString() == "max" || UD->getNameAsString() == "min") {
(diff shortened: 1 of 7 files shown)


tomflinda (approved):
LGTM

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -13826,6 +13826,9 @@ void NamespaceRule::runRule(const MatchFinder::MatchResult &Result) {
         Repl += ";";
       }
       emplaceTransformation(new ReplaceText(Beg, Len, std::move(Repl)));
+      if (UD->getNameAsString() == "max" || UD->getNameAsString() == "min") {
```
zhimingwang36: it should be better to remove using ::max instead of introduce dpct::max() function.

zhimingwang36 (changes requested):
LGTM

zhimingwang36 (changes requested):
LGTM

clang/lib/DPCT/ASTTraversal.cpp
zhiweij1: ok

zhimingwang36 (approved):
LGTM

PR #955, opened 2023-05-30
[DOC] Update options reference

krisdale:
Add new options: compilation-database, gen-helper-function.
Update values: use-experimental features, p

### DIFF
--- a/docs/_include_files/options_def.rst
+++ b/docs/_include_files/options_def.rst
@@ -646,6 +646,8 @@ By default, experimental features will not be used in migrated code.
 
 The values are:
 
+- ``=dpl-experimental-api``: Experimental extension that allows use of experimental
+  oneDPL APIs.
 - ``=free-function-queries``: Experimental extension that allows getting
   ``id``, ``item``, ``nd_item``, ``group``, and ``sub_group`` instances
   globally.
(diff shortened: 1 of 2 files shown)


docs/_include_files/options_def.rst
```diff
@@ -646,15 +646,22 @@ By default, experimental features will not be used in migrated code.
...
+- ``=dpl-experimental-api``: Experimental extension that allows use of experimental
+  oneDPL APIs.
 - ``=free-function-queries``: Experimental extension that allows getting
   ``id``, ``item``, ``nd_item``, ``group``, and ``sub_group`` instances
   globally.
 - ``=local-memory-kernel-scope-allocation``: Experimental extension that
   allows allocation of local memory objects at the kernel functor scope.
 - ``=logical-group``: Experimental helper function used to logically
   group work-items.
+- ``=masked_sub_group_operation``: Experimental helper function used to execute
```
zhimingwang36:
```suggestion
- ``=masked-sub_group-operation``: Experimental helper function used to execute
```
How about this? keep same style as other,  while "nd_range_barrier" is a helper API name. I mean we use "-" instead of "_" is there is no exactly API name or Class name in helper function.

zhiweij1: @zhimingwang36  `logical-group` is also a helper class name

krisdale: My vote is for consistency - if experimental functions will always use `_` and extensions will use `-`, we should do it consistently. Did you intend to suggest mixed use of `-` and `_` (`masked-sub_group-operation`)?

zhimingwang36: I was thinking sub_group is key words.  Yes, let's unify the style to use "-" instead of "_".   so  to use :  masked-sub-group-operation

krisdale: Updated

zhimingwang36 (approved):
LGTM

PR #992, opened 2023-06-07
SYCLomatic] Add functor workaround which const-ifys functors

danhoeflinger:
The intention for this PR is provide a workaround wrapper to allow user code to function when no `const` function call operator is provided.  
This provided function object should wrap incoming user function objects when calling oneDPL or dpct algorithms with an appropriate message stating that use within SYCL kernels of function objects which modify themselves during their functional call operators results in undefined behavior.  The user should be prompted via a comment to check that their function object is in fact const, mark it as such, then remove this workaround wrapper function.

Here is the proposed addition to oneDPL's documentation which adds this requirement: https://github.com/oneapi-src/oneDPL/pull/1083

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/functional.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/functional.h
@@ -26,6 +26,21 @@ namespace dpct {
 
 struct null_type {};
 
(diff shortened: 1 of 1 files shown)


tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #994, opened 2023-06-07
[DOC] Update intercept-build options format

krisdale:
Update formatting to be consistent with other
option source definition.


zhimingwang36 (approved):
LGTM

PR #1004, opened 2023-06-12
[SYCLomatic] Support migration of 33 cusparse API

zhiweij1:


clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
```diff
@@ -331,6 +331,524 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+  /// \param [in] base Indicates how input arrays are indexed.
+  /// \param [in] value_type Data type of the \p value .
+  static void create_csr(sparse_matrix_desc_t *desc, std::int64_t row_num,
+                         std::int64_t col_num, std::int64_t nnz, void *row_ptr,
+                         void *col_ind, void *value,
+                         library_data_t row_ptr_type,
+                         library_data_t col_ind_type,
+                         oneapi::mkl::index_base base,
+                         library_data_t value_type) {
+    *desc =
```
spencerpatty: does it make sense to check if desc is nullptr or not before allocating on the stack?  What happens if you call this more than once on same sparse_matrix_desc_t?  Is that something you need to handle ?

spencerpatty: would requite a different initialization routine besides the constructor,  but that could be reused in constructor as well ...

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
```diff
@@ -331,6 +331,524 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+        _uplo = oneapi::mkl::uplo::upper;
+      } else if (*uplo_ptr == oneapi::mkl::uplo::lower) {
+        _uplo = oneapi::mkl::uplo::lower;
+      } else {
+        throw std::runtime_error("unsupported uplo value");
+      }
+    } else {
+      throw std::runtime_error("unsupported attribute");
+    }
+  }
```
spencerpatty: oneMKL design is to have the uplo/diag/op attributes be separate from the matrix data itself so you can utilize the same matrix handle for different calls and have all usage optimizations/code paths internally coexist for the different calls...  Does it make sense to likewise move the attributes here to a separate companion class that can go with your descriptor?  Or what is the design you are planning -- the descriptor represents the data and attributes for a single operation ?

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
```diff
@@ -331,6 +331,524 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+    _col_ind = col_ind;
+    _value = value;
+  }
+
+private:
+  enum class routine_name { gemv, trmv, trsv };
+  std::optional<std::tuple<routine_name, oneapi::mkl::transpose,
+                           std::optional<oneapi::mkl::uplo>,
+                           std::optional<oneapi::mkl::diag>>>
+      optimizd_matrix_info;
```
spencerpatty:
```suggestion
      optimized_matrix_info;
```
it is only one letter, but makes readability so much better :)

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
```diff
@@ -331,6 +331,524 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+    auto beta_value =                                                          \
+        detail::get_value(reinterpret_cast<const Ty *>(beta), queue);          \
+    auto data_x = detail::get_memory(reinterpret_cast<Ty *>(x->_value));       \
+    auto data_y = detail::get_memory(reinterpret_cast<Ty *>(y->_value));       \
+    if (a->_diag.has_value() && a->_uplo.has_value()) {                        \
+      if (!a->is_optimized(sparse_matrix_desc::routine_name::trmv, trans,      \
+                           a->_uplo.value(), a->_diag.value())) {              \
+        oneapi::mkl::sparse::optimize_trmv(queue, a->_uplo.value(), trans,     \
+                                           a->_diag.value(),                   \
+                                           a->get_matrix_handle());            \
```
spencerpatty: I don't see event handling here.  Are you assuming this is an ordered queue?  or are you just missing the events for asynchronous behaviour?

clang/test/dpct/helper_files_ref/include/sparse_utils.hpp
```diff
@@ -262,6 +262,524 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
 }
 #endif
 
+/// Structure for describe a dense vector
```
spencerpatty: Can you explain the difference between this file and the sparse_utils.hpp.inc file?  it sort of looks like one has event handling and the other doesn't.  Is that it?  I put all my comments in the sparse_utils.hpp.inc file, but most apply to both files.

spencerpatty (commented):
Hi, here is my first review pass.  It looks nice to have dense matrix, dense vector and sparse matrix classes -- similar to cuSPARSE design.

clang/test/dpct/helper_files_ref/include/sparse_utils.hpp
zhiweij1: sparse_utils.hpp.inc the "source" code, before building dpct, there is a python script generating some special info about the helper functions and a final header file (clang/test/dpct/helper_files_ref/include/sparse_utils.hpp is the reference used to check the header final file).
The only difference of those two files is that *.hpp.inc file conatins special comment like "// DPCT_???" which will be parsed by a python script.
We plan to remove those  *.hpp.inc file recently, so you can treat those two files are same.

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
zhiweij1: Yes, the input queue should be an in-ordered queue.
We documented this requirement in the commnet. (Line 691)
```
/// \param [in] queue The queue where the routine should be executed. It must
/// have the in_order property when using the USM mode.
```

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
zhiweij1: I will fix it.

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
zhiweij1: I think I can remove the create/destroy method and use std::make_shared/reset to migrate cusparse???Create/Destroy.
Then the re-creation issue can be solved by the std::shared_ptr.

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
zhiweij1: The design is containing uplo and diag in the descriptor. `cusparseSpMatDescr_t` also contains those attributes. If we use an attribute struct and a descriptor separately, we need insert another variable. This will make a 1:1 mapping to 1:2 mapping which is what we want to avoid.

zhiweij1:
Hi @spencerpatty  , I updated the PR according to your comments.

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
spencerpatty: i get this design, but have questions about the reuse of such a descriptor.  One of the benefits of a matrix handle like this is that we can add multiple optimizations and those optimizations could potentially reuse internal structures benefiting ultimate timing.  I don't know if your descriptor will be used for multiple versions of uplo and diag, but the onemkl implementation supports this, and it would be common in an iterative solver preconditioner like symgs to use both upper trsv and lower trsv, for instance.  does this requrie two descriptors with their separate optimizations ?

zhiweij1: I think this descriptor can also support multiple versions of uplo and diag.
This descriptor just a wrapper of oneapi::mkl::sparse::matrix_handle, if user pass the dense_matrix_descriptor to optimize_trmv with different uplo/diag, oneapi::mkl::sparse::matrix_handle can save the different optimizations, so that the dense_matrix_descriptor also can do that.

zhiweij1: @spencerpatty is there any further comment?

spencerpatty: ok, so I completely agree that our mkl::sparse::matrix_handle will keep all the different optimizations you throw at it so they are all usable.  I wonder about your set_optimized_info and is_optimized() function design -- they seem to only be tracking the last optimized entry, unless I am mistaken and am missing some storing mechanism...

zhiweij1: > they seem to only be tracking the last optimized entry

@spencerpatty , yes, I added this check to avoid calling optimize_xyz() for the same configs. But now I only check the current config and the previous config.
Will mkl::optimize_xyz() do duplicate optimizations if I call it multiple times with same config (same data and same trans/uplo/diag) ?

spencerpatty: optimize_xyz will check if previous optimizations are already completed and return early (keeping the DAG consistent) if you call it multiple times with same config.  So yes, you are safe.

spencerpatty (approved):
changes look reasonable from oneMKL perspective.  I approve

clang/runtime/dpct-rt/include/sparse_utils.hpp.inc
zhiweij1: @spencerpatty , then I think I can remove the `is_optimized` check safely.

spencerpatty: yes this is also a good solution -- just depend on our internal handling of all the opts in the handle

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+                     library_data_t value_type, matrix_format data_format)
+      : _row_num(row_num), _col_num(col_num), _nnz(nnz), _row_ptr(row_ptr),
+        _col_ind(col_ind), _value(value), _row_ptr_type(row_ptr_type),
+        _col_ind_type(col_ind_type), _base(base), _value_type(value_type),
+        _data_format(data_format) {
+    if (_data_format != matrix_format::csr) {
+      throw std::runtime_error("the sparse matrix data format is unsupported");
+    }
+    oneapi::mkl::sparse::init_matrix_handle(&_matrix_handle);
+    constrcut();
```
zhiweij1:
```suggestion
    construct();
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+                library_data_t *value_type) {
+    *ele_num = _ele_num;
+    *value = _value;
+    *value_type = _value_type;
+  }
+  void get_desc(std::int64_t *ele_num, void **value,
+                library_data_t *value_type) {
+    get_desc(ele_num, const_cast<const void **>(value), value_type);
+  }
+  std::int64_t _ele_num;
```
zhiweij1, line 291: remove _ or move to private

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+                std::int64_t *leading_dim, void **value,
+                library_data_t *value_type, oneapi::mkl::layout *layout) {
+    *row_num = _row_num;
+    *col_num = _col_num;
+    *leading_dim = _leading_dim;
+    *value = _value;
+    *value_type = _value_type;
+    *layout = _layout;
+  }
+  std::int64_t _row_num;
```
zhiweij1, line 322: remove _ or move to private

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+    // Assume the new data is different from the old data
+    _row_ptr = row_ptr;
+    _col_ind = col_ind;
+    _value = value;
+    constrcut();
+  }
+
+private:
+  void constrcut() {
+#define SET_DATA(INDEX_TYPE, VALUE_TYPE)                                       \
```
zhiweij1: change to template

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+      a->get_matrix_handle(), data_x, beta_value, data_y);                     \
+  a->add_dependency(e);
+#define GEMV_STMT                                                              \
+  sycl::event e = oneapi::mkl::sparse::gemv(queue, trans, alpha_value,         \
+                                            a->get_matrix_handle(), data_x,    \
+                                            beta_value, data_y);               \
+  a->add_dependency(e);
+#endif
+
+#define SPMV(Ty)                                                               \
```
zhiweij1: change to template

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,484 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+  oneapi::mkl::sparse::gemv(queue, trans, alpha_value, a->get_matrix_handle(), \
+                            data_x, beta_value, data_y);
+#else
+#define TRMV_STMT                                                              \
+  sycl::event e = oneapi::mkl::sparse::trmv(                                   \
+      queue, a->_uplo.value(), trans, a->_diag.value(), alpha_value,           \
+      a->get_matrix_handle(), data_x, beta_value, data_y);                     \
+  a->add_dependency(e);
+#define GEMV_STMT                                                              \
+  sycl::event e = oneapi::mkl::sparse::gemv(queue, trans, alpha_value,         \
```
zhiweij1:
```suggestion
#define CALL(x) sycl::event e = x; a->add_dependency(e);
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

ziranzha (approved):

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -262,6 +262,469 @@ void optimize_csrsv(sycl::queue &queue, oneapi::mkl::transpose trans,
...
+    get_desc(ele_num_data, const_cast<const void **>(value_data),
+             value_type_data);
+  }
+  std::int64_t ele_num;
+  void *value;
+  library_data_t value_type;
+};
+
+/// Structure for describe a dense matrix
+struct dense_matrix_desc {
```
zhimingwang36: Let's unify the style of class/struct for dense_matrix_desc and sparse_matrix_desc

zhiweij1: done

zhimingwang36 (approved):
LGTM

PR #1020, opened 2023-06-16
[SYCLomatic] Support migration of __half/__half2 constuctor in kernel arg.

tangjj11:

### DIFF
--- a/clang/lib/DPCT/ExprAnalysis.cpp
+++ b/clang/lib/DPCT/ExprAnalysis.cpp
@@ -576,12 +576,7 @@ void ExprAnalysis::analyzeExpr(const CXXUnresolvedConstructExpr *Ctor) {
 }
 
 void ExprAnalysis::analyzeExpr(const CXXTemporaryObjectExpr *Temp) {
-  std::string TypeName = DpctGlobalInfo::getUnqualifiedTypeName(
-      Temp->getType().getCanonicalType());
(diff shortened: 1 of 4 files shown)


clang/lib/DPCT/ExprAnalysis.cpp
```diff
@@ -584,6 +584,10 @@ void ExprAnalysis::analyzeExpr(const CXXTemporaryObjectExpr *Temp) {
        StringRef(TypeName).startswith("cooperative_groups::")) {
     analyzeType(Temp->getTypeSourceInfo()->getTypeLoc());
   }
+  if (TypeName == "__half" || TypeName == "__half2") {
```
wenhuiNi: Adding special code in ExprAnalysis.cpp is a not good way.
In ExprAnalysis.cpp, we hope that the code can be used to solve a type of problem.
Can we analyze why it cannot be constuct/convert in the kernel parameters?

tangjj11: If I delete these 2 `if` and call `analyzeType` anyway. the `ctad` case will failed because in `ctad` mode, we should not infer template argument.

tangjj11: fixed

wenhuiNi (approved):

ziranzha (approved):

PR #1051, opened 2023-06-29
[SYCLomatic] Migrate sincos* to sin and cos seperately

intwanghao:


clang/lib/DPCT/CallExprRewriterMath.cpp
```diff
@@ -636,41 +636,30 @@ std::optional<std::string> MathSimulatedRewriter::rewrite() {
           MigratedArg0 = "(double)(" + MigratedArg0 + ")";
       }
     }
+    auto &SM = DpctGlobalInfo::getSourceManager();
```
zhiweij1, line 639: Line 616~674 can be implemented with the new math rewriter.

clang/lib/DPCT/CallExprRewriterMath.cpp
```diff
@@ -636,41 +636,30 @@ std::optional<std::string> MathSimulatedRewriter::rewrite() {
...
+    std::string Indent =
+        getIndent(SM.getExpansionLoc(Call->getBeginLoc()), SM).str();
```
zhiweij1, line 641: I think keeping the original migration is better. But it need to be re-implemented by the new math rewriter.


And the new migration can only be enabled if --optimized-migration is specified. We can use MULTI_STMTS_FACTORY_ENTRY in the rewriter.

PR #1083, opened 2023-07-11
[SYCLomatic] Add run length encode nontrivial runs API

mmichel11:
This PR implements the run length encoding nontrivial runs API. In order for the added source code in this PR to compile, https://github.com/oneapi-src/oneDPL/pull/1019 must be first merged into oneDPL. The corresponding test PR for this implementation is provided in https://github.com/oneapi-src/SYCLomatic-test/pull/393

Nontrivial runs are a subsequence of length two or greater in a buffer that are equivalent. This API marks the start of each nontrivial run, length of each nontrivial run, and the number of nontrivial runs.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -1761,6 +1761,7 @@ template <typename ExecutionPolicy, typename InputIterator,
       typename ::std::iterator_traits<OutputIterator2>::value_type;
 
   auto n = ::std::distance(input_beg, input_end);
+  // First element must be nontrivial run (start of first segment)
   auto first_adj_it = oneapi::dpl::adjacent_find(policy, input_beg, input_end);
   auto first_adj_idx = ::std::distance(input_beg, first_adj_it);
   if (first_adj_it == input_end) {
@@ -1775,40 +1776,50 @@ template <typename ExecutionPolicy, typename InputIterator,
     auto out_idx = idx + first_adj_idx;
     return (out_idx == n - 1) ? n - 1 : out_idx + 1;
   };
+  // TODO: Use shifted view to pad range once oneDPL ranges is non-experimental
   auto left_shifted_input_beg =
       oneapi::dpl::make_permutation_iterator(input_beg, get_prev_idx_element);
   auto right_shifted_input_beg =
       oneapi::dpl::make_permutation_iterator(input_beg, get_next_idx_element);
+  // Segment type for ith idx consists of zip of iterators at (i-1, i, i+1)
+  // padded at the ends
   auto zipped_keys_beg = make_zip_iterator(
       left_shifted_input_beg, input_beg, right_shifted_input_beg,
       oneapi::dpl::counting_iterator<offsets_t>(0));
-  auto flags_beg =
+  // Set flag at the beginning of new nontrivial run (ex: (0, 1, 1) -> 1)
+  auto key_flags_beg =
       make_transform_iterator(zipped_keys_beg, [n](const auto &zipped) {
         using ::std::get;
-        if (get<3>(zipped) == n - 1)
-          return false;
-        if (get<3>(zipped) == 0)
-          return get<1>(zipped) == get<2>(zipped);
-        return get<0>(zipped) != get<1>(zipped) &&
-               get<1>(zipped) == get<2>(zipped);
+        bool zero_idx_mask = get<3>(zipped) != 0;
+        bool last_idx_mask = get<3>(zipped) != n - 1;
+        return (zero_idx_mask * (get<0>(zipped) != get<1>(zipped)) &&
+                get<1>(zipped) == get<2>(zipped)) *
+               last_idx_mask;
       });
   auto count_beg = oneapi::dpl::counting_iterator<offsets_t>(0);
   auto const_it = dpct::make_constant_iterator(lengths_t(1));
-  auto zipped_vals_beg =
+  // Check for presence of nontrivial element at current index
+  auto tr_nontrivial_flags = make_transform_iterator(
       make_zip_iterator(left_shifted_input_beg, input_beg,
-                        right_shifted_input_beg, count_beg, const_it);
+                        right_shifted_input_beg),
+      [](const auto &zip) {
+        using ::std::get;
+        return get<0>(zip) == get<1>(zip);
+      });
+  auto zipped_vals_beg =
+      make_zip_iterator(tr_nontrivial_flags, count_beg, const_it);
   auto pred = [](bool lhs, bool rhs) { return !rhs; };
+  // Update count of run. Starting index of run stored in assignment of initial
+  // element (first element in each segment).
   auto op = [](auto lhs, const auto &rhs) {
     using ::std::get;
-    if (get<0>(rhs) == get<1>(rhs))
-      ++::std::get<4>(lhs);
-    return lhs;
+    get<2>(lhs) += get<0>(rhs);
+    return ::std::move(lhs);
   };
-  auto zipped_out_beg = make_zip_iterator(
-      oneapi::dpl::discard_iterator(), oneapi::dpl::discard_iterator(),
-      oneapi::dpl::discard_iterator(), offsets_out, lengths_out);
+  auto zipped_out_beg = make_zip_iterator(oneapi::dpl::discard_iterator(),
+                                          offsets_out, lengths_out);
   auto [_, zipped_out_vals_end] = oneapi::dpl::reduce_by_segment(
-      policy, flags_beg + first_adj_idx, flags_beg + n,
+      policy, key_flags_beg + first_adj_idx, key_flags_beg + n,
       zipped_vals_beg + first_adj_idx, oneapi::dpl::discard_iterator(),
       zipped_out_beg, pred, op);
   auto ret_dist = ::std::distance(zipped_out_beg, zipped_out_vals_end);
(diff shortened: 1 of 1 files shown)


clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,87 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+  auto left_shifted_input_beg =
+      oneapi::dpl::make_permutation_iterator(input_beg, get_prev_idx_element);
+  auto right_shifted_input_beg =
+      oneapi::dpl::make_permutation_iterator(input_beg, get_next_idx_element);
+  // Segment type for ith idx consists of zip of iterators at (i-1, i, i+1)
+  // padded at the ends
+  auto zipped_keys_beg = make_zip_iterator(
+      left_shifted_input_beg, input_beg, right_shifted_input_beg,
+      oneapi::dpl::counting_iterator<offsets_t>(0));
+  // Set flag at the beginning of new nontrivial run (ex: (0, 1, 1) -> 1)
```
danhoeflinger: It might be more clear to use an example like `(ex: (2, 3, 3) -> 1)` to better indicate that the values are element keys rather than flags of some sort (0,1).

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,87 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+        return get<0>(zip) == get<1>(zip);
+      });
+  auto zipped_vals_beg =
+      make_zip_iterator(tr_nontrivial_flags, count_beg, const_it);
+  auto pred = [](bool lhs, bool rhs) { return !rhs; };
+  // Update count of run. Starting index of run stored in assignment of initial
+  // element (first element in each segment).
+  auto op = [](auto lhs, const auto &rhs) {
+    using ::std::get;
+    get<2>(lhs) += get<0>(rhs);
```
danhoeflinger, line 1821: I think a comment here about how the `lhs` count index (`get<1>`) is preserved and passed through to the output here would be helpful.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,87 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+  auto count_beg = oneapi::dpl::counting_iterator<offsets_t>(0);
+  auto const_it = dpct::make_constant_iterator(lengths_t(1));
+  // Check for presence of nontrivial element at current index
+  auto tr_nontrivial_flags = make_transform_iterator(
+      make_zip_iterator(left_shifted_input_beg, input_beg,
+                        right_shifted_input_beg),
+      [](const auto &zip) {
+        using ::std::get;
+        return get<0>(zip) == get<1>(zip);
+      });
```
danhoeflinger: Do we need the `(i+1)` input here in the zip?  The zip only needs 2 elements, since that is all the transform uses.  

The fact that we seed the count at 1 is good, and allows us to make use the of knowledge a non-trivial run exists in this segment and avoid a second comparison for each element.

Generally, I think it ends up being equivalent to compare `*i == *(i+1)` vs comparing `*(i-1) == *i`.  However, depending on which we choose, I think we need special casing for the first or the last element, due to the padding strategy of repeating the element.  Without it I believe we can end up with an inflated result (+1) for our first or last non-trivial run.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,87 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+  // Segment type for ith idx consists of zip of iterators at (i-1, i, i+1)
+  // padded at the ends
+  auto zipped_keys_beg = make_zip_iterator(
+      left_shifted_input_beg, input_beg, right_shifted_input_beg,
+      oneapi::dpl::counting_iterator<offsets_t>(0));
+  // Set flag at the beginning of new nontrivial run (ex: (0, 1, 1) -> 1)
+  auto key_flags_beg =
+      make_transform_iterator(zipped_keys_beg, [n](const auto &zipped) {
+        using ::std::get;
+        bool zero_idx_mask = get<3>(zipped) != 0;
```
danhoeflinger: I think this `zero_idx_mask` is unnecessary.  The flag predicate only cares about the `rhs`, not the `lhs`.  Also, `reduce_by_segment` always assumes that a segment starts at the zeroth element.  To pair with that we always ensure in this algo that the first element starts a segment of a non-trivial run due to the `adjacent_find` call. 

In the end, we ignore the segment flag corresponding to the zeroth index of the key, so there is no need to special case it actually.  (If we did want it, I'd think we would want `get<3>(zipped) == 0` with an `||`, not an `*` or `&&`).
 
 We **do** want the last element special case though.  If we could pad on the end with a value we know returns a false to an `==` comparison with the last element, that would be better.  However, I don't know of a general way to do that.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,87 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+        bool last_idx_mask = get<3>(zipped) != n - 1;
+        return (zero_idx_mask * (get<0>(zipped) != get<1>(zipped)) &&
+                get<1>(zipped) == get<2>(zipped)) *
+               last_idx_mask;
```
danhoeflinger: Is `*` better than `&&` here for some reason?

danhoeflinger (changes requested):
I think this is improved, but I think there may be a bug and some unnecessary code.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
mmichel11: Good point. I have made this change.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
mmichel11: I have added a comment to explain all get<1> holds the start of each run and is preserved throughout each segment's reduction.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
mmichel11: We do not need the (i+1) input here. I have removed `right_shifted_input_beg`, so we end up just comparing `*(i-1) == *i`.

Also, there is no correctness issue in this case. The bogus `1` flag set when we compare `i-1` and `i` when `i=0` never affects the run count as it never appears as a `rhs` argument in `op`. I have added a comment to explain this. It makes the code a bit more complicated but allows us to have less comparisons in our operator and start the count at 1. This case is also added in the test PR which passes.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
mmichel11: Yes, you are correct. I have removed `zero_idx_mask`.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
mmichel11: No, I have switched to only use `&&` which is clearer.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1746,6 +1746,91 @@ segmented_reduce_argmax(Policy &&policy, Iter1 keys_in, Iter2 keys_out,
...
+  // will never be added to the length count which is only updated in
+  // get<2>(lhs) += get<0>(rhs) and initialized to 1.
+  auto op = [](auto lhs, const auto &rhs) {
+    using ::std::get;
+    // Update count of run. The run's starting index is stored in get<1>(lhs) as
+    // the initial value in the segment and is preserved throughout the
+    // segment's reduction.
+    get<2>(lhs) += get<0>(rhs);
+    return ::std::move(lhs);
+  };
```
danhoeflinger:
```suggestion
  auto op = [](auto lhs, const auto &rhs) {
    using ::std::get;
    
    // Update length count of run.
    // The first call to this op will use the first element of the input as lhs and second element
    // as rhs.  get<0>(first_element) is always ignored in favor of a constant `1` in get<2>, 
    // avoiding the need for special casing the first element. The constant `1` utilizes the 
    // knowledge that each segment begins with a non-trivial run.
    get<2>(lhs) += get<0>(rhs);
    
    // The run's starting index is stored in get<1>(lhs) as
    // the initial value in the segment and is preserved throughout the
    // segment's reduction to end up as the non-trivial run's offset.
    
    return ::std::move(lhs);
  };
```

I took a stab at revising these comments to makes things a little more clear.  Take/adapt them if you wish.

mmichel11: Thanks, I like these revisions. I have added them with some minor wording changes.

danhoeflinger (approved):
Thanks, LGTM.

yihwang-nv (approved):
LGTM

PR #1106, opened 2023-07-20
[SYCLomatic]Fix max_work_item_sizes return type from id to range

wenhuiNi:


wenhuiNi:
need waiting uplift compiler

yihwang-nv (approved):

clang/runtime/dpct-rt/include/device.hpp
```diff
@@ -138,7 +138,7 @@ class device_info {
       _name[255] = '\0';
     }
   }
-  void set_max_work_item_sizes(const sycl::id<3> max_work_item_sizes) {
+  void set_max_work_item_sizes(const sycl::range<3> max_work_item_sizes) {
```
zhimingwang36, line 151: 1. please add a new interface for backward compatibility. 
2. try to unify _max_work_item_sizes and _max_work_item_sizes_i.

AndyCHHuang (approved):
LGTM

ziranzha (approved):

clang/runtime/dpct-rt/include/device.hpp
wenhuiNi: for comments 1, fixed.
for comments 2 ,  in cuda side,  the type of maxThreadsDim is int. In some case, we need keep the same type

clang/runtime/dpct-rt/include/device.hpp
```diff
@@ -64,22 +64,22 @@ class device_info {
   // get interface
   const char *get_name() const { return _name; }
   char *get_name() { return _name; }
-  template <typename WorkItemSizesTy = sycl::id<3>,
-            std::enable_if_t<std::is_same_v<WorkItemSizesTy, sycl::id<3>> ||
+  template <typename WorkItemSizesTy = sycl::range<3>,
```
zhimingwang36, line 67: Please add back old interface for sycl::id,  at same time add "deprecated"  keywords.

clang/runtime/dpct-rt/include/device.hpp
```diff
@@ -138,11 +138,17 @@ class device_info {
       _name[255] = '\0';
     }
   }
-  void set_max_work_item_sizes(const sycl::id<3> max_work_item_sizes) {
+  void set_max_work_item_sizes(const sycl::range<3> max_work_item_sizes) {
     _max_work_item_sizes = max_work_item_sizes;
     for (int i = 0; i < 3; ++i)
       _max_work_item_sizes_i[i] = max_work_item_sizes[i];
   }
+  void set_max_work_item_sizes(const sycl::id<3> max_work_item_sizes) {
```
zhimingwang36: Please add "deprecated" keywords

clang/runtime/dpct-rt/include/device.hpp
wenhuiNi: not needed, sycl::id can be implicitly converted by sycl::range

clang/runtime/dpct-rt/include/device.hpp
zhimingwang36: To avoid sync between _max_work_item_sizes and _max_work_item_sizes_i,  just keep one is enough. 
construct another if need in use point.

zhimingwang36 (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1109, opened 2023-07-24
[SYCLomatic] Enable support of system level atomic functions' migration

tomflinda:


clang/test/dpct/atomic_functions_system_wide.cu
```diff
@@ -0,0 +1,34 @@
...
+// RUN: FileCheck --input-file %T/atomic_functions_system_wide/atomic_functions_system_wide.dp.cpp --match-full-lines %s
+
+#include <cuda_runtime.h>
+
+#include <iostream>
+#include <memory>
+
+// CHECK:void atomic_kernel(int *atomic_array, const sycl::nd_item<3> &item_ct1) {
+// CHECK-NEXT:  unsigned int tid = item_ct1.get_local_range(2) * item_ct1.get_group(2) + item_ct1.get_local_id(2);
+// CHECK-NEXT:  dpct::atomic_fetch_add<sycl::access::address_space::generic_space, sycl::memory_order::relaxed, sycl::memory_scope::system>(&atomic_array[0], 10);
```
Pennycook, line 11: Why are we migrating things to a DPCT-specific function here, when the SYCL 2020 version is standard and less verbose?

The current migration is 147 characters:
```c++
dpct::atomic_fetch_add<sycl::access::address_space::generic_space, sycl::memory_order::relaxed, sycl::memory_scope::system>(&atomic_array[0], 10);
```

Using standard SYCL is only 103 characters:
```c++
sycl::atomic_ref<int, sycl::memory_order::relaxed, sycl::memory_scope::system>(atomic_array[0]) += 10;
```

tomflinda: 1. the reason we have 'sycl::access::address_space::generic_space' as a template argument is because it caused runtime error in cuda backend. therefore, the default value of the template parameter was set to 'sycl::access::address_space::global_space'. Since the issue has been fixed in cuda backend, we update the default value to 'sycl::access::address_space::generic_space'. in that case. the code length of  using helper function is the same as the one of using native sycl api.
```
dpct::atomic_fetch_add<sycl::memory_order::relaxed, sycl::memory_scope::system>(&atomic_array[0], 10);
sycl::atomic_ref<int, sycl::memory_order::relaxed, sycl::memory_scope::system>(atomic_array[0]) += 10;
```
2. the reason we prefer using helper function is because the helper function solves the implicit casting issue between the two operands:
```
inline T1 atomic_fetch_add(T1 *addr, T2 operand)
```

Pennycook: Does `+=` not do the implicit casting?  `atomic_ref<T1, ...>` knows what data type the pointer points to, and I would have thought that `+=` would accept `T2` on the RHS if it was convertible to `T1`.

tomflinda: No

Pennycook: Should it?  I'm trying to work out if this is a specification issue or an implementation issue.

tomflinda: Take the example below:
```
#include<iostream>
void foo(int a, int b) { std::cout<<"foo int"<<std::endl; }
void foo(float a, float b) { std::cout<<"foo float"<<std::endl; }

template<class T>
void foo2(T a, T b) {std::cout<<"foo2"<<std::endl;}

int main(){
  float a;
  double b;
  foo(a, b); // std output "foo float"
  foo2(a, b); // compile error
}
```
cuda runtime API uses overloading instead of template to implement the atomic functions. (similar to the foo() in the example), while std and sycl use template implementation like foo2() in the example.
Therefore, when the types of the arguments in the function call are different, sycl and std version of atomic functions will emit compile fail.
In my opinion, there is no issue in SYCL spec and implementation, therefore, it makes sense to use helper function to bridge the gap of the two libraries.

Pennycook: I don't think your example matches how `atomic_ref` is defined.

The templates on `atomic_ref` are on the class. The `fetch_add` and `+=` member functions are just functions defined to take a `T`, which means they should accept any type convertible to `T`.  For example, https://godbolt.org/z/TrbvfE4vf shows that the following code compiles:
```c++
#include <atomic>

void atomic_ref_example(float* ptr) {
    std::atomic_ref<float>(*ptr) += (int) 1; // calls atomic_ref<float>::operator+=(float)
}
```

To match your example:
```c++
#include<iostream>

template <class T>
struct S {
    void foo(T a, T b) { std::cout << "foo" << std::endl; }
};

template <>
struct S<float> {
    void foo(float a, float b) { std::cout << "foo float" << std::endl; }
};

template <>
struct S<double> {
    void foo(double a, double b) { std::cout << "foo double" << std::endl; }
};

int main(){
  float a;
  double b;
  S<float>{}.foo(a, b);  // foo float
  S<double>{}.foo(a, b); // foo double
}
```

tomflinda: Take the demo.cu  for example:
```
#define CALL_ATOMIC_ADD(a, b) atomicAdd(a, b) // cannot unify the template argument of the migrated code if using the native SYCL AP

__global__ void atomic_kernel_int(int *atomic_array) {
  unsigned int tid = blockDim.x * blockIdx.x + threadIdx.x;
  CALL_ATOMIC_ADD(&atomic_array[0], 10);
}

__global__ void atomic_kernel_float(float *atomic_array) {
  unsigned int tid = blockDim.x * blockIdx.x + threadIdx.x;
  CALL_ATOMIC_ADD(&atomic_array[0], 10.0);
}
```
Using helper function, current migration code is as follow:
```
#define CALL_ATOMIC_ADD(a, b)                                                  \
  dpct::atomic_fetch_add<sycl::access::address_space::generic_space>(          \
      a, b) // cannot unify the template argument of the migrated code if using the native SYCL AP

void atomic_kernel_int(int *atomic_array, const sycl::nd_item<3> &item_ct1) {
  unsigned int tid = item_ct1.get_local_range(2) * item_ct1.get_group(2) +
                     item_ct1.get_local_id(2);
  CALL_ATOMIC_ADD(&atomic_array[0], 10);
}

void atomic_kernel_float(float *atomic_array, const sycl::nd_item<3> &item_ct1) {
  unsigned int tid = item_ct1.get_local_range(2) * item_ct1.get_group(2) +
                     item_ct1.get_local_id(2);
  CALL_ATOMIC_ADD(&atomic_array[0], 10.0);
}
```

Overall, statically analyzing the type of an expression and using the type in the migrated code as template argument or static cast will create code unify issue in Macros. Therefore, we prefer to using the helper functions.

Moreover, cuda atomicAdd() returns the original value, while sycl atomic_ref += return the new value 
image in https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#architecture chapter 4.15.3.

Pennycook: I'm disappointed that we can't generate code that looks more like standard C++ and standard SYCL, but I concede that macros make this difficult.  Thanks for the explanation.

FWIW, I don't think the `+=` return value is a show-stopper.  Since `atomic_ref` also has a `fetch_add` function, we could just generate that instead.

tomflinda: Thanks for your feedback.

AndyCHHuang (approved):
LGTM

zhiweij1 (approved):

ziranzha (approved):

PR #1126, opened 2023-07-28
[SYCLomatic] Add missing scatter_if and gather_if APIs

mmichel11:
This PR adds the missing `scatter_if` and `gather_if` APIs where no user defined unary predicate is provided. In this scenario, the mask value is assumed to be convertible to `bool` and directly used as the predicate value.

These APIs are added to the `dpct` namespace in the `algorithm.h` header.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -375,7 +375,7 @@ template <typename Policy, typename InputIter1, typename InputIter2,
 void scatter_if(Policy &&policy, InputIter1 first, InputIter1 last,
                 InputIter2 map, InputIter3 mask, OutputIter result) {
   scatter_if(::std::forward<Policy>(policy), first, last, map, mask, result,
-             [=](auto &&mask_element) { return mask_element; });
+             internal::no_op_fun());
 }
 
 template <typename Policy, typename InputIter1, typename InputIter2,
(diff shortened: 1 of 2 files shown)


danhoeflinger (commented):
I think you can skip a copy by adding and using something similar to `oneapi::dpl::__internal::__no_op()`.  Unless I'm mistaken, we are adding an unnecessary copy in the return operation from the supplied lambda.  

https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/utils.h#L87C1-L95C3

mmichel11:
> I think you can skip a copy by adding and using something similar to `oneapi::dpl::__internal::__no_op()`. Unless I'm mistaken, we are adding an unnecessary copy in the return operation from the supplied lambda.
> 
> https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/utils.h#L87C1-L95C3

You are correct. I will make this change by using the `oneDPL` utility.

danhoeflinger:
> > I think you can skip a copy by adding and using something similar to `oneapi::dpl::__internal::__no_op()`. Unless I'm mistaken, we are adding an unnecessary copy in the return operation from the supplied lambda.
> > https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/utils.h#L87C1-L95C3
> 
> You are correct. I will make this change by using the `oneDPL` utility.

I think it will be best to basically copy this oneDPL utility into the dpct detail namespace, so that we are not depending on non-public APIs of oneDPL here.  I'd prefer the redundancy of copying it here vs requiring tracking changes to the internals of oneDPL here.

danhoeflinger (approved):
LGTM

tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1128, opened 2023-07-31
[SYCLomatic] Migrate 10 cuSparse APIs

zhiweij1:
Signed-off-by: Jiang, Zhiwei zhiwei.jiang@intel.com


clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,11 +513,33 @@ class sparse_matrix_desc {
...
+      row_ptr = _temp_row_ptr = dpct::dpct_malloc(
+          sizeof(index_t) * (_row_num + 1), get_default_queue());
+    } else if (!_row_ptr && _temp_row_ptr) {
+      row_ptr = _temp_row_ptr;
+    } else if (_row_ptr && !_temp_row_ptr) {
+      row_ptr = _row_ptr;
+    } else if (_row_ptr && _temp_row_ptr) {
+      dpct::dpct_memcpy(
+          _row_ptr, _temp_row_ptr, sizeof(index_t) * (_row_num + 1),
+          memcpy_direction::device_to_device, get_default_queue());
```
gajanan-choudhary: Careful here. This may not always be device to device copy, unless the DPCT tool explicitly always allocates both `_row_ptr` and `_temp_row_ptr` with `sycl::malloc_device()` or `sycl::aligned_alloc_device()`. If that is true here, then this if fine. If not, please look into it.

gajanan-choudhary: At the same time, how (if at all) are you ensuring that `len(_row_ptr) >= len(_temp_row_ptr)` so that the copy doesn't go out-of-bounds and doesn't segfault?

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,11 +513,33 @@ class sparse_matrix_desc {
...
+      row_ptr = _row_ptr;
+      dpct::dpct_free(_temp_row_ptr, get_default_queue());
+      _temp_row_ptr = nullptr;
+    }
```
gajanan-choudhary: I see what you are trying to do here, but this is not allowed for oneMKL's sparse BLAS `matrix_handle_t`. You are not allowed to switch out arrays that you have already passed to `set_csr_data()` with new arrays. This is possibly why you had problems with `cusparseSpGEMM` to `matmat` portability support.
Although it _might_ work, it is not guaranteed to work the way you have set this up.

clang/lib/DPCT/APINamesCUSPARSE.inc
```diff
@@ -153,8 +153,8 @@ ASSIGNABLE_FACTORY(ASSIGN_FACTORY_ENTRY(
     "cusparseCreateCsr", DEREF(0),
     CALL("std::make_shared<" + MapNames::getDpctNamespace() +
              "sparse::sparse_matrix_desc>",
-         ARG(1), ARG(2), ARG(3), ARG(4), ARG(5), ARG(6), ARG(7), ARG(8), ARG(9),
-         ARG(10),
+         ARG(1), ARG(2), ARG(3), NULLPTR_OR_ARG(4), NULLPTR_OR_ARG(5),
+         NULLPTR_OR_ARG(6), ARG(7), ARG(8), ARG(9), ARG(10),
```
gajanan-choudhary, line 157: Assuming `ARG(4)` is for row-pointer array, shouldn't it be the following?
```suggestion
         ARG(1), ARG(2), ARG(3), ARG(4), NULLPTR_OR_ARG(5),
         NULLPTR_OR_ARG(6), ARG(7), ARG(8), ARG(9), ARG(10),
```
The size of the matrices are always known _a priori_, which means we know the length of the row-pointer array (`= nRows + 1`).

oneMKL sparse BLAS SYCL APIs do not support modifying arrays inside a matrix handle once they have been set, see "User agreements (2)" in the [documentation](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/sparse-blas-matrix-handle-contract-between-user.html):
> The User agrees not to modify the data arrays directly while they are in a sparse::matrix_handle_t.

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +767,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+  if (!a->get_uplo() || !a->get_diag()) {
+    throw std::runtime_error("oneapi::mkl::sparse::trsv needs uplo and diag "
+                             "attributes to be specified.");
+  }
+  oneapi::mkl::uplo uplo = a->get_uplo().value();
+  oneapi::mkl::diag diag = a->get_diag().value();
+
+  std::uint64_t key = dpct::detail::get_type_combination_id(
+      a->get_value_type(), x->get_value_type(), y->get_value_type());
+  switch (key) {
```
gajanan-choudhary, line 1162: oneMKL Sparse BLAS doesn't support mixed precision. You need to throw an exception if data type of `alpha` does not match these data types (`a`, `x`, `y`).

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +767,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+/// \param [in] data_type Specifies the data type of \param a, \param x and
+/// \param y .
+inline void spsv(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                 const void *alpha, sparse_matrix_desc_t a,
+                 std::shared_ptr<dense_vector_desc> x,
+                 std::shared_ptr<dense_vector_desc> y,
+                 library_data_t data_type) {
+  switch (data_type) {
+  case library_data_t::real_float:
+    detail::check_alpha_value<float>(alpha, queue);
```
gajanan-choudhary: Does this check include checking if `alpha` is on host or on device? oneMKL Sparse BLAS does not support `alpha` being on device. `alpha` is passed by value on the host to `sparse::trsv()`. Are you taking care of this?

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +767,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+    *bufSize = static_cast<std::int64_t>(*size_temp_buffer);
+    oneapi::mkl::sparse::matmat(queue, a->get_matrix_handle(),
+                                b->get_matrix_handle(), c->get_matrix_handle(),
+                                oneapi::mkl::sparse::matmat_request::compute,
+                                matmat_descr, bufSize, temp_buffer, {});
+    std::int64_t *c_nnz = sycl::malloc_host<std::int64_t>(1, queue);
+    oneapi::mkl::sparse::matmat(queue, a->get_matrix_handle(),
+                                b->get_matrix_handle(), c->get_matrix_handle(),
+                                oneapi::mkl::sparse::matmat_request::get_nnz,
+                                matmat_descr, c_nnz, nullptr, {});
```
gajanan-choudhary: Your USM APIs are problematic. You are not keeping track of and carrying over dependencies. This won't work. Please fix this (extract the event, `ev_compute = matmat(...)`, then `ev_nnz = matmat(..., {ev_compute})`, etc.)

gajanan-choudhary (commented):
We need quite a few changes for this to work. One important change I'd suggest is to have no `_temp_row_ptr` business in your matrix descriptor. The matrix should always be created with a user-provided row-pointer array instead of allowing a `nullptr` and internal allocation that you are later switching out.

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: > If that is true here, then this if fine. If not, please look into it.

This class is used to migrate `cusparseSpMatDescr_t`. In cuda side, the `csrRowOffsets` is device memory, so the migrated `_row_ptr` is also device memory. `_temp_row_ptr` is allocated by `dpct::dpct_malloc`, so it is also device memory.

> At the same time, how (if at all) are you ensuring that len(_row_ptr) >= len(_temp_row_ptr) so that the copy doesn't go out-of-bounds and doesn't segfault?

`len(_row_ptr)` sbould be same as `len(_temp_row_ptr)` because `_row_num` comes from the constructor argument which is provided by user. If it is not correct, the cuda code may also fail.

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: The queue used here should be an in-ordered queue. So I think the explicit dependency setting is not necessary.

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: Yes. We checked it. If it is device memory, we will copy it to host and then do comparing.

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: Yes. I noticed this. But I think it is unnecessary for `sparse::trsv` because `sparse::trsv` does not need to use `alpha`, it only support `alpha == 1`. So I think we can ignore alpha's type here.
For other APIs which accept alpha/beta, I will throw the exception if data type mismatch.

clang/lib/DPCT/APINamesCUSPARSE.inc
zhiweij1: Yes, I agree this size of the matrix is known here.
But cusparse support to pass nullptr when creating a `cusparseSpMatDescr_t` and then use some set API to set those pointers.
So the cuda code may like below:
```
cusparseSpMatDescr_t* spMatDescr;
cusparseCreateCsr(&spMatDescr, rows, cols, nnz, nullptr, nullptr, nullptr, ...);
cudaMalloc(...);
cusparseCsrSetPointers(spMatDescr, ...);
```
SYCLomatic cannot move the malloc before constructing descriptor (we cannot change the original code logic).
Do you have any suggestion to support the above case?

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: See below reply.

clang/lib/DPCT/APINamesCUSPARSE.inc
zhiweij1: And I checked the example `sparse_matmat_usm.cpp` in MKL package in 2023.2
In this file, matrix handle is set data twice:
```
249        auto ev_setC = oneapi::mkl::sparse::set_csr_data(main_queue, C, c_nrows, c_ncols, c_index, c_rowptr,
250                                                        (intType *)nullptr, (fpType *)nullptr, {});
...
333        ev_setC = oneapi::mkl::sparse::set_csr_data(main_queue, C, c_nrows, c_ncols, c_index, c_rowptr, c_colind, c_values, {ev3_1});
```
Does it violate the "User agreements (2)"?

gajanan-choudhary: To be clearer, _once the pointers have been set_ in oneMKL's `matrix_handle_t` type and they are no longer `nullptr`, they should not be switched out. It's okay to create the handle with `nullptr`s first, but once you set them to some pointer containing data, you should not switch it out with some other array -- I believe that is not allowed. Once you set the `_temp_row_ptr` in `matrix_handle_t`, switching it out with `_row_ptr` even if you are copying it could lead to unpredictable behavior.

gajanan-choudhary: So here, I think it might be okay to keep `NULLPTR_OR_ARG(4)`.

clang/runtime/dpct-rt/include/sparse_utils.hpp
gajanan-choudhary: Isn't the queue provided to you by the user? Is the queue under the complete control of DPCT? Doesn't the user have the freedom to change the queue to be out-of-order on their end once DPCT tool ports over CUDA code?

clang/lib/DPCT/APINamesCUSPARSE.inc
zhiweij1: OK. But there is still one issue left.
The cusparse allows to create `cusparseSpMatDescr_t` with nullptr of `row_ptr`, while the `matrix_handle_t` does not allow it.
```
  1 cusparseSpMatDescr_t* matC;
  2 cusparseCreateCsr(&matC, rows, cols, nnz, NULL, NULL, NULL, ...); // MKL need row_ptr ready here
  3 cusparseSpGEMM_workEstimation(...);
  4 cusparseSpGEMM_workEstimation(...);
  5 cusparseSpGEMM_compute(...);
  6 cusparseSpGEMM_compute(...);
  7 cusparseSpMatGetSize(...);
  8 cudaMalloc(&cols, ...);
  9 cudaMalloc(&values, ...);
 10 cudaMalloc(&row_ptr, ...);
 11 cusparseCsrSetPointers(matC, ..., cols, values, row_ptr); // cusparse can set row_ptr here
 12 cusparseSpGEMM_copy(...);
```
Since `matrix_handle_t` does not allow switch the pointer `row_ptr` at line 11, I think at least DPCT can emit a warning message if the `row_ptr` in the constructor is NULL (can be evaluated in compile-time) and throw an exception in the constructor during runtime if it is nullptr. Then I can open a feature request to oneMKL for supporting `matrix_handle_t` to accept NULL `row_ptr` and then switch it later.

Do you have better solution to support the above case at this time?

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: If the code is totally migrated by DPCT, the tool will maintain all queues and make sure queues are all in-ordered (in USM case).
If user manually changes the migrated code, user needs to make sure the input queue of this API is in-ordered. We have added document for this parameter above this API.

clang/lib/DPCT/APINamesCUSPARSE.inc
gajanan-choudhary: So I see what's happening now: There's a difference between how cuSPARSE finalizes things and how oneMKL does it. In case of cuSPARSE and oneMKL, the work estimation stages match up, but the compute and finalize/spgemm_copy stages don't exactly match. The problem is that oneMKL finalizes/fills up the `row_ptr` of the C matrix in the compute stage itself, and finalizes/fills up the `cols`, `values` arrays of the C matrix in the finalize stage, but on the other hand, cuSPARSE finalizes/copies/fills up the C matrix in a single step in the end in their copy step. The steps for using oneMKL are:
```
sparse::matrix_handle_t matC;
init_matrix_handle(&matC);
malloc_device(row_ptr, sizeof(IntType) * (nrows+1));
sparse::set_csr_data(matC, rows, cols, row_ptr, NULL, NULL, ...); // row_ptr is necessary here
sparse::matmat(..., matmat_request::get_work_estimation_buf_size);
// Allocate tempBuffer1
sparse::matmat(..., matmat_request::work_estimation);
sparse::matmat(..., matmat_request::get_compute_buf_size);
sparse::matmat(..., matmat_request::compute); // Finalizes/copies row_ptr in C handle
sparse::matmat(..., matmat_request::get_nnz);
malloc_device(&cols, sizeof(IntType) * nnz);
malloc_device(&values, sizeof(DataType) * nnz);
sparse::set_csr_data(matC, rows, cols, row_ptr, cols, values, ...); // row_ptr is repeated here
sparse::matmat(..., matmat_request::finalize); // Finalizes/copies cols and values in C handle
```

gajanan-choudhary: I don't know how much control you have over the mapping and porting over in the DPCT tool, but note that we might not be able to change how we do these steps in oneMKL. The oneMKL `matmat` [documentation](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/oneapi-mkl-sparse-matmat.html) says in a table there:
<html>
<body>


compute/compute_structure | do internal products for computing the C matrix including the calculation of size of C matrix data and filling the row pointer array for C.
-- | --



</body>
</html>

zhiweij1: Then I think the current solution is: DPCT emit warning and throw exception for NULL row_ptr. If oneMKL changes the behavior in the future, we can update the helper function.

zhiweij1: I updated the PR.

clang/lib/DPCT/APINamesCUSPARSE.inc
```diff
@@ -149,13 +149,41 @@ FEATURE_REQUEST_FACTORY(HelperFeatureEnum::device_ext,
...
-ASSIGNABLE_FACTORY(ASSIGN_FACTORY_ENTRY(
-    "cusparseCreateCsr", DEREF(0),
-    CALL("std::make_shared<" + MapNames::getDpctNamespace() +
-             "sparse::sparse_matrix_desc>",
-         ARG(1), ARG(2), ARG(3), ARG(4), ARG(5), ARG(6), ARG(7), ARG(8), ARG(9),
-         ARG(10),
-         ARG(MapNames::getDpctNamespace() + "sparse::matrix_format::csr"))))
+CONDITIONAL_FACTORY_ENTRY(
+    [](const CallExpr *C) -> bool {
+      const Expr *E = C->getArg(4)->IgnoreImpCasts();
```
zhiweij1: move to a utility function.

clang/lib/DPCT/APINamesCUSPARSE.inc
```diff
@@ -149,13 +149,41 @@ FEATURE_REQUEST_FACTORY(HelperFeatureEnum::device_ext,
...
+      Expr::EvalResult ExprResult;
+      if ((!E->isValueDependent()) &&
+          (E->EvaluateAsInt(ExprResult, dpct::DpctGlobalInfo::getContext()))) {
+        int64_t Value = ExprResult.Val.getInt().getExtValue();
+        if (!Value)
+          return true;
+      }
+      if (dyn_cast<CXXNullPtrLiteralExpr>(E) || dyn_cast<GNUNullExpr>(E)) {
```
zhiweij1: dyn_cast => is_a

clang/lib/DPCT/CallExprRewriter.h
```diff
@@ -651,6 +651,42 @@ void printCapture(StreamT &Stream, bool IsCaptureRef) {
     Stream << "[=]";
 }
 
+class AddrOfExpr {
+  const Expr *E = nullptr;
+  const CallExpr *C = nullptr;
+
+  template <class StreamT> void print(StreamT &Stream, ExprAnalysis &EA) const {
+    Stream << "&";
```
zhiweij1: if the E is *ABC, the result should not be "&*ABC"

clang/lib/DPCT/CallExprRewriterCUSPARSE.cpp
```diff
@@ -11,6 +11,49 @@
...
+class NullptrOrCallArgExpr {
+  NullptrOrCallArgExpr() = default;
+
+public:
+  const Expr *E = nullptr;
+
+  template <class StreamT> void print(StreamT &Stream) const {
+    const Expr *Arg = E->IgnoreImpCasts();
+    Expr::EvalResult ArgResult;
+    if ((!Arg->isValueDependent()) &&
```
zhiweij1: Use the extracted utility function

clang/lib/DPCT/CallExprRewriterCUSPARSE.cpp
```diff
@@ -11,6 +11,49 @@
...
+      Stream << "nullptr";
+      return;
+    }
+    clang::dpct::print(Stream, E);
+  }
+
+  static NullptrOrCallArgExpr create(const Expr *E);
+};
+
+NullptrOrCallArgExpr NullptrOrCallArgExpr::create(const Expr *E) {
```
zhiweij1: uncecessary

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -425,7 +429,10 @@ class sparse_matrix_desc {
   /// Set the value pointer of this descriptor
   /// \param [in] value The input value pointer
   void set_value(void *value) {
-    // Assume the new data is different from the old data
+    if (value) {
```
zhiweij1:
```suggestion
    if (_value || !value) {
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -365,6 +366,9 @@ class sparse_matrix_desc {
     if (_data_format != matrix_format::csr) {
       throw std::runtime_error("the sparse matrix data format is unsupported");
     }
+    if (!_row_ptr) {
+      throw std::runtime_error("the row_ptr is NULL");
+    }
     oneapi::mkl::sparse::init_matrix_handle(&_matrix_handle);
     construct();
```
zhiweij1:
```suggestion
    set_data();
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -499,7 +506,17 @@ class sparse_matrix_desc {
...
+      throw std::runtime_error("the col_ind pointer can be reset only when the "
+                               "original pointer is NULL");
+    }
+    if (_value) {
+      throw std::runtime_error("the value pointer can be reset only when the "
+                               "original pointer is NULL");
+    }
     _row_ptr = row_ptr;
     _col_ind = col_ind;
     _value = value;
```
zhiweij1: call set_value() here, and add a comment to note the descriptor will be updated in the set_value

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +766,311 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+/// \param [in] temp_buffer Specifies the memory of the workspace.
+inline void spgemm_compute(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                           oneapi::mkl::transpose trans_b, const void *alpha,
+                           sparse_matrix_desc_t a, sparse_matrix_desc_t b,
+                           const void *beta, sparse_matrix_desc_t c,
+                           oneapi::mkl::sparse::matmat_descr_t matmat_descr,
+                           size_t *size_temp_buffer, void *temp_buffer) {
+  if (temp_buffer) {
+#ifdef DPCT_USM_LEVEL_NONE
+    sycl::buffer<std::uint8_t> extBuf =
```
zhiweij1: add a helper class to wrap the temp memory

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +766,311 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+        dpct::detail::get_memory(reinterpret_cast<float *>(y->get_value()));
+    oneapi::mkl::sparse::trsv(queue, uplo, trans_a, diag,
+                              a->get_matrix_handle(), data_x, data_y);
+    break;
+  }
+  case dpct::detail::get_type_combination_id(library_data_t::real_double,
+                                             library_data_t::real_double,
+                                             library_data_t::real_double): {
+    auto data_x =
+        dpct::detail::get_memory(reinterpret_cast<double *>(x->get_value()));
```
zhiweij1: add a template version of get_memory accept void * and do reinterpret_cast in it.

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +766,311 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+inline void spsv(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                 const void *alpha, sparse_matrix_desc_t a,
+                 std::shared_ptr<dense_vector_desc> x,
+                 std::shared_ptr<dense_vector_desc> y,
+                 library_data_t data_type) {
+  switch (data_type) {
+  case library_data_t::real_float:
+    detail::check_alpha_value<float>(alpha, queue);
+    break;
+  case library_data_t::real_double:
```
zhiweij1: merge check to the below switch

clang/lib/DPCT/CallExprRewriterCUSPARSE.cpp
zhiweij1: done

clang/lib/DPCT/APINamesCUSPARSE.inc
zhiweij1: done

clang/lib/DPCT/APINamesCUSPARSE.inc
zhiweij1: done

clang/lib/DPCT/CallExprRewriterCUSPARSE.cpp
zhiweij1: done

clang/lib/DPCT/CallExprRewriter.h
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/lib/DPCT/CallExprRewriter.h
```diff
@@ -651,6 +651,61 @@ void printCapture(StreamT &Stream, bool IsCaptureRef) {
     Stream << "[=]";
 }
 
+class AddrOfExpr {
```
zhiweij1, line 654: Emit warning when input expr is rvalue

clang/lib/DPCT/CallExprRewriterCUSPARSE.cpp
```diff
@@ -11,6 +11,47 @@
 
 namespace clang {
 namespace dpct {
+
+bool isNull(const Expr *Expr) {
+  const auto *E = Expr->IgnoreImpCasts();
+  Expr::EvalResult ExprResult;
+  if ((!E->isValueDependent()) &&
+      (E->EvaluateAsInt(ExprResult, dpct::DpctGlobalInfo::getContext()))) {
+    int64_t Value = ExprResult.Val.getInt().getExtValue();
```
zhiweij1: move to line 21

zhiweij1: done

clang/lib/DPCT/CallExprRewriter.cpp
```diff
@@ -44,6 +44,34 @@ std::function<bool(const CallExpr *C)> hasManagedAttr(int Idx) {
...
+  this->C = C;
+  this->E = getAddrOfedExpr(E);
+  if (this->E) {
+    this->E = this->E->IgnoreParens();
+    this->DerefRemoved = true;
+  } else {
+    this->E = E;
+  }
+
+  if (!E->IgnoreImpCasts()->isLValue()) {
```
zhiweij1: use conditional rewriter

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -499,11 +504,21 @@ class sparse_matrix_desc {
...
+      throw std::runtime_error("the row_ptr cannot be changed");
+    }
+    if (!col_ind) {
+      throw std::runtime_error("the col_ind pointer is NULL");
+    }
+    if (_col_ind) {
+      throw std::runtime_error("the col_ind pointer can be reset only when the "
+                               "original pointer is NULL");
+    }
     _row_ptr = row_ptr;
```
zhiweij1, line 555:
```suggestion

```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +757,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
     throw std::runtime_error("the combination of data type is unsupported");
   }
 }
+
+namespace detail {
+template <typename T> struct temp_memory {
+  temp_memory(sycl::queue queue) : _queue(queue) {}
+  void set_host_ptr(T *host_ptr) {
```
zhiweij1:
```suggestion
  void set_host_ptr(T host_value) {
```

zhiweij1: or as an input arg of ctor

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +757,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+    _is_host_memory = false;
+#ifdef DPCT_USM_LEVEL_NONE
+    sycl::buffer<T, 1> *_memory_ptr =
+        new sycl::buffer<T, 1>(std::move(dpct::get_buffer<T>(device_ptr)));
+#else
+    _memory_ptr = static_cast<T *>(device_ptr);
+#endif
+  }
+  ~temp_memory() {
+    if (_is_host_memory.has_value()) {
```
zhiweij1: add wait in this d-ctor

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +757,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+template <typename T> struct temp_memory {
+  temp_memory(sycl::queue queue) : _queue(queue) {}
+  void set_host_ptr(T *host_ptr) {
+    if (_is_host_memory.has_value()) {
+      throw std::runtime_error("memory pointer has been set");
+    }
+    _original_host_ptr = host_ptr;
+    _is_host_memory = true;
+#ifdef DPCT_USM_LEVEL_NONE
+    sycl::buffer<T, 1> *_memory_ptr =
```
zhiweij1: let this buffer as member

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +757,316 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+    oneapi::mkl::sparse::matmat(
+        queue, a->get_matrix_handle(), b->get_matrix_handle(),
+        c->get_matrix_handle(),
+        oneapi::mkl::sparse::matmat_request::get_work_estimation_buf_size,
+        matmat_descr, size_memory.get_memory_ptr(), nullptr
+#ifndef DPCT_USM_LEVEL_NONE
+        , {}
+#endif
+    );
+    queue.wait();
```
zhiweij1: No need wait for buffer

clang/lib/DPCT/CallExprRewriter.h
zhiweij1: done

clang/lib/DPCT/CallExprRewriter.cpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,21 +526,24 @@ class sparse_matrix_desc {
   /// Get the uplo attribute
   /// \return uplo value
   std::optional<oneapi::mkl::uplo> get_uplo() const noexcept { return _uplo; }
+  /// Set the number of non-zero elements
+  /// \param nnz [in] The number of non-zero elements.
+  void set_nnz(std::int64_t nnz) { _nnz = nnz; }
```
zhiweij1:
```suggestion
  void set_nnz(std::int64_t nnz) noexcept { _nnz = nnz; }
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,21 +526,24 @@ class sparse_matrix_desc {
...
+  /// Get the type of the value pointer.
+  /// \return The type of the value pointer.
+  library_data_t get_value_type() { return _value_type; }
```
zhiweij1:
```suggestion
  library_data_t get_value_type() const noexcept { return _value_type; }
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+namespace detail {
+template <typename T, bool is_host_memory> struct temp_memory {
+  temp_memory(sycl::queue queue, void *ptr)
+      : _queue(queue)
+#ifdef DPCT_USM_LEVEL_NONE
+        ,
+        _buffer(
+            is_host_memory
+                ? sycl::buffer<T, 1>(static_cast<T *>(ptr), sycl::range<1>(1))
+                : sycl::buffer<T, 1>(std::move(dpct::get_buffer<T>(ptr))))
```
zhiweij1:
```suggestion
                : sycl::buffer<T, 1>(dpct::get_buffer<T>(ptr)))
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+      _memory_ptr = static_cast<T *>(ptr);
+#endif
+    }
+  }
+
+  ~temp_memory() {
+#ifndef DPCT_USM_LEVEL_NONE
+    if constexpr (is_host_memory) {
+      T temp;
+      _queue.wait();
```
zhiweij1, line 787:
```suggestion

```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+    } else {
+      _memory_ptr = static_cast<T *>(ptr);
+#endif
+    }
+  }
+
+  ~temp_memory() {
+#ifndef DPCT_USM_LEVEL_NONE
+    if constexpr (is_host_memory) {
+      T temp;
```
zhiweij1:
```suggestion

```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+      _queue.wait();
+      _queue.memcpy(&temp, _memory_ptr, sizeof(T)).wait();
```
zhiweij1:
```suggestion

```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+      *_original_host_ptr = static_cast<std::size_t>(temp);
```
zhiweij1:
```suggestion
      *_original_host_ptr = static_cast<std::size_t>(*_memory_ptr);
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+/// \param [in, out] size_temp_buffer Specifies the size of workspace.
+/// \param [in] temp_buffer Specifies the memory of the workspace.
+inline void
+spgemm_work_estimation(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                       oneapi::mkl::transpose trans_b, const void *alpha,
+                       sparse_matrix_desc_t a, sparse_matrix_desc_t b,
+                       const void *beta, sparse_matrix_desc_t c,
+                       oneapi::mkl::sparse::matmat_descr_t matmat_descr,
+                       size_t *size_temp_buffer, void *temp_buffer) {
+  std::int64_t size_temp_buffer_64;
```
zhiweij1, line 835: remove

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,309 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+/// to be executed.
+/// \param [in, out] size_temp_buffer Specifies the size of workspace.
+/// \param [in] temp_buffer Specifies the memory of the workspace.
+inline void spgemm_compute(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                           oneapi::mkl::transpose trans_b, const void *alpha,
+                           sparse_matrix_desc_t a, sparse_matrix_desc_t b,
+                           const void *beta, sparse_matrix_desc_t c,
+                           oneapi::mkl::sparse::matmat_descr_t matmat_descr,
+                           size_t *size_temp_buffer, void *temp_buffer) {
+  std::int64_t size_temp_buffer_64;
```
zhiweij1, line 891: remove

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -286,6 +281,7 @@ class dense_vector_desc {
   }
   void *get_value() const noexcept { return _value; }
   void set_value(void *value) { _value = value; }
+  library_data_t get_value_type() { return _value_type; }
```
zhiweij1:
```suggestion
  library_data_t get_value_type() const noexcept { return _value_type; }
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
     throw std::runtime_error("the combination of data type is unsupported");
   }
 }
+
+namespace detail {
+template <typename T, bool is_host_memory, typename host_memory_t = void>
```
zhiweij1, line 761:
```suggestion
template <typename T, bool is_host_memory, typename host_memory_t = size_t>
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+    detail::temp_memory<std::int64_t, true, size_t> size_memory(
+        queue, size_temp_buffer);
+    detail::temp_memory<std::uint8_t, false> work_memory(queue, temp_buffer);
+    oneapi::mkl::sparse::matmat(
+        queue, a->get_matrix_handle(), b->get_matrix_handle(),
+        c->get_matrix_handle(),
+        oneapi::mkl::sparse::matmat_request::work_estimation, matmat_descr,
+        size_memory.get_memory_ptr(), work_memory.get_memory_ptr()
+#ifndef DPCT_USM_LEVEL_NONE
+                                          ,
```
zhiweij1: manually format

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
     throw std::runtime_error("the combination of data type is unsupported");
   }
 }
+
+namespace detail {
+template <typename T, bool is_host_memory, typename host_memory_t = void>
+struct temp_memory {
```
zhiweij1, line 867: static_assert

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+        oneapi::mkl::sparse::matrix_view_descr::general);
+    detail::temp_memory<std::int64_t, true, size_t> size_memory(
+        queue, size_temp_buffer);
+    oneapi::mkl::sparse::matmat(
+        queue, a->get_matrix_handle(), b->get_matrix_handle(),
+        c->get_matrix_handle(),
+        oneapi::mkl::sparse::matmat_request::get_work_estimation_buf_size,
+        matmat_descr, size_memory.get_memory_ptr(), nullptr
+#ifndef DPCT_USM_LEVEL_NONE
+        ,
```
zhiweij1: manually format

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+      detail::temp_memory<std::uint8_t, false> work_memory(queue, temp_buffer);
+      detail::temp_memory<std::int64_t, true, std::int64_t> nnz_memory(
+          queue, &nnz_value);
+      oneapi::mkl::sparse::matmat(
+          queue, a->get_matrix_handle(), b->get_matrix_handle(),
+          c->get_matrix_handle(), oneapi::mkl::sparse::matmat_request::compute,
+          matmat_descr, size_memory.get_memory_ptr(),
+          work_memory.get_memory_ptr()
+#ifndef DPCT_USM_LEVEL_NONE
+              ,
```
zhiweij1: manually format

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+          {}
+#endif
+      );
+      oneapi::mkl::sparse::matmat(
+          queue, a->get_matrix_handle(), b->get_matrix_handle(),
+          c->get_matrix_handle(), oneapi::mkl::sparse::matmat_request::get_nnz,
+          matmat_descr, nnz_memory.get_memory_ptr(), nullptr
+#ifndef DPCT_USM_LEVEL_NONE
+          ,
```
zhiweij1: manually format

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +756,314 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+  } else {
+    detail::temp_memory<std::int64_t, true, size_t> size_memory(
+        queue, size_temp_buffer);
+    oneapi::mkl::sparse::matmat(
+        queue, a->get_matrix_handle(), b->get_matrix_handle(),
+        c->get_matrix_handle(),
+        oneapi::mkl::sparse::matmat_request::get_compute_buf_size, matmat_descr,
+        size_memory.get_memory_ptr(), nullptr
+#ifndef DPCT_USM_LEVEL_NONE
+        ,
```
zhiweij1: manually format

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -425,9 +442,15 @@ class sparse_matrix_desc {
   /// Set the value pointer of this descriptor
   /// \param [in] value The input value pointer
   void set_value(void *value) {
-    // Assume the new data is different from the old data
+    if (!value) {
+      throw std::runtime_error("the new value pointer is NULL");
```
zhimingwang36:
```suggestion
      throw std::runtime_error("dpct::sparse_matrix_desc::set_value(): The new value pointer is NULL");
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -425,9 +442,15 @@ class sparse_matrix_desc {
...
+    }
+    if (_value) {
+      throw std::runtime_error("the value pointer can be reset only when the "
+                               "original pointer is NULL");
```
zhimingwang36: dpct::...: ,  The value pointer is not NULL, can not be reset.

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +803,312 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+}
+
+namespace detail {
+template <typename T>
+inline void check_alpha_value(const void *alpha, sycl::queue queue) {
+  auto alpha_value =
+      dpct::detail::get_value(reinterpret_cast<const T *>(alpha), queue);
+  if (alpha_value != T(1.0f)) {
+    throw std::runtime_error(
+        "oneapi::mkl::sparse::trsv only supports alpha == 1.");
```
zhimingwang36: Please try to remove this runtime_error() for end user if alpha is not 1.

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -743,6 +803,312 @@ inline void spmm(sycl::queue queue, oneapi::mkl::transpose trans_a,
...
+/// Performs internal optimizations for spsv by analyzing the provided matrix
+/// structure and operation parameters.
+/// \param [in] queue The queue where the routine should be executed. It must
+/// have the in_order property when using the USM mode.
+/// \param [in] trans_a Specifies operation on input matrix a.
+/// \param [in] a Specifies the sparse matrix a.
+inline void spsv_optimize(sycl::queue queue, oneapi::mkl::transpose trans_a,
+                          sparse_matrix_desc_t a) {
+  if (!a->get_uplo() || !a->get_diag()) {
+    throw std::runtime_error("oneapi::mkl::sparse::trsv needs uplo and diag "
```
zhimingwang36: Please figure out mapping for L1035,  it's not friend for end user.

zhimingwang36 (commented):
Please split the PR into refine part and new enabled part.

clang/test/dpct/cusparse_generic.cu
```diff
@@ -132,5 +132,89 @@ int main() {
...
+  void *csrRowOffsets;
+  void *csrColInd;
+  void *csrValues;
+  cusparseIndexType_t csrRowOffsetsType;
+  cusparseIndexType_t csrColIndType;
+  cusparseIndexBase_t idxBase;
+  cudaDataType valueType;
+
+  //CHECK:/*
+  //CHECK-NEXT:DPCT1119:{{[0-9]+}}: The memory of row_ptr is not allocated. You need to allocate the memory before invoking oneapi::mkl::sparse::set_csr_data().
```
zhimingwang36: we can hold a internal buffer and sync with the buffer caller specified.  (then we don't need DPCT1119).

clang/test/dpct/cusparse_generic.cu
```diff
@@ -132,5 +132,89 @@ int main() {
...
+void foo2(cusparseSpGEMMDescr_t *descr) {
+  cusparseSpGEMM_destroyDescr(*descr);
+}
+
+//CHECK:oneapi::mkl::sparse::matmat_descr_t foo3();
+//CHECK-NEXT:void foo4() {
+//CHECK-NEXT:  /*
+//CHECK-NEXT:  DPCT1120:{{[0-9]+}}: Taking the address of an rvalue "foo3()" is not allowed. You need to adjust the code.
+//CHECK-NEXT:  */
+//CHECK-NEXT:  oneapi::mkl::sparse::release_matmat_descr(&foo3());
```
zhimingwang36: Please consider remove DPCT1120.

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: I checked the cusparse doc. cusparseSpSV is only support for triangle matrix. So if the SYCL code is directly migrated from CUDA without user manually changes, this exception will not be thrown.

clang/test/dpct/cusparse_generic.cu
zhiweij1: Done

clang/runtime/dpct-rt/include/sparse_utils.hpp
zhiweij1: done

clang/test/dpct/cusparse_generic.cu
zhiweij1: done

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -499,11 +530,21 @@ class sparse_matrix_desc {
   /// numbering.
   /// \param [in] value An array containing the non-zero elements of the sparse matrix.
   void set_pointers(void *row_ptr, void *col_ind, void *value) {
-    // Assume the new data is different from the old data
+    if (!col_ind) {
+      throw std::runtime_error(
+          "dpct::sparse::sparse_matrix_desc::set_pointers(): The "
+          "new col_ind pointer is NULL.");
```
zhiweij1:
```suggestion
          "col_ind pointer is NULL.");
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -425,9 +447,18 @@ class sparse_matrix_desc {
   /// Set the value pointer of this descriptor
   /// \param [in] value The input value pointer
   void set_value(void *value) {
-    // Assume the new data is different from the old data
+    if (!value) {
+      throw std::runtime_error(
+          "dpct::sparse::sparse_matrix_desc::set_value(): The new "
```
zhiweij1:
```suggestion
          "dpct::sparse::sparse_matrix_desc::set_value(): The "
```

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,76 +553,128 @@ class sparse_matrix_desc {
...
+  void set_nnz(std::int64_t nnz) noexcept { _nnz = nnz; }
+  /// Get the type of the value pointer.
+  /// \return The type of the value pointer.
+  library_data_t get_value_type() const noexcept { return _value_type; }
+  /// Get the row_ptr.
+  /// \return The row_ptr.
+  void *get_row_ptr() const noexcept { return _row_ptr; }
+  /// Get the temp_row_ptr.
+  /// \return The temp_row_ptr.
+  void *get_temp_row_ptr() const noexcept { return _temp_row_ptr; }
```
zhimingwang36: how about get_shadow_row_ptr?

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -598,6 +691,17 @@ class sparse_matrix_desc {
   matrix_format _data_format;
   std::optional<oneapi::mkl::uplo> _uplo;
   std::optional<oneapi::mkl::diag> _diag;
+  void *_temp_row_ptr = nullptr;
```
zhimingwang36: Please add comment to tell the relation ship between _temp_row_ptr and _row_ptr.

zhiweij1: done

zhimingwang36 (approved):
LGTM

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -366,13 +370,33 @@ class sparse_matrix_desc {
...
-    construct();
+    set_data();
   }
   /// Destructor
   ~sparse_matrix_desc() {
     oneapi::mkl::sparse::release_matrix_handle(get_default_queue(),
                                                &_matrix_handle, _deps)
         .wait();
+#ifdef DPCT_USM_LEVEL_NONE
+    if (_data_row_ptr_32)
```
zhiweij1, line 381: can use unique_ptr

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -366,13 +370,33 @@ class sparse_matrix_desc {
...
+    if (_data_value_s)
+      delete _data_value_s;
+    if (_data_value_d)
+      delete _data_value_d;
+    if (_data_value_c)
+      delete _data_value_c;
+    if (_data_value_z)
+      delete _data_value_z;
+#endif
+    if (_shadow_row_ptr)
```
zhiweij1, line 398: can use unique_ptr with custom deleter

clang/runtime/dpct-rt/include/sparse_utils.hpp
```diff
@@ -512,76 +565,130 @@ class sparse_matrix_desc {
...
+    void *row_ptr = nullptr;                                                   \
+    if (_shadow_row_ptr) {                                                     \
+      row_ptr = _shadow_row_ptr;                                               \
+    } else if (_row_ptr) {                                                     \
+      row_ptr = _row_ptr;                                                      \
+    } else {                                                                   \
+      row_ptr = _shadow_row_ptr = dpct::dpct_malloc(                           \
+          sizeof(INDEX_TYPE) * (_row_num + 1), get_default_queue());           \
+    }                                                                          \
+    _data_row_ptr##INDEX_SUFFIX =                                              \
```
zhiweij1, line 602:
```
union {
  char placeholer[??];
  buffer<t1> b1;
  buffer<t2> b2;
};
or std::variant;
```

zhiweij1: can shadow_ptr use set_final_data replace?

PR #1146, opened 2023-08-03
[SYCLomatic][BSS] Adding python script to scan binaries repo

tovinkere:
+ Adding new binary scanning capability to scan binaries for function calls from performance libraries
+ Allows building logic to compare functions in a binary and determine conversion success rate without having to invoke a compile step

### DIFF
--- a/clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
+++ b/clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
@@ -1,7 +1,7 @@
 #--------------------------------------------------------------------------------
 # MIT License
 # 
-# Copyright (c) 2023 Intel Corporation
+# Copyright (c) Intel Corporation
 # 
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -64,8 +64,7 @@ else:
         os_name = "Linux"
         command = ["nm", "--demangle"]
     else:
-        os_name = "Unknown"
-        command = ["unknown"]
+        sys.exit("Running script on an unsupported OS: Aborting! [Supported OSes - Windows, Linux]")
 
 
 # Regular expression dictionary for all the function signatures from Intel oneAPI copmponents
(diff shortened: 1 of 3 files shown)


clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
```diff
@@ -0,0 +1,446 @@
+#--------------------------------------------------------------------------------
+# MIT License
+# 
+# Copyright (c) 2023 Intel Corporation
```
zhimingwang36: I think Year info is not required, and to avoid update the year info after updating the script, suggest remove the year info.

clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
```diff
@@ -0,0 +1,446 @@
...
+    for line in result.stdout.decode().splitlines():
+        if debug_output:
+            out_text = "========== All Functions:  '{}'"
+            print(out_text.format(line))
+        if not re.search(".@GLIB|.so", line):
+            tokens = line.strip().split()
+            token_count = len(tokens)
+            if token_count > 0:
+                if re.match("U|u|w|W", tokens[0]):
+                #if tokens[0] == "U" or tokens[0] == "T" or tokens[0] == "w" or tokens[0] == "W":
```
zhimingwang36: Suggest to remove dead code in L187

clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
```diff
@@ -0,0 +1,446 @@
...
+if os.name == "nt":
+    os_name = "Windows"
+    command = ["dumpbin", "/imports"]
+else:
+    if os.name == "posix":
+        os_name = "Linux"
+        command = ["nm", "--demangle"]
+    else:
+        os_name = "Unknown"
+        command = ["unknown"]
```
zhimingwang36: L68,  i think it's better to exit, if the OS is not supported, like MAC.

clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
tovinkere: Done!

clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
tovinkere: Done!

clang/tools/binary-symbol-scanner/BinarySymbolScanner.py
tovinkere: Done!

zhimingwang36 (approved):
LGTM

PR #1165, opened 2023-08-14
[SYCLomatic] Add malloc, free, and temporary allocation mappings

mmichel11:
In this PR, we add the APIs for `malloc`, `free`, and temporary buffers (`get_temporary_allocation` and `release_temporary_allocation`). A `tagged_pointer` type is added that contains a template parameter of type `host_sys_tag` or `device_sys_tag` specifying the location of the allocation. This type is returned by `malloc` and `get_temporary_allocation`. `tagged_pointer` currently provides basic pointer functionality and will be improved in the future once we get access to Boost iterator adaptors.

In addition to passing execution policies to the memory allocation APIs, passing a system tag type is also supported in this PR. The implemented functionality is only supported for USM mode. Static asserts are hit if `DPCT_USM_LEVEL_NONE` is set.

There are two PRs that test this functionality:
- https://github.com/oneapi-src/SYCLomatic-test/pull/426
- https://github.com/oneapi-src/SYCLomatic-test/pull/427

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h
@@ -745,38 +745,22 @@ inline constexpr bool is_host_policy_or_tag_v =
 
 } // namespace internal
 
-// TODO: Make this class an iterator adaptor
-template <typename Tag, typename T> class tagged_pointer_base {
-public:
-  using pointer = T *;
-  using difference_type = ::std::ptrdiff_t;
-  tagged_pointer_base() : m_ptr(nullptr) {}
-  tagged_pointer_base(T *ptr) : m_ptr(ptr) {}
-  operator const T *() const { return m_ptr; }
-  operator T *() { return m_ptr; }
-
-protected:
-  T *m_ptr;
-};
-
+// TODO: Make this class an iterator adaptor.
 // tagged_pointer provides a wrapper around a raw pointer type with a tag of the
 // location of the allocated memory. Standard pointer operations are supported
 // with this class.
-template <typename Tag, typename T = void>
-class tagged_pointer : public tagged_pointer_base<Tag, T> {
-  using super_t = tagged_pointer_base<Tag, T>;
-
+template <typename Tag, typename T = void> class tagged_pointer {
 public:
   using value_type = T;
-  using difference_type = typename super_t::difference_type;
-  using pointer = typename super_t::pointer;
+  using difference_type = ::std::ptrdiff_t;
+  using pointer = T *;
   using reference = T &;
   using iterator_category = std::random_access_iterator_tag;
   using is_hetero = ::std::false_type;
   using is_passed_directly = std::true_type;
 
-  tagged_pointer() : super_t() {}
-  tagged_pointer(T *ptr) : super_t(ptr) {}
+  tagged_pointer() : m_ptr(nullptr) {}
+  tagged_pointer(T *ptr) : m_ptr(ptr) {}
   T &operator[](difference_type idx) { return this->m_ptr[idx]; }
   const T &operator[](difference_type idx) const { return this->m_ptr[idx]; }
   tagged_pointer operator+(difference_type forward) const {
@@ -785,6 +769,8 @@ public:
   tagged_pointer operator-(difference_type backward) const {
     return tagged_pointer{this->m_ptr - backward};
   }
+  operator const T *() const { return m_ptr; }
+  operator T *() { return m_ptr; }
   T &operator*() { return *this->m_ptr; }
   const T &operator*() const { return *this->m_ptr; }
   T *operator->() { return this->m_ptr; }
@@ -818,24 +804,29 @@ public:
     this->m_ptr = this->m_ptr - backward;
     return *this;
   }
+
+private:
+  T *m_ptr;
 };
 
 // Void specialization for tagged pointers. Iterator traits are not provided but
 // conversion to other non-void tagged pointers is allowed. Pointer arithmetic
 // is disallowed with this specialization.
-template <typename Tag>
-class tagged_pointer<Tag, void> : public tagged_pointer_base<Tag, void> {
-  using super_t = tagged_pointer_base<Tag, void>;
-
+template <typename Tag> class tagged_pointer<Tag, void> {
 public:
-  using difference_type = typename super_t::difference_type;
-  using pointer = typename super_t::pointer;
-  tagged_pointer() : super_t() {}
-  tagged_pointer(pointer ptr) : super_t(ptr) {}
+  using difference_type = ::std::ptrdiff_t;
+  using pointer = void *;
+  tagged_pointer() : m_ptr(nullptr) {}
+  tagged_pointer(pointer ptr) : m_ptr(ptr) {}
+  operator const void *() const { return m_ptr; }
+  operator void *() { return m_ptr; }
   // Enable tagged void pointer to convert to all other raw pointer types.
   template <typename OtherPtr> operator OtherPtr *() const {
     return static_cast<OtherPtr *>(this->m_ptr);
   }
+
+private:
+  void *m_ptr;
 };
 
 namespace internal {
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h
@@ -763,21 +763,13 @@ public:
   using pointer = typename super_t::pointer;
   using reference = T &;
   using iterator_category = std::random_access_iterator_tag;
-  using is_hetero = ::std::is_same<Tag, host_sys_tag>; // required
-  using is_passed_directly = std::false_type;
+  using is_hetero = ::std::false_type;
+  using is_passed_directly = std::true_type;
 
   tagged_pointer() : super_t() {}
   tagged_pointer(T *ptr) : super_t(ptr) {}
   T &operator[](difference_type idx) { return this->m_ptr[idx]; }
   T &operator[](difference_type idx) const { return this->m_ptr[idx]; }
-  // Enable conversion to other pointer types only if it is convertible
-  template <
-      typename OtherType,
-      ::std::enable_if_t<::std::is_convertible_v<T *, OtherType *>, int> = 0>
-  operator tagged_pointer<Tag, OtherType>() {
-    return tagged_pointer<Tag, OtherType>(
-        static_cast<OtherType *>(this->m_ptr));
-  }
   tagged_pointer operator+(difference_type forward) const {
     return tagged_pointer{this->m_ptr + forward};
   }
@@ -828,9 +820,12 @@ public:
   using pointer = typename super_t::pointer;
   tagged_pointer() : super_t() {}
   tagged_pointer(pointer ptr) : super_t(ptr) {}
-  // Enable void pointer to convert to all other pointer types.
-  template <typename OtherType> operator tagged_pointer<Tag, OtherType>() {
-    return static_cast<OtherType *>(this->m_ptr);
+  // Enable tagged void pointer to convert to all other raw pointer types.
+  template <typename OtherPtr> operator OtherPtr *() const {
+    return static_cast<OtherPtr *>(this->m_ptr);
+  }
+  template <typename OtherPtr> operator OtherPtr *() {
+    return static_cast<OtherPtr *>(this->m_ptr);
   }
 };
 
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h
@@ -719,8 +719,9 @@ private:
 public:
   static constexpr bool is_policy_v =
       oneapi::dpl::execution::is_execution_policy_v<decayed_policy_or_tag_t>;
-  static constexpr bool is_sys_tag_v =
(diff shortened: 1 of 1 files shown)


tomflinda (commented):
Pls add test case in https://github.com/oneapi-src/SYCLomatic-test/tree/SYCLomatic/help_function/src for new added helper APIs in this PR

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+
+} // namespace internal
+
+// TODO: Improve this pointer wrapper and make an iterator adaptor
+// once we have the Boost library dependency
+template <typename Tag, typename T, typename Derived>
+class tagged_pointer_base {
+public:
+  using pointer = T *;
+  using difference_type = ::std::make_signed_t<std::size_t>;
```
ziranzha: How about use `ptrdiff_t`?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+  tagged_pointer_base() = default;
```
ziranzha: `m_ptr` is not initialized.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+  using pointer = typename super_t::pointer;
+  using reference = T &;
+  using iterator_category = std::random_access_iterator_tag;
+  using is_hetero = ::std::false_type;
+  using is_passed_directly = std::true_type;
+
+  tagged_pointer() : super_t() {}
+  tagged_pointer(T *ptr) : super_t(ptr) {}
+  T &operator[](difference_type idx) { return this->m_ptr[idx]; }
+  T &operator[](difference_type idx) const { return this->m_ptr[idx]; }
```
ziranzha: Should return `const T &` here?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+    return this->m_ptr - it.m_ptr;
+  }
+  tagged_pointer &operator+=(difference_type forward) {
+    this->m_ptr = this->m_ptr + forward;
+    return *this;
+  }
+  tagged_pointer &operator-=(difference_type backward) {
+    this->m_ptr = this->m_ptr - backward;
+    return *this;
+  }
```
ziranzha, line 806: `operator*` and `operator->` and pointer type conversion are also needed?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+      ::std::conditional_t<oneapi::dpl::__internal::__is_host_execution_policy<
+                               decayed_policy_or_tag_t>::value,
+                           host_sys_tag, device_sys_tag>;
+
+public:
+  static constexpr bool is_policy_v =
+      oneapi::dpl::execution::is_execution_policy_v<decayed_policy_or_tag_t>;
+  static constexpr bool is_sys_tag_v =
+      ::std::is_base_of_v<sys_tag, std::decay_t<decayed_policy_or_tag_t>>;
+  static_assert(is_policy_v || is_sys_tag_v,
```
ziranzha, line 723: Is user-defined tag that deriving `sys_tag` also permitted?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,235 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+};
+
+template <typename PolicyOrTag>
+using policy_or_tag_to_tag_t = typename policy_or_tag_to_tag<PolicyOrTag>::type;
+
+} // namespace internal
+
+// TODO: Improve this pointer wrapper and make an iterator adaptor
+// once we have the Boost library dependency
+template <typename Tag, typename T, typename Derived>
```
ziranzha: Why we need this base class?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Yes, I have made this change.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Fixed.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: I have switched to using this.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: I have added this, so we are not relying on conversion to pointer. I also updated the test to check `operator*` and `operator->`.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: This base class just contains the shared functions / members between the standard `tagged_pointer` case and the specialization for void pointers. The CRTP template parameter is currently unused. I have removed it.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Since we do not currently support any custom execution policies in oneDPL, I think we should not currently allow user-defined system tags. I have put a stricter check in here to ensure this.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,237 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+}
+#else
+namespace internal {
+
+// Utility that converts a policy to a tag or reflects a provided tag
+template <typename PolicyOrTag> struct policy_or_tag_to_tag {
+private:
+  using decayed_policy_or_tag_t = ::std::decay_t<PolicyOrTag>;
+  using policy_conversion =
+      ::std::conditional_t<oneapi::dpl::__internal::__is_host_execution_policy<
```
danhoeflinger: We should do our best to stay away from using non-public API of oneDPL outside of oneDPL.  Here I believe there exists `dpct::is_hetero_execution_policy<>`, you can either negate this or add a similar host version to the SYCLomatic headers.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,237 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+                                    decayed_policy_or_tag_t>;
+};
+
+template <typename PolicyOrTag>
+using policy_or_tag_to_tag_t = typename policy_or_tag_to_tag<PolicyOrTag>::type;
+
+} // namespace internal
+
+// TODO: Improve this pointer wrapper and make an iterator adaptor
+// once we have the Boost library dependency
```
danhoeflinger: Probably don't need to mention boost here explicitly as that is still under discussion and an implementation detail, but rather just mention an iterator adaptor implementation

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,237 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+  auto allocation_ptr =
+      malloc<T>(::std::forward<PolicyOrTag>(policy_or_tag), num_elements);
+  if (allocation_ptr == nullptr)
+    return ::std::make_pair(allocation_ptr, SizeType(0));
+  return ::std::make_pair(allocation_ptr, num_elements);
+}
+
+template <typename PolicyOrTag, typename Pointer>
+void release_temporary_allocation(PolicyOrTag &&policy_or_tag, Pointer ptr) {
+  free(::std::forward<PolicyOrTag>(policy_or_tag), ptr);
```
danhoeflinger: May be good to explicitly use `dpct::free()` here so there is no confusion.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,237 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+  return tagged_pointer<internal::policy_or_tag_to_tag_t<PolicyOrTag>, T>(
+      static_cast<T *>(
+          internal::malloc_base(::std::forward<PolicyOrTag>(policy_or_tag),
+                                num_elements * sizeof(T))));
+}
+
+template <typename PolicyOrTag, typename Pointer>
+void free(PolicyOrTag &&policy_or_tag, Pointer ptr) {
+  using tag = internal::policy_or_tag_to_tag<PolicyOrTag>;
+  if constexpr (::std::is_same_v<typename tag::type, host_sys_tag>) {
```
danhoeflinger: Does it make sense to abstract this out to something like:
`dpct::is_host_policy_or_tag_v<PolicyOrTag>`?

If you do this, you'd have to adjust `tag::is_policy_v`  a bit as well, probably to use `oneapi::dpl::execution::is_execution_policy_v`

Similar above.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -670,6 +670,237 @@ template <typename T> class device_iterator : public device_pointer<T> {
...
+    }
+    sycl::free(ptr, q);
+  }
+}
+
+template <typename T, typename PolicyOrTag, typename SizeType>
+auto get_temporary_allocation(PolicyOrTag &&policy_or_tag,
+                              SizeType num_elements) {
+  auto allocation_ptr =
+      malloc<T>(::std::forward<PolicyOrTag>(policy_or_tag), num_elements);
```
danhoeflinger: May be good to explicitly use `dpct::malloc()` so there is no confusion.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: I have updated this to use the negation of `dpct::is_hetero_execution_policy`.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Done

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Updated

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: I have added `dpct::is_host_policy_or_tag_v<PolicyOrTag>` and switched `tag::is_policy_v` to just use `oneapi::dpl::execution::is_execution_policy_v` on the decayed `PolicyOrType`.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Updated

mmichel11:
@tomflinda I have added tests for the system tag internal helper APIs in: https://github.com/oneapi-src/SYCLomatic-test/pull/427/files and a test for the `malloc_base` helper API in https://github.com/oneapi-src/SYCLomatic-test/pull/426

danhoeflinger (approved):
LGTM, it would be good to get and approval from @ziranzha before merging.

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/memory.h
ziranzha: Do we need some type conversion like:
```
  template<class OtherT>
  operator tagged_pointer<Tag, OtherT>();
```

clang/runtime/dpct-rt/include/dpl_extras/memory.h
ziranzha: The base class just provide two simple methods:
```
  operator const T *() const { return m_ptr; }
  operator T *() { return m_ptr; }
```
I think we can remove it.

mmichel11: I have made this change.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: I don't think this operator is necessary since we already provide a raw pointer conversion to `otherT` which can then call the `tagged_pointer<Tag, OtherT>` constructor. This functionality is tested in `onedpl_test_tagged_pointer` in the test PR.

We also don't provide this operator in similar classes (ex: device_pointer).

ziranzha (approved):

PR #1168, opened 2023-08-15
[SYCLomatic] Save length of macro instead of SourceLocation to avoid internal error when SourceManager switched

zhiweij1:

### DIFF
--- a/clang/lib/DPCT/AnalysisInfo.cpp
+++ b/clang/lib/DPCT/AnalysisInfo.cpp
@@ -87,7 +87,7 @@ std::map<std::string, std::shared_ptr<DpctGlobalInfo::MacroDefRecord>>
 std::map<std::string, std::string>
     DpctGlobalInfo::FunctionCallInMacroMigrateRecord;
 std::map<std::string, SourceLocation> DpctGlobalInfo::EndOfEmptyMacros;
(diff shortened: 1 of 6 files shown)


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -87,7 +87,7 @@ std::map<std::string, std::shared_ptr<DpctGlobalInfo::MacroDefRecord>>
 std::map<std::string, std::string>
     DpctGlobalInfo::FunctionCallInMacroMigrateRecord;
 std::map<std::string, SourceLocation> DpctGlobalInfo::EndOfEmptyMacros;
-std::map<std::string, SourceLocation> DpctGlobalInfo::BeginOfEmptyMacros;
+std::map<std::string, unsigned int> DpctGlobalInfo::BeginOfEmptyMacros;
```
yihwang-nv, line 90: What about to use `unsigned`?

zhiweij1: `unsigned` and `unsigned int` are equivalent. I think both are OK here.

AndyCHHuang (approved):
LGTM

ziranzha (approved):

PR #1186, opened 2023-08-17
[SYCLomatic] Histogram functional implementation

danhoeflinger:
Provides reasonable performance functional implementation of Histogram APIs:

`dpct::histogram_even` - Evenly Divided Histogram of a 1-D array
`dpct::histogram_even_roi` - Evenly Divided Histogram of a 2-D ROI in a flattened 2-D array
`dpct::multi_histogram_even` -  Evenly Divided Multi-Channel Histogram of a 1-D array
`dpct::multi_histogram_even_roi` - Evenly Divided Multi-Channel Histogram of a 2-D ROI in a flattened 2-D array
`dpct::histogram_range` - Custom Range Histogram of a 1-D array
`dpct::histogram_range_roi` - Custom Range Histogram of a 2-D ROI in a flattened 2-D Array
`dpct::multi_histogram_range` -  Custom Range Multi-Channel Histogram of a 1-D array
`dpct::multi_histogram_range_roi` - Custom Range Multi-Channel Histogram of a 2-D ROI in a flattened 2-D array

note:
This PR includes the `dpct::internal::dpl_histogram` namespace which provides a prototype implementation of functionality intended to be provided by oneDPL.  In a future PR, this namespace will be removed and references to it will be replaced by calls to `oneapi::dpl::histogram`.  There are two instances of non-public oneDPL APIs used in this namespace to enable full functionality without large amounts or redundant code coming into the SYCLomatic headers.  These will be removed with the namespace when the histogram API is available from oneDPL.  These non-public APIs used are stable and should not be expected to change in the timeframe where this code should exist.

Tested by https://github.com/oneapi-src/SYCLomatic-test/pull/429

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -1801,11 +1801,10 @@ __histogram_general_select_best(_ExecutionPolicy &&policy,
       policy.queue()
           .get_device()
           .template get_info<sycl::info::device::local_mem_size>();
-
-  ::std::size_t N = __last - __first;
+  constexpr ::std::uint8_t __max_registers = 16;
 
   // if bins fit into registers, use register private accumulation
-  if (num_bins < MAX_BINS_PER_WORK_ITEM) {
+  if (num_bins < __max_registers) {
     return __histogram_general_registers_local_reduction<1024, 32, 16>(
         ::std::forward<_ExecutionPolicy>(policy), __first, __last,
         __histogram_first, num_bins, __func);
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -1533,18 +1533,20 @@ void __clear_wglocal_histograms(_HistAccessor local_histogram, _Size num_bins,
 }
 
 template <::std::uint16_t __work_group_size,
-          ::std::uint8_t __iters_per_work_item, typename _BinIdxType,
-          typename _Iter1, typename _HistAccessor, typename _MemorySpace,
-          typename _BinFunc>
+          typename _BinIdxType,
+          sycl::access::address_space _AddressSpace,
+          typename _Iter1, typename _HistAccessor,
+          typename _BinFunc, typename _SelfItem>
 void __accum_local_atomics(_Iter1 in_acc, const ::std::size_t &seg_start,
                            const ::std::size_t &N,
                            const ::std::uint32_t &self_lidx,
                            _HistAccessor wg_local_histogram,
-                           const _MemorySpace &memory_space, _BinFunc func) {
-  using __histo_value_type = ::std::iterator_traits<_HistAccessor>::value_type;
-  using __value_type = ::std::iterator_traits<_Iter1>::value_type;
+                           _BinFunc func, 
+                           _SelfItem self_item, const std::uint8_t& iters_per_work_item) {
+  using __histo_value_type = typename ::std::iterator_traits<_HistAccessor>::value_type;
+  using __value_type = typename ::std::iterator_traits<_Iter1>::value_type;
   const ::std::size_t __seg_end =
-      sycl::min(seg_start + __work_group_size * __iters_per_work_item, N);
+      sycl::min(seg_start + __work_group_size * iters_per_work_item, N);
 #pragma unroll
   for (::std::size_t __val_idx = seg_start + self_lidx; __val_idx < __seg_end;
        __val_idx += __work_group_size) {
@@ -1552,11 +1554,11 @@ void __accum_local_atomics(_Iter1 in_acc, const ::std::size_t &seg_start,
     _BinIdxType c = func(x);
 
     sycl::atomic_ref<__histo_value_type, sycl::memory_order::relaxed,
-                     sycl::memory_scope::work_group, memory_space>
-        global_bin(hacc_private[c]);
+                     sycl::memory_scope::work_group, _AddressSpace>
+        global_bin(wg_local_histogram[c]);
     global_bin++;
   }
-  __self_item.barrier(sycl::access::fence_space::local_space);
+  self_item.barrier(sycl::access::fence_space::local_space);
 }
 
 template <typename _HistAccessorIn, typename _HistAccessorOut, typename _Size,
@@ -1705,10 +1707,12 @@ Iter2 __histogram_general_local_atomics(Policy &&policy, _Iter1 __first,
           __clear_wglocal_histograms(&(local_histogram[0]), num_bins,
                                      __self_item);
 
-          __accum_local_atomics<__work_group_size, __iters_per_work_item,
-                                ::std::uint16_t>(
+          __accum_local_atomics<__work_group_size,
+                                ::std::uint16_t, 
+                                sycl::access::address_space::local_space>(
               &(macc[0]), __seg_start, N, __self_lidx, &(local_histogram[0]),
-              sycl::access::address_space::local_space, __func);
+               __func, __self_item,
+              __iters_per_work_item);
 
           __reduce_out_histograms(&(local_histogram[0]), &(hacc[0]), num_bins,
                                   __self_item);
@@ -1770,11 +1774,13 @@ Iter2 __histogram_general_private_global_atomics(Policy &&policy,
           __clear_wglocal_histograms(&(hacc_private[__wgroup_idx * num_bins]),
                                      num_bins, __self_item);
 
-          __accum_local_atomics<__work_group_size, __iters_per_work_item,
-                                ::std::uint32_t>(
+          __accum_local_atomics<__work_group_size,
+                                ::std::uint32_t, 
+                                sycl::access::address_space::global_space>(
               &(macc[0]), __seg_start, N, __self_lidx,
               &(hacc_private[__wgroup_idx * num_bins]),
-              sycl::access::address_space::global_space, __func);
+              __func, __self_item,
+              iters_per_work_item);
 
           __reduce_out_histograms(&(hacc_private[__wgroup_idx * num_bins]),
                                   &(hacc[0]), num_bins, __self_item);
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -1467,16 +1467,17 @@ struct __evenly_divided_binhash_impl<T, /* is_floating_point = */ true> {
   T __minimum;
   ::std::uint32_t __num_bins;
   T __scale;
+  T __maximum;
   __evenly_divided_binhash_impl(const T &min, const T &max,
                                 const ::std::uint32_t &num_bins)
-      : __minimum(min), __num_bins(num_bins),
+      : __minimum(min), __maximum(max), __num_bins(num_bins),
         __scale(T(num_bins) / (max - min)) {}
   template <typename T2>::std::uint32_t operator()(T2 &&value) const {
     return ::std::uint32_t((::std::forward<T2>(value) - __minimum) * __scale);
   }
 
   template <typename T2> bool is_valid(const T2 &value) const {
-    return value >= __minimum && value < __minimum + __scale * T(__num_bins);
+    return value >= __minimum && value < __maximum;
   }
 };
 
--- a/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
@@ -1866,8 +1866,8 @@ template <typename Policy, typename Iter1, typename Iter2, typename T,
     dpct::internal::is_iterator<Iter1>::value &&
     dpct::internal::is_iterator<Iter2>::value &&
     internal::is_hetero_execution_policy<::std::decay_t<Policy>>::value>
-HistogramEven(Policy &&policy, Iter1 d_samples, Iter2 d_histogram,
-              int num_levels, T lower_level, T upper_level, Size num_samples) {
+histogram_even(Policy &&policy, Iter1 d_samples, Iter2 d_histogram,
+               int num_levels, T lower_level, T upper_level, Size num_samples) {
   internal::dpl_histogram::histogram(::std::forward<Policy>(policy), d_samples,
                                      d_samples + num_samples, d_histogram,
                                      num_levels - 1, lower_level, upper_level);
@@ -1880,11 +1880,11 @@ template <typename Policy, typename Iter1, typename Iter2, typename T,
     dpct::internal::is_iterator<Iter1>::value &&
     dpct::internal::is_iterator<Iter2>::value &&
     internal::is_hetero_execution_policy<::std::decay_t<Policy>>::value>
-HistogramEven(Policy &&policy, Iter1 d_samples, Iter2 d_histogram,
-              int num_levels, T lower_level, T upper_level,
-              OffsetT num_row_samples, OffsetT num_rows,
-              ::std::size_t row_stride_bytes) {
-  return HistogramEven(
+histogram_even_roi(Policy &&policy, Iter1 d_samples, Iter2 d_histogram,
+                   int num_levels, T lower_level, T upper_level,
+                   OffsetT num_row_samples, OffsetT num_rows,
+                   ::std::size_t row_stride_bytes) {
+  return histogram_even(
       ::std::forward<Policy>(policy),
       oneapi::dpl::permutation_iterator(
           d_samples,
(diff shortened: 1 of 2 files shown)


clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+      __evenly_divided_binhash<T>(first_bin_min_val, last_bin_max_val,
+                                  num_bins));
+}
+
+template <typename Policy, typename Iter1, typename Iter2, typename Iter3>
+inline ::std::enable_if_t<
+    dpct::internal::is_iterator<Iter1>::value &&
+        dpct::internal::is_iterator<Iter2>::value &&
+        dpct::internal::is_iterator<Iter3>::value &&
+        internal::is_hetero_execution_policy<::std::decay_t<Policy>>::value,
```
mmichel11, line 1906: Do we currently plan to support a host policy version? This may be more relevant for when putting into oneDPL.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+  auto N = last - first;
+  if (num_bins < __max_registers) {
+    // If bins fit into registers, use register private accumulation
+    __histogram_general_registers_local_reduction<1024, 32, 16,
+                                                  __histo_value_type>(
+        ::std::forward<Policy>(policy), input_buf.all_view(),
+        bins_buf.all_view(), num_bins, func,
+        std::forward<Range...>(opt_range)...);
+  } else if (num_bins * sizeof(__histo_value_type) < __local_mem_size) {
+    // If bins fit into SLM, use local atomics
```
mmichel11: Have you experimented with subgroup size in the implementation that uses SLM?

If not, this can probably be marked as a future optimization, but we may want to consider experimenting with subgroup size of 16 on Intel devices when large amounts of SLM are being used per work-group as this may achieve higher occupancy.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+inline Iter2
+__histogram_general_select_best(Policy &&policy, Iter1 first, Iter1 last,
+                                Iter2 histogram_first, const Size &num_bins,
+                                IdxHashFunc func, Range &&...opt_range) {
+  using __histo_value_type = typename ::std::iterator_traits<Iter2>::value_type;
+  auto __local_mem_size =
+      policy.queue()
+          .get_device()
+          .template get_info<sycl::info::device::local_mem_size>();
+  constexpr ::std::uint8_t __max_registers = 16;
```
mmichel11, line 1819: Do we know if there are any limitations with this hardcoded value for devices that may not have enough register space or should this be sufficient for most devices?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+          .template get_info<sycl::info::device::global_mem_size>();
+  const ::std::size_t max_segments =
+      ::std::min(__global_mem_size / (num_bins * sizeof(BinType)),
+                 __ceiling_div(N, WorkGroupSize * __min_iters_per_work_item));
+  const ::std::size_t iters_per_work_item =
+      __ceiling_div(N, max_segments * WorkGroupSize);
+  ::std::size_t segments =
+      __ceiling_div(N, WorkGroupSize * iters_per_work_item);
+
+  sycl::buffer<BinType, 1> private_histograms(
```
mmichel11, line 1757: Is this API intended to support an async version? If not, it may be useful to experiment with USM instead of buffers here and see if there is any meaningful performance gain.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+#pragma unroll
+          for (::std::uint8_t k = 0; k < BinsPerWorkItem; k++) {
+            histogram[k] = 0;
+          }
+
+#pragma unroll
+          for (::std::uint8_t idx = 0; idx < ItersPerWorkItem; idx++) {
+            ::std::size_t __val_idx =
+                __seg_start + idx * WorkGroupSize + __self_lidx;
+            if (__val_idx < N) {
```
mmichel11: Similar comment as above with regards to checking if a work group is "full" before this loop.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+inline void __accum_local_atomics(
+    const Iter1 &in_acc, const ::std::size_t &seg_start, const ::std::size_t &N,
+    const ::std::uint32_t &self_lidx, const HistAccessor &wg_local_histogram,
+    const OffsetT &offset, IdxHashFunc func, const sycl::nd_item<1> &self_item,
+    const T &iters_per_work_item) {
+  using __histo_value_type = typename HistAccessor::value_type;
+#pragma unroll
+  for (::std::uint8_t idx = 0; idx < iters_per_work_item; idx++) {
+    ::std::size_t __val_idx = seg_start + idx * WorkGroupSize + self_lidx;
+    if (__val_idx < N) {
```
mmichel11: Is my understanding correct that this conditional is only false in the last work-group when there are not `WorkGroupSize * iters_per_work_item` elements left to process? 

It might be quicker to have one path for a full work group without this check and another for the last work group where we need to make this check in each loop iteration.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,604 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+template <typename HistAccessor, typename OffsetT, typename Size>
+inline void __clear_wglocal_histograms(const HistAccessor &local_histogram,
+                                       const OffsetT &offset,
+                                       const Size &num_bins,
+                                       const sycl::nd_item<1> &self_item) {
+  ::std::uint32_t gSize = self_item.get_local_range()[0];
+  ::std::uint32_t self_lidx = self_item.get_local_id(0);
+  ::std::uint8_t factor = __ceiling_div(num_bins, gSize);
+  ::std::uint8_t k;
+#pragma unroll
```
mmichel11: We have a preprocessor macro in oneDPL that only inserts an unroll pragma if it is supported by the compiler. Is this something useful to port here, or is it a safe assumption that every compiler a user might use supports this?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: There will be a plan to have this in oneDPL, but perhaps not initially.  We do not plan to support a host policy version in this PR.  However, once it is in oneDPL, it could be supported here.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: Yes, this should be an improvement.  I've made it already in my oneDPL branch, and I'll port it back here.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: see above.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: In the context of this PR, we will not support an async version, but eventually in oneDPL this should be added to the async APIs but likely not at the start.  I'll look into this, thanks.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: It will result in slower code but not incorrect code, as it will spill to global memory.  It is currently 32 bytes, as each private bin is `std::uint16_t`.  

Interestingly, I saw decreased performance when using `std::uint8_t` register bins (tested in segment sizes which wouldn't possibly overflow).

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: I have looked at subgroup size some and have found that intel hardware does do better with 16, but did not see huge differences at the time.  It would be good to revisit it now.  However, I also want this PR to compile without warning on all hardware, so I did not want to specify 16 explicitly.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: This is a good question.  It may be good to port that functionality here.

danhoeflinger: I'm not sure if we should be assuming intel LLVM compiler, but I've ported the macros for unrolling from oneDPL.  

@yihanwg - perhaps you can weigh in here if this is necessary.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: I've done this, the performance impact is minimal if anything, but I still think it is a good change.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1455,8 +1455,653 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+template <typename T>
+struct __evenly_divided_binhash_impl<T, /* is_floating_point = */ true> {
+  T __minimum;
+  ::std::uint32_t __num_bins;
+  T __scale;
+  __evenly_divided_binhash_impl(const T &min, const T &max,
+                                const ::std::uint32_t &num_bins)
+      : __minimum(min), __num_bins(num_bins),
+        __scale(T(num_bins) / (max - min)) {}
+  template <typename T2>::std::uint32_t operator()(T2 &&value) const {
```
yihwang-nv:
```suggestion
  template <typename T2> ::std::uint32_t operator()(T2 &&value) const {
```
We can add a whitespace here.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
yihwang-nv: Current Implementation and `#pragram unroll` both looks good to me. SYCLomatic implemented an optimization that add `#pragma unroll` on loops to boost performance (https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/test/dpct/opt_loop_unroll.cu), we can use `#pragma unroll` here.

yihwang-nv (approved):
LGTM, need rebase to resolve conflict.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1449,8 +1449,664 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+struct __evenly_divided_binhash_impl<T, /* is_floating_point= */ false> {
+  T __minimum;
+  ::std::uint32_t __num_bins;
+  T __range_size;
+  __evenly_divided_binhash_impl(const T &min, const T &max,
+                                const ::std::uint32_t &num_bins)
+      : __minimum(min), __num_bins(num_bins), __range_size(max - min) {}
+  template <typename T2>::std::uint32_t operator()(T2 &&value) const {
+    return ::std::uint32_t(
+        ((std::uint64_t(::std::forward<T2>(value)) - __minimum) * std::uint64_t(__num_bins)) /
```
mmichel11: `std::` -> `::std::`

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1449,8 +1449,664 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+  {
+    // Temporary use of stable non-public API from oneDPL,  this function will be
+    // replaced with oneDPL call in an upcoming release.
+    auto keep_input = oneapi::dpl::__ranges::__get_sycl_range<
+        oneapi::dpl::__par_backend_hetero::access_mode::read, Iter1>();
+    auto input_buf = keep_input(first, last);
+
+
+    if (num_bins < __max_registers) {
+      //std::cout<<"registers"<<std::endl;
```
mmichel11: Should these commented couts be removed?

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1449,8 +1449,664 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+
+      // If bins fit into registers, use register private accumulation
+      __histogram_general_registers_local_reduction<1024, 32, 16,
```
mmichel11: We might want to check if `1024` is a supported work group size to ensure this can run on all supported devices.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1449,8 +1449,664 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+          .template get_info<sycl::info::device::local_mem_size>();
+  constexpr ::std::uint8_t __max_registers = 16;
+
+  // Temporary use of stable non-public API from oneDPL,  this function will be
+  // replaced with oneDPL call in an upcoming release.
+  auto keep_bins = oneapi::dpl::__ranges::__get_sycl_range<
+      oneapi::dpl::__par_backend_hetero::access_mode::write, Iter2>();
+  auto bins_buf = keep_bins(histogram_first, histogram_first + num_bins);
+
+  oneapi::dpl::fill(std::forward<Policy>(policy), bins_buf.all_view().begin(),
```
mmichel11: We forward the policy here and then later on the histogram call. I think it is only safe to forward the policy in the last call.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
```diff
@@ -1449,8 +1449,664 @@ inline void segmented_sort_pairs_by_two_pair_sorts(
...
+      if (N <= 524288) {
+        __histogram_general_local_atomics<1024, 4, __histo_value_type>(
+            ::std::forward<Policy>(policy), input_buf.all_view(),
+            bins_buf.all_view(), num_bins, func,
+            std::forward<Range...>(opt_range)...);
+      } else {
+        __histogram_general_local_atomics<1024, 32, __histo_value_type>(
+            ::std::forward<Policy>(policy), input_buf.all_view(),
+            bins_buf.all_view(), num_bins, func,
+            std::forward<Range...>(opt_range)...);
```
mmichel11: `std::` -> `::std::` here and in the surrounding conditionals.

mmichel11 (changes requested):
A few more comments from me. Once these are considered I am ready to approve.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: done, thanks

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: done, thanks

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: yes, done.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: I've converted wg_size to a runtime parameter, and am checking against the maximum supported.  I don't think we were getting much performance wise from this being compile time anyway. I'll confirm this in the morning.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: good catch, fixed.

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: fixed.

yihwang-nv (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/algorithm.h
danhoeflinger: Changing to a runtime parameter seems to have a minor negative impact on performance but its necessary to do this for correctness.

mmichel11 (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1221, opened 2023-08-25
[SYCLomatic] Make the binary not rely on the order of waring message list in file Diagnostics.inc

intwanghao:


zhimingwang36 (approved):
LGTM

clang/lib/DPCT/Diagnostics.h
```diff
@@ -92,7 +117,7 @@ enum class Warnings {
 #undef DEF_ERROR
 #undef DEF_WARNING
 #undef DEF_COMMENT
-  END
+  END = MaxWarningID + 1
```
ziranzha: How about remove `BEGIN` and `END`, use static variable in class `DiagnosticsMessage` for the same functionality.

ziranzha (approved):

PR #1247, opened 2023-08-29
[SYCLomatic] Support coloring the output of query-api-mapping.

tangjj11:

### DIFF
--- a/clang/lib/DPCT/DPCT.cpp
+++ b/clang/lib/DPCT/DPCT.cpp
@@ -755,7 +755,8 @@ int runDPCT(int argc, const char **argv) {
     SourcePathList = OptParser->getSourcePathList();
   }
   RefactoringTool Tool(OptParser->getCompilations(), SourcePathList);
-  std::string QueryAPIMappingStr;
+  std::string QueryAPIMappingSrc;
+  std::string QueryAPIMappingOpt;
   if (DpctGlobalInfo::isQueryAPIMapping()) {
     APIMapping::initEntryMap();
     auto SourceCode = APIMapping::getAPISourceCode(QueryAPIMapping);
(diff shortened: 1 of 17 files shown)


clang/lib/DPCT/DPCT.cpp
```diff
@@ -808,7 +808,9 @@ int runDPCT(int argc, const char **argv) {
     }
     StartPos = StartPos + StartStr.length();
     EndPos = SourceCode.find_last_of('\n', EndPos);
-    QueryAPIMappingStr += SourceCode.substr(StartPos, EndPos - StartPos + 1);
+    QueryAPIMappingStr +=
+        "\033[32m" + SourceCode.substr(StartPos, EndPos - StartPos + 1).str() +
```
zhimingwang36: suggest to define a Macor/constExpr to replace the "033[32m" with a right name to improve readability.

clang/lib/DPCT/DPCT.cpp
```diff
@@ -1130,6 +1132,7 @@ int runDPCT(int argc, const char **argv) {
     static const std::string StartStr{"// Start"};
     static const std::string EndStr{"// End"};
     bool Flag = false;
+    llvm::outs() << llvm::raw_ostream::BLUE;
```
zhimingwang36, line 1135: if the output color is controlled by here, then don't try to embed raw string with color control

tangjj11: fixed

clang/lib/DPCT/DPCT.cpp
tangjj11: fixed the next comment and this problem is not exist.

wenhuiNi (approved):

yihwang-nv (approved):

zhimingwang36 (approved):
LGTM

PR #1259, opened 2023-08-31
[SYCLomatic] Emit warning message for dpct_placeholder in kernel name

intwanghao:


clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2785,30 +2792,46 @@ void CallFunctionExpr::emplaceReplacement() {
                                          getExtraArguments(), nullptr));
 }
 
-std::string CallFunctionExpr::getTemplateArguments(bool WrittenArgsOnly,
+std::string CallFunctionExpr::getTemplateArguments(KernelPrinter &Printer,
```
ziranzha: How about using std::string instead of KernelPrinter to avoid confusing.

ziranzha (approved):

zhimingwang36 (approved):
LGTM

PR #1273, opened 2023-09-05
Switch clang-format to new file to prevent conflict from pulldown

DoyleLi:


tangjj11 (approved):

DoyleLi:
CI failures are not relate with the patch. WIll merge it now.

PR #1287, opened 2023-09-08
[SYCLomatic] Add workaround for templated kernel call in macro

AndyCHHuang:


zhiweij1 (approved):

clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2825,6 +2825,7 @@ std::string CallFunctionExpr::getTemplateArguments(bool &IsNeedWarning,
     if ((TA.isNull() || !TA.isWritten()) && WrittenArgsOnly)
       continue;
     std::string Str = TA.getString();
+
```
zhimingwang36: there is no logic change here, to keep small footprint for this PR, so please revert this change.

zhimingwang36 (approved):
LGTM

PR #1295, opened 2023-09-12
[SYCLomatic] Use FD->getDeclName().getAsString() to avoid assert checking in FD->getName()

zhiweij1:

### DIFF
--- a/clang/lib/DPCT/ASTTraversal.cpp
+++ b/clang/lib/DPCT/ASTTraversal.cpp
@@ -8673,7 +8673,8 @@ void DeviceFunctionDeclRule::runRule(
     // For Xe-LP architecture, if the sub-group size is 32, then each work-item
     // can use 128 * 32 Byte / 32 = 128 Byte registers.
     if(LocalVariableSize > 128) {
-      report(SM.getExpansionLoc(FD->getBeginLoc()), Warnings::REGISTER_USAGE, false, FD->getName());
+      report(SM.getExpansionLoc(FD->getBeginLoc()), Warnings::REGISTER_USAGE,
+             false, FD->getDeclName().getAsString());
     }
   }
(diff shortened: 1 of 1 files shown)


tomflinda (approved):
LGTM

ziranzha (approved):

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -8673,7 +8673,8 @@ void DeviceFunctionDeclRule::runRule(
     // For Xe-LP architecture, if the sub-group size is 32, then each work-item
     // can use 128 * 32 Byte / 32 = 128 Byte registers.
     if(LocalVariableSize > 128) {
-      report(SM.getExpansionLoc(FD->getBeginLoc()), Warnings::REGISTER_USAGE, false, FD->getName());
+      report(SM.getExpansionLoc(FD->getBeginLoc()), Warnings::REGISTER_USAGE,
+             false, FD->getDeclName().getAsString());
```
zhimingwang36, line 8677: why not call FD->getNameAsString() directly?

zhiweij1: getNameAsString is depracated.
```
  /// Get a human-readable name for the declaration, even if it is one of the
  /// special kinds of names (C++ constructor, Objective-C selector, etc).
  ///
  /// Creating this name requires expensive string manipulation, so it should
  /// be called only when performance doesn't matter. For simple declarations,
  /// getNameAsCString() should suffice.
  //
  // FIXME: This function should be renamed to indicate that it is not just an
  // alternate form of getName(), and clients should move as appropriate.
  //
  // FIXME: Deprecated, move clients to getName().
  std::string getNameAsString() const { return Name.getAsString(); }
```

PR #1335, opened 2023-09-27
[SYCLomatic] Resolve compilations errors and correctness issues in dpct::device_new and dpct::device_delete

mmichel11:
## Summary
This PR resolves identified compilation and correctness issues within `dpct::device_new` and `dpct::device_delete`. Furthermore, `dpct::device_free` is corrected to properly work in USM mode. A corresponding test PR is provided in [SYCLomatic-test #477](https://github.com/oneapi-src/SYCLomatic-test/pull/477)

## Details
The following bugs have been identified and resolved:

- A correct implementation for USM enabled mode is provided for `dpct::device_new`.
- The placement new `dpct::device_new` implementation now accepts an allocation passed as `dpct::device_pointer<void>`.
- For non-USM mode, `dpct::device_new` is made a friend function of `dpct::device_pointer` in order to access its protected members.
- `dpct::device_new` now constructs objects in allocated memory.
- `dpct::device_new` allocates the correct number of elements requested.
- `dpct::device_delete` now frees device memory in USM mode.
- `dpct::device_free` now frees device memory in USM mode.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h
@@ -269,52 +269,6 @@ public:
   } // required
 };
 
-template <typename T, sycl::access_mode Mode, typename Allocator>
-class device_pointer
-    : public device_pointer_base<T, Allocator,
-                                 device_pointer<T, Mode, Allocator>> {
-private:
-  using base_type = device_pointer_base<T, Allocator, device_pointer>;
-
-public:
-  using value_type = T;
-  using difference_type = std::make_signed<std::size_t>::type;
-  using pointer = T *;
-  using reference = T &;
-  using iterator_category = std::random_access_iterator_tag;
-  using is_hetero = std::true_type; // required
-  using is_passed_directly = std::false_type;
-  static constexpr sycl::access_mode mode = Mode; // required
-
-  device_pointer(sycl::buffer<T, 1> in, std::size_t i = 0) : base_type(in, i) {}
-#ifdef __USE_DPCT
-  template <typename OtherT> device_pointer(OtherT *ptr) : base_type(ptr) {}
-#endif
-  // needed for malloc_device, count is number of bytes to allocate
-  device_pointer(const std::size_t count) : base_type(count) {}
-  device_pointer() : base_type() {}
-  device_pointer(const device_pointer &in) : base_type(in) {}
-  device_pointer &operator+=(difference_type forward) {
-    this->idx += forward;
-    return *this;
-  }
-  device_pointer &operator-=(difference_type backward) {
-    this->idx -= backward;
-    return *this;
-  }
-  // include operators from base class
-  using base_type::operator++;
-  using base_type::operator--;
-  device_pointer &operator++() {
-    this->idx += 1;
-    return *this;
-  }
-  device_pointer &operator--() {
-    this->idx -= 1;
-    return *this;
-  }
-};
-
 template <sycl::access_mode Mode, typename Allocator>
 class device_pointer<void, Mode, Allocator>
     : public device_pointer_base<dpct::byte_t, Allocator,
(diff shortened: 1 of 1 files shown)


tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -917,34 +928,56 @@ device_pointer<T> malloc_device(const std::size_t num_elements) {
...
+                             const std::size_t count) {
+  auto converted_buf = p.buffer.template reinterpret<T>(sycl::range<1>(count));
+  ::std::uninitialized_fill(
+      oneapi::dpl::execution::make_device_policy(dpct::get_default_queue()),
+      oneapi::dpl::begin(converted_buf), oneapi::dpl::end(converted_buf),
+      value);
+  return device_pointer<T>(converted_buf, p.idx);
+}
+// buffer manages lifetime
+template <typename T> void free_device(device_pointer<T> ptr) {}
```
ziranzha, line 951: Should put in namespace "internal"

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -917,34 +928,56 @@ device_pointer<T> malloc_device(const std::size_t num_elements) {
...
-  return p + count;
+  dpct::device_pointer<T> converted_p(static_cast<T *>(p.get()));
+  ::std::uninitialized_fill(
+      oneapi::dpl::execution::make_device_policy(dpct::get_default_queue()),
+      converted_p, converted_p + count, value);
+  return converted_p;
 }
-template <typename T>
-device_pointer<T> device_new(device_pointer<T> p, const std::size_t count = 1) {
+template <typename T> void free_device(device_pointer<T> ptr) {
```
ziranzha, line 962: Same as above.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -917,34 +928,56 @@ device_pointer<T> malloc_device(const std::size_t num_elements) {
...
-  return device_pointer<T>(count);
+  return device_new(device_pointer<T>(sizeof(T) * count), T{}, count);
 }
 
-template <typename T> void free_device(device_pointer<T> ptr) {}
-
 template <typename T>
 typename std::enable_if<!std::is_trivially_destructible<T>::value, void>::type
 device_delete(device_pointer<T> p, const std::size_t count = 1) {
   for (std::size_t i = 0; i < count; ++i) {
```
ziranzha: Can we use onedpl::destroy to accelerate the destruction?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -917,34 +928,56 @@ device_pointer<T> malloc_device(const std::size_t num_elements) {
...
   return device_pointer<void>(num_bytes);
 }
-template <typename T>
-device_pointer<T> device_new(device_pointer<T> p, const T &value,
+
+#ifdef DPCT_USM_LEVEL_NONE
+template <typename T, typename U>
+device_pointer<T> device_new(device_pointer<U> p, const T &value,
+                             const std::size_t count) {
+  auto converted_buf = p.buffer.template reinterpret<T>(sycl::range<1>(count));
```
danhoeflinger: Is this more permissive than we would like to be?  
I think this is designed mostly to accommodate a `void` pointer type `U` and an arbitrary type `T`. 

Do we gain anything useful by opening it up to an arbitrary type `U` rather than just using `void`?

I wonder if this could be lead to easy to miss bugs for a user.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
danhoeflinger, line 951: I know this is just moving existing code, not new code, but as we are improving this I wonder if we want some sort of warning to the user here that for USM_NONE mode this does not actually control the "freeing".  I think the code is the right approach, but I wonder if this could get users into trouble...

@tomflinda do you have an opinion here, and what would the best way be to make such a warning?

clang/runtime/dpct-rt/include/dpl_extras/memory.h
```diff
@@ -917,34 +928,56 @@ device_pointer<T> malloc_device(const std::size_t num_elements) {
...
+// buffer manages lifetime
+template <typename T> void free_device(device_pointer<T> ptr) {}
+#else
+template <typename T, typename U>
+device_pointer<T> device_new(device_pointer<U> p, const T &value,
                              const std::size_t count = 1) {
-  std::vector<T> result(count, value);
-  p.buffer = sycl::buffer<T, 1>(result.begin(), result.end());
-  return p + count;
+  dpct::device_pointer<T> converted_p(static_cast<T *>(p.get()));
```
danhoeflinger: Same comment here about `void` vs `typename U`

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: `free_device` is actually a public function that is used to free `device_malloc` calls and is not an internal utility.

mmichel11: > `free_device` is a public function that is used to free `device_malloc` calls and is not an internal utility.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Yes, I have added this change.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Yes, I have changed this function to only have a single template parameter `T` and to have an overload that accepts a `device_pointer<void>` type.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: Changed.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
mmichel11: See comment above.

mmichel11:
After some offline discussion with @danhoeflinger with regards to an ambiguous function call when `T = size_t` and similar types in `device_new`, I have switched `device_new` to only accept a `device_pointer<void>` type which safely solves the problem. To do this, a conversion operator from `device_pointer<T>` to `device_pointer<void>` was provided. Some code was reordered to enable this change (`device_pointer<void>` specialization was moved above the general device_pointer<T> case).

danhoeflinger (approved):
LGTM.
I'm interested about the answer on the user warning raised in this review, but I don't think its required for this PR.

clang/runtime/dpct-rt/include/dpl_extras/memory.h
tomflinda: I'm OK to current solution.
BTW, I found a new issue for device_pointer, will file a new ticket to you.

tomflinda: Why name free_device instead of device_free?  The name stype is different from device device_new, device_delete.

danhoeflinger: I believe this is just modifying code which has existed here for some time, not new APIs (so the naming decision was not made in this PR).  However, if `free_device` is not already in use by SYCLomatic, I would also be in favor of switching it to `device_free` as well.

tomflinda: free_device has been used in the migrated code https://github.com/oneapi-src/SYCLomatic/blob/1b1335850c4b7fcde30659dcfef08bffbd340e94/clang/test/dpct/thrust-cast.cu#L37

So, I'm OK to current solution.

tomflinda (approved):
LGTM

ziranzha (approved):

PR #1358, opened 2023-10-16
[SYCLomatic] Fix 3 internal error bugs

zhiweij1:
1. Use getAssistNodeAsType when trying to get node which has been accessed
2. Check the return pointer
3. Check the argument number. When template is not instantiated, the member call is not recognized (treated as CallExpr without linking the correct declaration).


tomflinda (approved):
LGTM

ziranzha (approved):

PR #1363, opened 2023-10-18
[SYCLomatic] Fix two bug issues related to auto-generated build script

tomflinda:
1. Fix issue of missing adding including path with "-I" option when working directory of including path is in-root directory
2. Fix issue that target binary is generated outside of out-root directory


zhiweij1 (approved):

zhimingwang36 (approved):
LGTM

PR #1365, opened 2023-10-18
[SYCLomatic] Diagnose if the argument of __syncwarp is not 0xffffffff

yihwang-nv:

### DIFF
--- a/clang/test/dpct/sync_api_noneusm.cu
+++ b/clang/test/dpct/sync_api_noneusm.cu
@@ -69,6 +69,13 @@ __global__ void k() {
   // CHECK: sycl::group_barrier(item_ct1.get_sub_group());
   __syncwarp(0xffffffff);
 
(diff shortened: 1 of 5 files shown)


clang/test/dpct/sync_api_noneusm.cu
```diff
@@ -69,6 +69,13 @@ __global__ void k() {
   // CHECK: sycl::group_barrier(item_ct1.get_sub_group());
   __syncwarp(0xffffffff);
 
+  // CHECK: DPCT1119:{{.*}}: The argument of __syncwarp is assumed to be 0xffffffff. You may need to adjust the code.
```
zhimingwang36, line 72:
```suggestion
  // CHECK: DPCT1119:{{.*}}: The __syncwarp(1) which cause the executing thread to wait subset of threads in sub group, is migrated to sycl::group_barrier() which cause the executing thread to wait all threads in sub group. You may need to adjust the code.
```

PR #1368, opened 2023-10-20
[SYCLomatic] Fix the issue that emit unnecessary waring for shfl function with mask 0xFFFFFFFF

intwanghao:


zhimingwang36 (approved):
LGTM

PR #1372, opened 2023-10-26
[SYCLomatic] Allow non-standard syntax for template method out-of-class definition for CUDA language and fix it in the migrated code

zhiweij1:


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -14613,3 +14613,26 @@ void TypeRemoveRule::runRule(
 }
 
 REGISTER_RULE(TypeRemoveRule, PassKind::PK_Analysis)
+
+void CompatNVCCRule::registerMatcher(ast_matchers::MatchFinder &MF) {
+  MF.addMatcher(functionDecl().bind("FunctionDecl"), this);
```
ShengchenJ: How about use the functionTemplateDecl as a replace of functionDecl?

zhiweij1: I have updated the matcher.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -14613,3 +14613,28 @@ void TypeRemoveRule::runRule(
 }
 
 REGISTER_RULE(TypeRemoveRule, PassKind::PK_Analysis)
+
+void CompatNVCCRule::registerMatcher(ast_matchers::MatchFinder &MF) {
```
zhimingwang36: how about change the rule to: CompatWithClang,  add some comment to detail on diff b/w EDG front end and Clang.

clang/lib/Sema/SemaDecl.cpp
```diff
@@ -10540,8 +10540,20 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
         HasExplicitTemplateArgs = false;
       } else if (FunctionTemplate) {
         // Function template with explicit template arguments.
+#ifdef SYCLomatic_CUSTOMIZATION
+
+        if (getLangOpts().CUDA) {
```
zhimingwang36, line 10544: Please check if need add this condition.

zhiweij1: `nvcc -x c++` cannot compile this code also. So it is reasonable to add this condition.

clang/lib/DPCT/ASTTraversal.cpp
zhiweij1: Done

tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

Issue #497, opened 2023-01-12
error: unknown type name '__grid_constant__'


zjin-lcf:
Please see https://github.com/intel/llvm/issues/7998

tomflinda:
@zjin-lcf  currently there is no mapping concept in SYCL side, we are considering to remove this annotation since it does not impact the result.

PR #1396, opened 2023-11-02
[SYCLomatic] Cleanup dpct::tagged_pointer and dpct::release_temporary_allocation

mmichel11:
This PR cleans up the function signatures of dpct::tagged_pointer and dpct::release_temporary_allocation to enable easier migration to these APIs by the SYCLomatic team.

Please note that these changes will cause SYCLomatic tests to fail until the corresponding test PR changes ([SYCLomatic-test #497](https://github.com/oneapi-src/SYCLomatic-test/pull/497)) are merged.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpl_extras/memory.h
+++ b/clang/runtime/dpct-rt/include/dpl_extras/memory.h
@@ -701,12 +701,13 @@ struct device_sys_tag : public sys_tag {};
 struct host_sys_tag : public sys_tag {};
 
 #ifdef DPCT_USM_LEVEL_NONE
-template <typename Tag, typename T = void> class tagged_pointer {
+template <typename T, typename Tag> class tagged_pointer {
   static_assert(false,
                 "tagged_pointer is not supported with DPCT_USM_LEVEL_NONE");
 };
 template <typename PolicyOrTag, typename Pointer>
-void release_temporary_allocation(PolicyOrTag &&policy_or_tag, Pointer ptr) {
+void release_temporary_allocation(PolicyOrTag &&policy_or_tag, Pointer ptr,
+                                  ::std::ptrdiff_t) {
   static_assert(
       false,
       "release_temporary_allocation is not supported with DPCT_USM_LEVEL_NONE");
@@ -774,7 +775,7 @@ inline constexpr bool is_host_policy_or_tag_v =
 // tagged_pointer provides a wrapper around a raw pointer type with a tag of the
 // location of the allocated memory. Standard pointer operations are supported
 // with this class.
-template <typename Tag, typename T = void> class tagged_pointer {
+template <typename T, typename Tag> class tagged_pointer {
 public:
   using value_type = T;
   using difference_type = ::std::ptrdiff_t;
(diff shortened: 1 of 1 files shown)


danhoeflinger (approved):
LGTM

danhoeflinger:
> LGTM

Thinking a little more, I don't like the extra (ignored) parameter for `release_temporary_allocation`.  It seems to me that this shouldn't be there if we aren't doing something with it.  

Should this third param just be discarded by SYCLomatic in migration instead if we are just ignoring it?

danhoeflinger (changes requested):
> LGTM

Thinking a little more, I don't like the extra (ignored) parameter for `release_temporary_allocation`.  It seems to me that this shouldn't be there if we aren't doing something with it.  

Should this third param just be discarded by SYCLomatic in migration instead if we are just ignoring it?

mmichel11:
> > LGTM
> 
> Thinking a little more, I don't like the extra (ignored) parameter for `release_temporary_allocation`. It seems to me that this shouldn't be there if we aren't doing something with it.
> 
> Should this third param just be discarded by SYCLomatic in migration instead if we are just ignoring it?

@tomflinda Do you have any objection to this? It would mean the `n` parameter (number of elements to release) would just be ignored in mapping as it is unnecessary with `sycl::free`.

danhoeflinger:
> @tomflinda Do you have any objection to this? It would mean the `n` parameter (number of elements to release) would just be ignored in mapping as it is unnecessary with `sycl::free`.

To provide a bit more justification for my stance.
If we add and ignore this third parameter:
1) It adds an extra (false) requirement to the user to keep track of the size of allocations
2) It makes it seem like the user has extra functionality which is not there.  We do not have the ability to free only part of what was allocated, we also do not check to make sure that the number provided is the same as what was originally allocated and throw or provide any error.

tomflinda:
> we are



> > > LGTM
> > 
> > 
> > Thinking a little more, I don't like the extra (ignored) parameter for `release_temporary_allocation`. It seems to me that this shouldn't be there if we aren't doing something with it.
> > Should this third param just be discarded by SYCLomatic in migration instead if we are just ignoring it?
> 
> @tomflinda Do you have any objection to this? It would mean the `n` parameter (number of elements to release) would just be ignored in mapping as it is unnecessary with `sycl::free`.
No, I haven't, you can discard the third param, I'm OK the change.

tomflinda:
> > @tomflinda Do you have any objection to this? It would mean the `n` parameter (number of elements to release) would just be ignored in mapping as it is unnecessary with `sycl::free`.
> 
> To provide a bit more justification for my stance. If we add and ignore this third parameter:
> 
> 1. It adds an extra (false) requirement to the user to keep track of the size of allocations
> 2. It makes it seem like the user has extra functionality which is not there.  We do not have the ability to free only part of what was allocated, we also do not check to make sure that the number provided is the same as what was originally allocated and throw or provide any error.

OK to me to ignore the third parameter.

mmichel11:
@danhoeflinger @tomflinda I have removed the third parameter in `dpct::release_temporary_allocation`. The only change now is reordering the templated parameters for `dpct::tagged_pointer`.

danhoeflinger (approved):
LGTM

tomflinda (approved):
LGTM

danhoeflinger:
@zhimingwang36 @tomflinda I think this is ready to merge.

ziranzha (approved):

PR #1412, opened 2023-11-09
[SYCLomatic]Support the usage of ccl enum being implicitly converted

wenhuiNi:


AndyCHHuang (approved):
LGTM
Normally, if ncclInt32 appears in callexpr, the parameter type is ncclDataType_t and thus no need for casting.

ziranzha (approved):

PR #1425, opened 2023-11-14
[SYCLomatic] Migration support for CUDA driver error APIs and cudaErrorr_enum

TejaX-Alaghari:
[Feature]: Added support for migration of - 
- CUDA Error Driver APIs - cuGetErrorName, cuGetErrorString
- CUDA Error Enum Type - cudaError_enum) and
- CUDA Error Enum Value - cudaErrorInvalidConfiguration

Signed-off-by: TejaX, Alaghari [teja.alaghari@intel.com](mailto:teja.alaghari@intel.com)


clang/lib/DPCT/APINames.inc
```diff
@@ -84,6 +84,10 @@ ENTRY(cudaGetErrorString, cudaGetErrorString, true, NO_FLAG, P0, "Successful")
 ENTRY(cudaGetLastError, cudaGetLastError, true, NO_FLAG, P0, "DPCT1010")
 ENTRY(cudaPeekAtLastError, cudaPeekAtLastError, true, NO_FLAG, P0, "DPCT1026/DPCT1027")
 
+// error handling functions of driver API
+ENTRY(cuGetErrorName, cuGetErrorName, true, NO_FLAG, P0, "Successful")
```
wenhuiNi, line 88: I think there are  already cuGetErrorName and cuGetErrorString in https://github.com/oneapi-src/SYCLomatic/blob/f2c0cab249c63dd0d729dd0a08e47bd3021ea02e/clang/lib/DPCT/APINames.inc#L1567

wenhuiNi (changes requested):

clang/lib/DPCT/APINames.inc
TejaX-Alaghari: Thanks for pointing that out @wenhuiNi.

PR #1459, opened 2023-11-29
[SYCLomatic][part3]Refine cmake migration framework to support full word match and global build-in rule file

tomflinda:
1 Enable full word match for cmake syntax migration rule
2 Skip user defined migration run for some cmake syntax if implicit migration rule is available
3 Use a global build-in built-in migration rule file for all the lit tests
4 Split a separate analysis parse to collect variable and its value before applying migration rules


AndyCHHuang (approved):
LGTM

clang/lib/DPCT/MigrateCmakeScript.cpp
```diff
@@ -501,18 +461,51 @@ int skipCmakeComments(const std::string &Input, int Index) {
...
+  Value = Input.substr(Begin, End - Begin);
+
+  std::string ReplStr = processArgOfCmakeVersionRequired(Value, CmakeVarMap);
+
+  Input.replace(Begin, End - Begin, ReplStr);
+  Size = Input.size();            // Update string size
+  Index = Begin + ReplStr.size(); // Update index
+}
+
+void applyImplicitMigrationRules(std::string &Input) {
```
zhimingwang36, line 502: Please add comment on: implict migration Rules is used in what kind of scenarsio.

clang/lib/DPCT/MigrateCmakeScript.cpp
```diff
@@ -604,3 +566,93 @@ std::string convertCmakeCommandsToLower(const std::string &InputString) {
...
+}
+
+static void applyCmakeMigrationRules() {
+  for (auto &Entry : CmakeScriptFileBufferMap) {
+    auto &Buffer = Entry.second;
+
+    // Apply implicit migration rules
+    applyImplicitMigrationRules(Buffer);
+
+    // Apply user define migration rules
```
zhimingwang36, line 603: please comment on:  when yaml based rule is used,  or used for what kind cases.

tomflinda: ImplicitMigrationRules is used when the migration logic is difficult to be described with yaml based rule syntax. 
Currently, only migration of cmake_minimum_required() is applied with ImplicitMigrationRule. Once ImplicitMigrationRule for some cmake syntax is implemented, the yaml based rules for the same cmake syntax will be skipped and a msg is emitted to notify the user.

clang/lib/DPCT/MigrateCmakeScript.cpp
tomflinda: ImplicitMigrationRules is used when the migration logic is difficult to be described with yaml based rule syntax.
Currently, only migration of cmake_minimum_required() is applied with ImplicitMigrationRule. Once ImplicitMigrationRule for some cmake syntax is implemented, the yaml based rules for the same cmake syntax will be skipped and a msg is emitted to notify the user.

zhimingwang36 (approved):
LGTM

PR #1480, opened 2023-12-04
[SYCLomatic][part4] Refine cmake migration framework to make Priority field be consistent with that of other user defined rules

tomflinda:
1. Add new field "CmakeSyntax"  to identify cmake syntax need to be migrated.
2. Add new kind value "CMakeRule" to identify user defined yaml rules.


clang/lib/DPCT/MigrateCmakeScript.cpp
```diff
@@ -656,11 +667,42 @@ static void storeBufferToFile() {
...
+void registerCmakeMigrationRule(MetaRuleObject &R) {
+  auto PR =
+      MetaRuleObject::PatternRewriter(R.In, R.Out, R.Subrules, R.MatchMode,
+                                      R.RuleId, R.CmakeSyntax, R.Priority);
+
+  auto Iter = CmakeBuildInRules.find(PR.CmakeSyntax);
+  if (Iter != CmakeBuildInRules.end()) {
+    if (PR.Priority == RulePriority::Takeover) {
+      assert(Iter->second.Priority < PR.Priority &&
+             "Two same cmake synaxes have \'Takeover\' priority.\n ");
```
zhiweij1:
```suggestion
             "Two same cmake syntaxes have \'Takeover\' priority.\n ");
```

clang/lib/DPCT/MigrateCmakeScript.cpp
```diff
@@ -594,26 +602,29 @@ static void unifyInputFileFormat() {
 }
 
 static void applyCmakeMigrationRules() {
+
+  std::map<std::string, void (*)(std::string &, size_t &, size_t &)>
```
zhiweij1:
```suggestion
  static const std::map<std::string, void (*)(std::string &, size_t &, size_t &)>
```

tomflinda: Okay, will update it.

clang/lib/DPCT/MigrateCmakeScript.cpp
tomflinda: Will fix it in the updated PR>

AndyCHHuang (approved):
LGTM

zhiweij1 (approved):

ziranzha (approved):

clang/lib/DPCT/MigrateCmakeScript.cpp
```diff
@@ -594,26 +602,30 @@ static void unifyInputFileFormat() {
...
-      // emitted to notify the user. Currently, only migration of
-      // cmake_minimum_required() is implemented by implicit migration rule.
-      if (PR.RuleId.find("cmake_minimum_required") != std::string::npos) {
-        llvm::outs() << "Migration for cmake_minimum_required has been "
-                        "implemented as implicit rule, so migration rule \'"
-                     << PR.RuleId << "\' is skipped.\n ";
-        continue;
+    for (const auto &CmakeSyntaxEntry : CmakeBuildInRules) {
+      const auto &PR = CmakeSyntaxEntry.second;
+      if (PR.In.empty() && PR.Out.empty()) {
```
zhimingwang36, line 618: please refine L618,  don't assume in/out is always not empty.

PR #1481, opened 2023-12-05
[SYCLomatic] Implement migration for bfe/bfi

yihwang-nv:


clang/runtime/dpct-rt/include/math.hpp
```diff
@@ -242,6 +242,27 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
       .as<sycl::vec<unsigned, 1>>();
 }
 
+template <typename T>
+inline T bfe(T source, uint32_t bit_start, uint32_t num_bits) {
+  if (!num_bits)
+    return 0;
+  bit_start &= 0xFF;
+  num_bits &= 0xFF;
```
danhoeflinger: I'm not sure I understand the reason to bitwise AND with the lowest 8 bits.  
Of course, if someone sends in a value greater than 255, it is somewhat nonsensical, but if we want to put in guardrails, wouldn't a branch with an error be better than masking these values?

clang/runtime/dpct-rt/include/math.hpp
```diff
@@ -242,6 +242,27 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
       .as<sycl::vec<unsigned, 1>>();
 }
 
+template <typename T>
+inline T bfe(T source, uint32_t bit_start, uint32_t num_bits) {
+  if (!num_bits)
+    return 0;
```
danhoeflinger: Is this special casing worth the branch?  
If num_bits is 0, the mask will be `(1 << 0) - 1 = 0`, and will return 0.  
I would expect the branch compared to the regular function path is not much faster, and this will slow down all the `num_bits != 0` cases.

clang/runtime/dpct-rt/include/math.hpp
```diff
@@ -242,6 +242,27 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+  num_bits &= 0xFF;
+  const T mask = (T{1} << num_bits) - 1;
+  return (source >> bit_start) & mask;
+}
+
+template <typename T>
+inline T bfi(T x, T y, uint32_t bit_start, uint32_t num_bits) {
+  constexpr unsigned msb = std::numeric_limits<T>::digits;
+  bit_start &= 0xFF;
+  num_bits &= 0xFF;
```
danhoeflinger: same comment as above... why mask here?

clang/runtime/dpct-rt/include/math.hpp
```diff
@@ -242,6 +242,27 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+  if (bit_start > msb || num_bits == 0)
+    return y;
+  T mask = (~(T{}) >> (msb - num_bits)) << bit_start;
```
danhoeflinger: Without knowing the input type, `T`, we are making assumptions about the default constructor.  Probably better to use `T{0}`?  

I also think it may be good to return the `::std::enable_if_t<::std::is_unsigned_v<T>, T>`  to both functions to ensure the incoming type makes sense.  If callers of this function want to reinterpret cast their input data, they can before calling.  If we did this, the assumptions about the default constructor are OK (but maybe more clear to be explicit anyway).

clang/runtime/dpct-rt/include/math.hpp
yihanwg: I try to use dpct::bfe as a mapping of cub::BFE and PTX bfe instruction. To keep the same semantic with the wording in https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#integer-arithmetic-instructions-bfe ,I masked `bit_start` and `num_bits`.

clang/runtime/dpct-rt/include/math.hpp
yihwang-nv: > Without knowing the input type, T, we are making assumptions about the default constructor. Probably better to use T{0}?

Agree.
> I also think it may be good to return the ::std::enable_if_t<::std::is_unsigned_v<T>, T> to both functions to ensure the incoming type makes sense. If callers of this function want to reinterpret cast their input data, they can before calling. If we did this, the assumptions about the default constructor are OK (but maybe more clear to be explicit anyway).

I relax the constraints, so it now allows T be a signed/unsigned integer type

clang/runtime/dpct-rt/include/math.hpp
yihwang-nv: Agree, I'll remove this branch

yihwang-nv:
@danhoeflinger Thanks for your review!

clang/runtime/dpct-rt/include/math.hpp
danhoeflinger: I would be inclined to remove it.  Any pre-migration code relying on this masking is almost surely relying on a bug.  I can't imagine someone sending in a bit start of 258 and expecting a bit start of 2 to be what is applied.

Additionally, this adds a couple extra (hopefully no-op) operations to every call.

clang/runtime/dpct-rt/include/math.hpp
danhoeflinger: OK, do we want to use `std::is_integral_v` then?
https://en.cppreference.com/w/cpp/types/is_integral

clang/runtime/dpct-rt/include/math.hpp
yihanwg: I updated this function so that it behaves the same as PTX in edge cases(Eg. bit_start/num_bits out of range, or `T` is signed integer). As far as I know, `std::min` can be optimized https://godbolt.org/z/rbjT7Wb9r , but I'm not sure yet if it will be optimized on GPU.

clang/runtime/dpct-rt/include/math.hpp
yihwang-nv: removed

clang/runtime/dpct-rt/include/math.hpp
yihwang-nv: done

clang/runtime/dpct-rt/include/math.hpp
yihwang-nv: I updated this function to keep behavior stable for boundary cases(Eg. bit_start/num_bits out of range, or `T` is signed integer). As far as I know, `std::min` can be optimized https://godbolt.org/z/rbjT7Wb9r , but I'm not sure yet if it will be optimized on GPU.

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,66 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+/// replicated sign bit of the extracted field.
+///
+/// \tparam T The type of \param source value, must be an integer.
+/// \param source The source value to extracting.
+/// \param bit_start The position to start extracting.
+/// \param num_bits The number of bits to extracting.
+template <typename T>
+inline std::enable_if_t<std::is_integral_v<T>, T>
+bfe(T source, uint32_t bit_start, uint32_t num_bits) {
+  unsigned max_bits_num = CHAR_BIT * sizeof(T);
```
zhimingwang36:
```suggestion
  unsigned bit_width = CHAR_BIT * sizeof(T);
```

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,66 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+///
+/// \tparam T The type of \param x and \param y , must be an unsigned integer.
+/// \param x The source of the bitfield.
+/// \param y The source where bitfield is inserted.
+/// \param bit_start The position to start insertion.
+/// \param num_bits The number of bits to insertion.
+template <typename T>
+inline std::enable_if_t<std::is_unsigned_v<T>, T>
+bfi(T x, T y, uint32_t bit_start, uint32_t num_bits) {
+  constexpr unsigned msb = std::numeric_limits<T>::digits;
```
zhimingwang36: msb  ==> bit_width

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,66 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+/// \param num_bits gives the bit field length in bits.
+///
+/// \tparam T The type of \param x and \param y , must be an unsigned integer.
+/// \param x The source of the bitfield.
+/// \param y The source where bitfield is inserted.
+/// \param bit_start The position to start insertion.
+/// \param num_bits The number of bits to insertion.
+template <typename T>
+inline std::enable_if_t<std::is_unsigned_v<T>, T>
+bfi(T x, T y, uint32_t bit_start, uint32_t num_bits) {
```
zhimingwang36: please add const for the args which will not changed.

yihwang-nv: done

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: done

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: done

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,138 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+    asm volatile("bfe.s32 %0, %1, %2, %3;"
+                 : "=r"(res)
+                 : "r"(source), "r"(bit_start), "r"(num_bits));
+    return res;
+  } else if constexpr (std::is_same_v<T, uint32_t>) {
+    T res{};
+    asm volatile("bfe.u32 %0, %1, %2, %3;"
+                 : "=r"(res)
+                 : "r"(source), "r"(bit_start), "r"(num_bits));
+    return res;
```
danhoeflinger: I know less about what the compiler can take care of with these explicit asm instructions, but will we pay a penalty for a cast of a uint32_t to a uint32_t?  
In other words, do we actually gain anything from separating the 32 bit versions from the 8 and 16 bit cases here?

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,138 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+      const T sign_bit_padding = (-sign_bit & ~mask);
+      return ((source >> pos) & mask) | sign_bit_padding;
+    } else {
+      // Is T is an unsigned integer type, the sign-bit does not need to be
+      // padded.
+      return (source >> pos) & mask;
+    }
+#if defined(__SYCL_DEVICE_ONLY__) && defined(__NVPTX__)
+  }
+#endif
```
danhoeflinger: You should be able to remove this preprocessor check by moving the `#endif` above between the `else` and the `{`.
It should be cleaner.

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,138 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+                 : "=r"(res)
+                 : "r"((uint32_t)x), "r"((uint32_t)y), "r"(bit_start),
+                   "r"(num_bits));
+    return res;
+  } else if constexpr (std::is_same_v<T, uint32_t>) {
+    T res{};
+    asm volatile("bfi.b32 %0, %1, %2, %3, %4;"
+                 : "=r"(res)
+                 : "r"(x), "r"(y), "r"(bit_start), "r"(num_bits));
+    return res;
```
danhoeflinger: Same question as above.

danhoeflinger (commented):
As I've mentioned before, I'd be in favor of separating the operation itself out from the guard rail checks, and exposing the unguarded version as well. 
 
It should be easy to do and I think it could be done in this PR, but if you want to hold that for a later PR, that would be OK.  I'll mention that if someone is using the current dpct::BFE (in main without guards), this PR would be a performance regression.  

I've only reviewed the math.hpp implementation, not the SYCLomatic migration part of this PR, I'll leave that for someone who has more understanding of those systems.

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -242,6 +243,138 @@ inline unsigned unordered_compare_mask(const sycl::marray<T, 2> a,
...
+                 : "=l"(res)
+                 : "l"(source), "r"(bit_start), "r"(num_bits));
+    return res;
+  } else if constexpr (std::is_same_v<T, uint64_t>) {
+    T res{};
+    asm volatile("bfe.u64 %0, %1, %2, %3;"
+                 : "=l"(res)
+                 : "l"(source), "r"(bit_start), "r"(num_bits));
+    return res;
+  } else {
```
zhimingwang36: we don't need this else actually.

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: I have merged these 2 branch.

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: done.

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: done

clang/runtime/dpct-rt/include/dpct/math.hpp
yihwang-nv: done

yihwang-nv:
> As I've mentioned before, I'd be in favor of separating the operation itself out from the guard rail checks, and exposing the unguarded version as well.
> 
> It should be easy to do and I think it could be done in this PR, but if you want to hold that for a later PR, that would be OK. I'll mention that if someone is using the current dpct::BFE (in main without guards), this PR would be a performance regression.
> 
> I've only reviewed the math.hpp implementation, not the SYCLomatic migration part of this PR, I'll leave that for someone who has more understanding of those systems.

I have introduce a new `bfe` function and rename the original function to `bfe_safe`.

danhoeflinger:
> I have introduce a new `bfe` function and rename the original function to `bfe_safe`.

Thanks I think this looks good.  

Sorry to keep requesting things, but I'd like to see the same breakdown for `bfi` and `bfi_safe`, so that there is a consistent structure.  If different mappings get sent to different `bfi` varieties it would be best to do this now, rather than have to change existing mappings later.

yihwang-nv:
> > I have introduce a new `bfe` function and rename the original function to `bfe_safe`.
> 
> Thanks I think this looks good.
> 
> Sorry to keep requesting things, but I'd like to see the same breakdown for `bfi` and `bfi_safe`, so that there is a consistent structure. If different mappings get sent to different `bfi` varieties it would be best to do this now, rather than have to change existing mappings later.

Thanks for your review, I have add a `bfi_safe` function.

danhoeflinger (approved):
LGTM (the dpl_extras/*.h part),  I'll leave the review of the rest to those on the SYCLomatic team.
Thanks!

zhimingwang36 (approved):
LGTM

PR #1483, opened 2023-12-05
[SYClomatic] Add block /scatter to striped exchange header 

abhilash1910:
Add cub::BlockExchange::BlockedToStriped header 
Edit: Added (Also cub::BlockExchange::StripedToBlocked will be added)
cc @yihanwg


abhilash1910:
@timmiesmith @danhoeflinger   @mmichel11  could you please help review ? Thanks

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -399,6 +400,34 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  /// Rearrange elements from blocked order to striped order
+  template <typename Item>
+  __dpct_inline__ void blocked_to_striped(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
+    T *buffer = reinterpret_cast<T *>(_local_memory);
+
+#pragma unroll
+    for (int i = 0; i < VALUES_PER_THREAD; i++) {
+      int offset = (item.get_local_id(0) * VALUES_PER_THREAD) + i;
```
danhoeflinger: Can we consolidate redundant code between `blocked_to_striped` and `scatter_to_blocked`?
The only place they differ is in generating the offsets.  
Either by passing lambdas or having constexpr branches and some named offset generating helpers,  I think this could be consolidated.

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -399,6 +400,34 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      if (INSERT_PADDING)
```
mmichel11: Since `INSERT_PADDING` is a constexpr variable, could we make this an `if constexpr (...)`?

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
```diff
@@ -399,6 +400,34 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    }
+
+    item.barrier(sycl::access::fence_space::local_space);
+
+#pragma unroll
+    for (int i = 0; i < VALUES_PER_THREAD; i++) {
+      int offset = int(i * item.get_local_range(2) * item.get_local_range(1) *
+                       item.get_local_range(0)) +
+                   item.get_local_id(0);
+      if (INSERT_PADDING)
```
mmichel11: Same comment as above with regards to `if constexpr (...)`.

mmichel11:
Does this API require a test case in https://github.com/oneapi-src/SYCLomatic-test?

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
abhilash1910: Added this , thanks

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
abhilash1910: @danhoeflinger  yes I have refactored the code to better align with this design. Could you please take a look . Thanks

abhilash1910:
@danhoeflinger @mmichel11  requesting re-review. Also could you suggest if test is needed in syclomatic test repo - as I could nt find scatter to blocked tests as well.

danhoeflinger:
@abhilash Yes, I think both of these deserve a test in SYCLomatic-test.  I think the reason you find no existing test for scatter to blocked is that it was just a helper function for this radix sort implementation, it was not originally implemented as a migration target on its own.  
If we intend to use them for this purpose, both should have SYCLomatic-tests.  

I'll review the updates shortly.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,28 +374,102 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
 
   exchange(uint8_t *local_memory) : _local_memory(local_memory) {}
 
+  template <typename Item>
+  __dpct_inline__ int get_offset(Item item,
+               T (&keys)[VALUES_PER_THREAD],
+               int (&ranks)[VALUES_PER_THREAD],
+               int offset,
+               int i) {
+      int offset_helper = [&](Item item,
```
mmichel11: Is `offset_helper` needed here since it's only called once? I think just having the offset calculation without the lambda is more straightforward.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -404,7 +479,7 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
   static constexpr bool INSERT_PADDING =
       (VALUES_PER_THREAD > 4) &&
       (detail::power_of_two<VALUES_PER_THREAD>::VALUE);
-
+  int offset = 0;
```
mmichel11: I see why you made offset a member variable but in my opinion having it local to each function makes more sense. Otherwise, it implies that there is some state to this functor in-between uses.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes correct, it is not needed . Adjusted.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Made it localised to the functions.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,18 +374,30 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
 
   exchange(uint8_t *local_memory) : _local_memory(local_memory) {}
 
+  template <typename Item>
+  __dpct_inline__ int get_offset(Item item,
```
mmichel11: Should this helper be called `compute_padding` or something more specific?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,18 +374,30 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
-  __dpct_inline__ void
-  scatter_to_blocked(Item item, T (&keys)[VALUES_PER_THREAD],
-                     int (&ranks)[VALUES_PER_THREAD]) {
+  __dpct_inline__ void scatter_to_blocked(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
+                                          
     T *buffer = reinterpret_cast<T *>(_local_memory);
-
+
```
mmichel11: I think some whitespace has been added here which should be removed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -404,7 +471,6 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
   static constexpr bool INSERT_PADDING =
       (VALUES_PER_THREAD > 4) &&
       (detail::power_of_two<VALUES_PER_THREAD>::VALUE);
-
```
mmichel11: I think we should just add this line back since it was originally there.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,18 +374,30 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  __dpct_inline__ int get_offset(Item item,
+               T (&keys)[VALUES_PER_THREAD],
+               int (&ranks)[VALUES_PER_THREAD],
+               int offset,
+               int i) {
+    if constexpr (INSERT_PADDING) {
+        offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
+    }
+    return offset;
+  }
```
danhoeflinger: It seems the only parameter used by this function is `offset`.  Why are the rest passed in?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Should this be named something more like `adjust_by_padding()`?  That seems more to be the purpose of the function (though I'm a bit confused by it generally).

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,18 +374,30 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
   exchange(uint8_t *local_memory) : _local_memory(local_memory) {}
 
+  template <typename Item>
+  __dpct_inline__ int get_offset(Item item,
+               T (&keys)[VALUES_PER_THREAD],
+               int (&ranks)[VALUES_PER_THREAD],
+               int offset,
+               int i) {
+    if constexpr (INSERT_PADDING) {
+        offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
```
danhoeflinger: It seems that `LOG_LOCAL_MEMORY_BANKS` is just hard coded to `5`.  Is this applicable to the general case (the 2 new APIs), or is it just being propagated from existing code?

Do we understand why it exists and why it is `5` (is it for specific hardware)?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -393,8 +406,62 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      keys[i] = buffer[offset];
+    }
+    
+  }
+
+  /// Rearrange elements from blocked order to striped order
+  template <typename Item>
+  __dpct_inline__ void blocked_to_striped(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
```
danhoeflinger: What is the point of `ranks` in this function? it seems to be unused...

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -393,8 +406,62 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      offset = get_offset(item, keys, ranks, offset, i);
+      keys[i] = buffer[offset];
+    }
+  }
+
+  /// Rearrange elements from striped order to blocked order
+  template <typename Item>
+  __dpct_inline__ void striped_to_blocked(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
```
danhoeflinger, line 459: What is the point of `ranks` in this function? it seems to be unused...

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Same question for the padding in general?  Is this what we want for the general case?

clang/runtime/dpct-rt/include/dpl_extras/dpcpp_extensions.h
danhoeflinger: I don't really see this as resolved, but perhaps it is more trouble than it is worth.  Let me better explain my intentions, and you can decide if its helpful or if its adding too much complication in order to remove the redundancy.  I don't see it as required.

Right now we have 3 copies of effectively the same function, the difference between them is the calculation of the incoming load offset index and the calculation of the outgoing store offset index.  
Moreover, for blocked to striped and striped to blocked, the offset index calculation for the input matches the offset index calculation of the output for the other and vice versa.  

Effectively you have one function helper (pseudocode):
```
helper_exchange(normal_inputs, offset_functor_input, offset_functor_output)
{
    reinterpret exchange temporary memory
    unrolled normal load loop, but calling offset_functor_input to get input index from item and `i`
    
    barrier

    unrolled normal store loop but calling offset_functor_output to get output index from item
}
```
Then you would have a struct functor definition for the blocked index calculation, and another for the striped index calculation from `item` and `i`.  

To reuse the building block for the scatter to blocked, you'd also need to create a functor to obtain the offset from `ranks`.

Finally, you would have the external APIs which use those building blocks to implement the general APIs.

mmichel11 (commented):
A few formatting suggestions from me. I also agree with the suggestion from @danhoeflinger that tests should be written for the two APIs.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes from what I see LOG_LOCAL_MEMORY_RANKS is hardcoded but that necessary should  not be the case. 
I guess for generic padding the functionality seems to be the same for different algos on cub side .

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Addressed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes changed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes ranks are not needed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Resolved

abhilash1910:
@danhoeflinger  could you review the code flow for the striped-> blocked functor pathway.? I will rectify the format and add for the blocked /scatter one  after this seems correct . Could you please help review soon as this is expected to be added before Jan 8 ? Also if it is an overkill , then I can revert back to the standard implementation as before. Thanks

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,18 +374,101 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+                          item.get_local_range(1) * item.get_local_range(0)) +
+                      item.get_local_id(0);
+      return adjust_by_padding(fw_offset);
+    }
+
+    int reverse_offset(Item item, int i) {
+      int rv_offset = int(item.get_local_id(0) * VALUES_PER_THREAD) + i;
+      return adjust_by_padding(rv_offset);
+    }
+  };
```
danhoeflinger: This is headed in the right direction I think, but still needs a bit of a tweak.

Here you name one function as `forward_offset` and the other as `reverse_offset`.  Those names only make sense in the context of the striped to blocked operation, but they are also used for the blocked to striped operation.  I don't see any reason to tie them together, it just makes things harder.

If it were me, I would just have two different functors, which each overload the `operator()`, `get_striped_offset` and `get_blocked_offset`.  Then you can send these directly in to `helper_exchange`, without needing a separate struct and member function.  You can just call them with the `operator()` within `helper_exchange`.

Then, if you want, I think this would just require one more functor to implement the `get_scatter_offset` which uses an map to get the index.

abhilash1910: Made the necessary changes , could you please review @danhoeflinger

abhilash1910: Since some methods ( scatter -> block/scatter->striped) rely on ranks, the helper_exchange has an overloaded counterpart with ranks arg .  The scatter -> striped part is also added for review.

danhoeflinger: Continuing to look better.  I think you can reuse the same `helper_exchange` for both cases if you just initialize the scatter functor with the `ranks` array in its constructor.

abhilash1910: Thanks for the suggestion @danhoeflinger  .Made the changes inside the constructor .  Requesting review.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,32 +374,120 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  template <typename Item> struct getScatterOffset {
+    int *ranks_;
+    int ranks_size;
+    getScatterOffset(int size, const int *ranks) : ranks_size(size) {
+      ranks_ = new int[ranks_size];
+      for (int i = 0; i < ranks_size; i++) {
+        ranks_[i] = ranks[i];
+      }
+    }
+    ~getScatterOffset() { delete[] ranks_; }
```
danhoeflinger: I doubt we want to allocate, copy and delete this every time.  Can we just add an iterator template argument to `getScatterOffset`, pass an iterator (which could be a pointer) and use a bracket operator to directly use the input data without the copy?
Probably can avoid storing the size as well.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,32 +374,120 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      int offset = ranks_[i];
+      return adjust_by_padding(offset);
+    }
+  };
+
+  template <typename Item> struct getBlockedFromScatterOffset {
+    int operator()(Item item, int i) {
+      int offset = (item.get_local_id(0) * VALUES_PER_THREAD) + i;
+      return adjust_by_padding(offset);
+    }
```
danhoeflinger: This seems functionally redundant with `getStripedOffset` above.  Also, we should reconsider the names here.  It is odd that very conceptually different names would have the same functionality.

We should stop and think about what each operation is doing within the exchange to determine the naming (rather than just naming according to what it is used for).  There is a copy to a temporary block which is one transformation, then a copy from that temporary block which may be another transformation.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,32 +374,120 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+
+    if constexpr (INSERT_PADDING) {
+      offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
+    }
+    return offset;
+  }
+
+  template <typename Item> struct getStripedOffset {
+    int operator()(Item item, int i) {
+      int offset = int(item.get_local_id(0) * VALUES_PER_THREAD) + i;
```
danhoeflinger: Is the `int()` construction necessary here?  There are some similar places.  I guess we are bringing things back from `size_t` to `int`.  On a related topic, I don't think we can overflow `int` with these offsets, but I also don't think they need to be signed, maybe they should be uint32_t?  (I would also be fine to just make them all `size_t`)

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Have you looked at if this padding actually improves performance (and are we using it in the correct situations and hardcoded values)?  For the best I can understand, it is a performance optimization implementation detail to avoid bank conflicts (though I'm not super confident of that).  

However, it isn't a satisfying reason to propagate the code here just because it was used in existing related or similar code.  Does it improve performance for the target hardware, and in these new APIs?

mmichel11: Just to add onto the discussion here, the [Intel GPU optimization guide](https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2024-0/shared-local-memory.html#BANK-CONFLICTS) states, "At the time of writing, 64 consecutive bytes are stored in 16 consecutive banks at 4-byte (32-bit) granularity."

I think this would mean that the `LOG_LOCAL_MEMORY_BANKS` should be `4` for Intel GPUs and that we might run into bank conflicts with a hardcoded padding for `32` banks.

Potentially, we could use a solution similar to [_ONEDPL_DETECT_SPIRV_COMPILATION](https://github.com/oneapi-src/oneDPL/blob/01afb07233681f3092c8580f6189ca7e8cf0728b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h#L62) to differentiate between compilation for different backends, but I think it would be only worth exploring if it has meaningful performance benefit. It might not also be a long-term solution depending on future device architectures.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes this is not required ; from functionality point of view all of them are mostly transpose / scatter operations. I adjusted the name to reflect only 3 transformations which can be re-used.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: size_t can be used but in the case of accessing buffer elements (buffer[offset] )  will uint8_t conversion not create any issues on different systems ? I have made the change to return the offsets as size_t (uint32 should also have been ok).

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: For the time being , can this be kept as 4 instead of 5. Since this is the functional side implementation, the perf has to be evaluated post the changes. A macro works too but having conditional macros will increase the code when more architectures comes into the picture.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Modified to use a pointer iterator template arg. Avoids memcpy of array /size.

abhilash1910:
Tagging @zhimingwang36  for review .

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,117 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    getScatterOffset(const int (&ranks)[VALUES_PER_THREAD]) {
+      begin = ranks;
+      end = ranks + VALUES_PER_THREAD;
+    }
+    ~getScatterOffset() {}
+    size_t operator()(Item item, int i) const {
+      Iterator it = begin + i;
+      size_t offset;
+      if (it >= begin && it < end) {
+        offset = size_t(*it);
```
danhoeflinger:
```suggestion
        offset = *it;
```
I see no reason to explicitly construct a size_t here, when the `=` will provide that conversion implicitly.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,117 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      return adjust_by_padding(offset);
+    }
+  };
+
+  template <typename Item> struct getBlockedOffset {
+    size_t operator()(Item item, int i) {
+      size_t offset =
+          size_t(i * item.get_local_range(2) * item.get_local_range(1) *
+                 item.get_local_range(0)) +
+          item.get_local_id(0);
```
danhoeflinger:
```suggestion
      size_t offset =
          i * item.get_local_range(2) * item.get_local_range(1) *
                 item.get_local_range(0) +
          item.get_local_id(0);
```
Similarly, I don't see a need to construct a size_t here, the conversions should happen implicitly.  (Please let me know if I'm missing something here.)

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,117 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    if constexpr (INSERT_PADDING) {
+      offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
+    }
+    return offset;
+  }
+
+  template <typename Item> struct getStripedOffset {
+    size_t operator()(Item item, int i) {
+      size_t offset =
+          size_t(item.get_local_id(0) * VALUES_PER_THREAD) + size_t(i);
```
danhoeflinger:
```suggestion
          item.get_local_id(0) * VALUES_PER_THREAD + i;
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,117 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  };
+
+  template <typename Item, typename Iterator> struct getScatterOffset {
+    Iterator begin;
+    Iterator end;
+    getScatterOffset(const int (&ranks)[VALUES_PER_THREAD]) {
+      begin = ranks;
+      end = ranks + VALUES_PER_THREAD;
+    }
+    ~getScatterOffset() {}
```
danhoeflinger: You can remove this destructor now, it does nothing.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Changed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,117 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    size_t operator()(Item item, int i) const {
+      Iterator it = begin + i;
+      size_t offset;
+      if (it >= begin && it < end) {
+        offset = *it;
+      }
+      return adjust_by_padding(offset);
+    }
```
danhoeflinger:
```suggestion
    size_t operator()(Item item, int i) const {
      return adjust_by_padding(begin[i]);
    }
```
I think its reasonable to assume in this setting that begin is a random access iterator, and also that `i` is within bounds of `VALUES_PER_THREAD`.  You can add some comments here stating assumptions, but these checks will slow things down in a low level helper and not actually prevent any errors.
If you take this suggestion, you can also remove the `end` iterator.

abhilash1910: Modified.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,111 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+          i * item.get_local_range(2) * item.get_local_range(1) *
+                 item.get_local_range(0) +
+          item.get_local_id(0);
+      return adjust_by_padding(offset);
+    }
+  };
+
+  template <typename Item, typename Iterator> struct getScatterOffset {
+    Iterator begin;
+    Iterator end;
```
danhoeflinger:
```suggestion
```
Still need to remove this line.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,112 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    return offset;
+  }
+
+  template <typename Item> struct getStripedOffset {
+    size_t operator()(Item item, int i) {
+      size_t offset =
+          item.get_local_id(0) * VALUES_PER_THREAD + i;
+      return adjust_by_padding(offset);
+    }
+  };
```
danhoeflinger:
```suggestion
  struct getStripedOffset {
    template <typename Item>
    size_t operator()(Item item, int i) {
      size_t offset =
          item.get_local_id(0) * VALUES_PER_THREAD + i;
      return adjust_by_padding(offset);
    }
  };
```

If you make the `operator()` the template rather than the `struct`, then you don't have to specify the type for all the operators when you create them.  The operator itself will deduce the type required based on the function argument.

`getStripedOffset<Item> get_striped_offset;`
->
`getStripedOffset get_striped_offset;`

The same goes for the other two functors.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+
+  template <typename Item, typename Iterator> struct getScatterOffset {
+    Iterator begin;
+    getScatterOffset(const int (&ranks)[VALUES_PER_THREAD]) { begin = ranks; }
+    size_t operator()(Item item, int i) const {
+      // iterator i is expected to be within bounds [0,VALUES_PER_THREAD)
+      return adjust_by_padding(begin[i]);
+    }
+  };
```
danhoeflinger:
```suggestion
  template <typename Iterator> struct getScatterOffset {
    Iterator begin;
    getScatterOffset(const int (&ranks)[VALUES_PER_THREAD]) { begin = ranks; }
    template <typename Item>
    size_t operator()(Item item, int i) const {
      // iterator i is expected to be within bounds [0,VALUES_PER_THREAD)
      return adjust_by_padding(begin[i]);
    }
  };
```

This should allow `getScatterOffset<Item, int *> get_scatter_offset(ranks);` -> `getScatterOffset<int*> get_scatter_offset(ranks);` .

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  // and if specializations are required for specific target hardware.
+  static size_t adjust_by_padding(size_t offset) {
+
+    if constexpr (INSERT_PADDING) {
+      offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
+    }
+    return offset;
+  }
+
+  template <typename Item> struct getStripedOffset {
```
yihwang-nv: Let's use snake_case naming rule in help function.
```suggestion
  template <typename Item> struct get_striped_offset {
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    size_t operator()(Item item, int i) {
+      size_t offset = item.get_local_id(0) * VALUES_PER_THREAD + i;
+      return adjust_by_padding(offset);
+    }
+  };
+
+  template <typename Item> struct getBlockedOffset {
```
yihwang-nv:
```suggestion
  template <typename Item> struct get_blocked_offset {
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    size_t operator()(Item item, int i) {
+      size_t offset = i * item.get_local_range(2) * item.get_local_range(1) *
+                          item.get_local_range(0) +
+                      item.get_local_id(0);
+      return adjust_by_padding(offset);
+    }
+  };
+
+  template <typename Item, typename Iterator> struct getScatterOffset {
```
yihwang-nv:
```suggestion
  template <typename Item, typename Iterator> struct get_scatter_offset {
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    Iterator begin;
+    getScatterOffset(const int (&ranks)[VALUES_PER_THREAD]) { begin = ranks; }
```
yihwang-nv:
```suggestion
    get_scatter_offset(const int (&ranks)[VALUES_PER_THREAD]) { begin = ranks; }
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+    size_t operator()(Item item, int i) const {
```
yihwang-nv: Is it possible that i here is a negative number? If not, size_t or unsigned should be used.

yihwang-nv:
@abhilash1910 Thank you working on this, can you please also provide a test patch in SYCLomatic-test repo?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: No i always starts from 0. In this case, size_t may not be needed and int should work.

danhoeflinger: I think @yihanwg is suggesting the opposite. Specifically, since we never should have a negative offset, that `int` should instead be `uint32_t` or `size_t`.

abhilash1910: Yes uint32 or size_t both works since it is positive signed. But is it a mandatory requirement?

abhilash1910: Added

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Added

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Modified

abhilash1910:
> @abhilash1910 Thank you working on this, can you please also provide a test patch in SYCLomatic-test repo?

Yes @yihanwg  I have the changes locally on the test repo, I will add the PR there.  Since the feature request is urgent hence added this first

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,109 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  template <typename Item, typename offsetFunctorTypeFW,
+            typename offsetFunctorTypeRV>
+  __dpct_inline__ void helper_exchange(Item item, T (&keys)[VALUES_PER_THREAD],
+                                       offsetFunctorTypeFW &offset_functor_fw,
+                                       offsetFunctorTypeRV &offset_functor_rv) {
+
     T *buffer = reinterpret_cast<T *>(_local_memory);
 
 #pragma unroll
     for (int i = 0; i < VALUES_PER_THREAD; i++) {
```
danhoeflinger:
```suggestion
    for (size_t i = 0; i < VALUES_PER_THREAD; i++) {
```

This should be switched along with the offset functors since we made that change.  Same for below.

danhoeflinger (approved):
LGTM, thanks for all the changes.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+
+    if constexpr (INSERT_PADDING) {
+      offset = detail::shr_add(offset, LOG_LOCAL_MEMORY_BANKS, offset);
+    }
+    return offset;
+  }
+
+  struct get_striped_offset {
+    template <typename Item> size_t operator()(Item item, size_t i) {
+      size_t offset = item.get_local_id(0) * VALUES_PER_THREAD + i;
```
mmichel11, line 389: This looks like a blocked offset to me. We are accessing contiguous elements in a single work item. Should this functor be called `get_blocked_offset`?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+      return adjust_by_padding(offset);
+    }
+  };
+
+  struct get_blocked_offset {
+    template <typename Item> size_t operator()(Item item, size_t i) {
+      size_t offset = i * item.get_local_range(2) * item.get_local_range(1) *
```
mmichel11, line 396: This looks like a striped offset to me. Adjacent work items are accessing adjacent elements in memory which is a striped pattern. Should this functor be called `get_striped_offset`?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  }
+
+  /// Rearrange elements from rank order to blocked order
+  template <typename Item>
+  __dpct_inline__ void scatter_to_blocked(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
+
+    get_scatter_offset<int *> getScatterOffset(ranks);
+    get_striped_offset getStripedOffset;
```
mmichel11: Why are we using a striped offset if we want to store in a blocked arrangement?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -373,34 +374,108 @@ template <typename T, int VALUES_PER_THREAD> class exchange {
...
+  }
+
+  /// Rearrange elements from scatter order to striped order
+  template <typename Item>
+  __dpct_inline__ void scatter_to_striped(Item item,
+                                          T (&keys)[VALUES_PER_THREAD],
+                                          int (&ranks)[VALUES_PER_THREAD]) {
+
+    get_scatter_offset<int *> getScatterOffset(ranks);
+    get_blocked_offset getBlockedOffset;
```
mmichel11: Why are we using a blocked offset if we want to store in a striped arrangement?

mmichel11 (commented):
I have a few comments related to the naming of striped and blocked offsets and confusion this might create.

After these are discussed, I am ready to approve.

danhoeflinger (changes requested):
I agree with @mmichel11 on the naming.  
I would swap the names `get_striped_offset` and `get_blocked_offset`.  
You'll need to then swap the one used for the `scatter_to_x` API, which will make more sense.
You can leave the ordering of the  `striped_to_blocked` and `blocked_to_striped` APIs though.  In practice, those should have the equivalent affect to each other, just with a different organization of the temporary data.  That is why it works currently despite the opposite labelling of the offset calculations.

In the future we may want to revisit from a performance perspective which is faster, `striped_to_blocked` and `blocked_to_striped`, because if one provides an advantage, I think we could use that internal implementation for both.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Addressed.

abhilash1910:
@danhoeflinger @mmichel11 yes made the changes in the api names. I think though that later these names need to be rechecked in another PR. 
@zhimingwang36  could you put it in merge queue once CI pass. Thanks.

zhimingwang36 (approved):
LGTM

PR #1498, opened 2023-12-07
[SYCLomatic] Support migration of asm vadd4/vsub4/vabsdiff4/vmin4/vmax4.

tangjj11:


clang/runtime/dpct-rt/include/math.hpp
```diff
@@ -712,6 +712,25 @@ template <typename T> sycl::vec<int32_t, 2> extractAndExtend2(T a) {
   return ret;
 }
 
+template <typename T> sycl::vec<int32_t, 4> extractAndExtend4(T a) {
```
yihwang-nv: Should we extend each half word to T?

tangjj11: [CUDA doc](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#simd-video-instructions) extended them to signed 17-bit, so we use int32.

yihwang-nv (approved):
LGTM

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -712,6 +712,25 @@ template <typename T> sycl::vec<int32_t, 2> extractAndExtend2(T a) {
   return ret;
 }
 
+template <typename T> sycl::vec<int32_t, 4> extractAndExtend4(T a) {
+  sycl::vec<int32_t, 4> ret;
+  sycl::vec<T, 1> va{a};
+  if constexpr (std::is_signed_v<T>) {
+    auto v = va.template as<sycl::vec<int8_t, 4>>();
+    ret[0] = zero_or_signed_extent(v[0], 17);
```
zhimingwang36: comment1: v[0] is int8, so Line 720, we only 9 bit is enough.
comment2:  zero_or_signed_extent() return int64, should here using a zero_or_signed_extent() which return a 32int.

tangjj11: fixed

zhimingwang36 (approved):
LGTM

PR #1516, opened 2023-12-12
[SYCLomatic] Fix bug that ${string} content by escaping the $ character is not treated as a normal string in the input pattern match

tomflinda:


zhiweij1 (approved):

zhimingwang36 (approved):
LGTM

PR #1519, opened 2023-12-13
[SYCLomatic] Fix 3 migration issues

zhiweij1:
1. Migrate cusparseHandle_t and cusolverDnHandle_t dpct::queue_ptr
2. Migrate cudaLaunchAttributeValue/cudaStreamAttrValue to int
3. Add typecast for the result of dim3 variable member access if it is used as an argument of std::min/std::max


wenhuiNi (approved):

zhimingwang36 (approved):
LGTM

PR #1526, opened 2023-12-14
[SYCLomatic] Add 3 Partition mappings

mmichel11:
This PR adds three SYCLomatic mappings to the compatibility headers:

- dpct::partition_if from an input to a single output buffer.
- dpct::partition_if from an input to three output buffers - elements satisfying pred1, pred2, or neither
- dpct::partition_flagged from an input to a single output buffer where a flag buffer dictates the partition where each element from the input belongs.

The corresponding test PR for these additions is https://github.com/oneapi-src/SYCLomatic-test/pull/552.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
@@ -2414,6 +2414,63 @@ void nontrivial_run_length_encode(ExecutionPolicy &&policy,
   ::std::fill(policy, num_runs, num_runs + 1, ret_dist);
 }
 
+template <typename ExecutionPolicy, typename InputIterator,
+          typename OutputIterator, typename CountIterator,
+          typename UnaryPredicate>
+void partition_if(ExecutionPolicy &&policy, InputIterator input,
+                  OutputIterator output, CountIterator num_true, int num_items,
+                  UnaryPredicate pred) {
+  dpl::copy(policy, input, input + num_items, output);
+  auto beg_false =
+      dpl::stable_partition(policy, output, output + num_items, pred);
+  dpl::reverse(policy, beg_false, output + num_items);
+  auto num_true_items = ::std::distance(output, beg_false);
+  dpl::fill(policy, num_true, num_true + 1, num_true_items);
+}
+
+template <typename ExecutionPolicy, typename InputIterator,
+          typename OutputIterator1, typename OutputIterator2,
+          typename OutputIterator3, typename CountIterator,
+          typename UnaryPredicate1, typename UnaryPredicate2>
+void partition_if(ExecutionPolicy &&policy, InputIterator input,
+                  OutputIterator1 output1, OutputIterator2 output2,
+                  OutputIterator3 output3, CountIterator partition_counts,
+                  int num_items, UnaryPredicate1 pred1, UnaryPredicate2 pred2) {
+  using internal::__buffer;
+  using ValueType = typename ::std::iterator_traits<InputIterator>::value_type;
+  using CountType = typename ::std::iterator_traits<CountIterator>::value_type;
+  __buffer<ValueType> tmp(num_items);
+
+  // Partition the first set (elements that satisfy pred1) and store the
+  // remaining in temporary storage.
+  auto [output1_end, tmp_end] = dpct::stable_partition_copy(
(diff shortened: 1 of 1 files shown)


danhoeflinger (commented):
Generally, I think these look good. As I mentioned offline,  I think it would be beneficial to make the "reverse" part of the semantics optional via some run-time parameter with the default to apply the reverse.  

This would allow an easy performance improvement if the reverse part is not relied upon.

mmichel11:
> Generally, I think these look good. As I mentioned offline, I think it would be beneficial to make the "reverse" part of the semantics optional via some run-time parameter with the default to apply the reverse.
> 
> This would allow an easy performance improvement if the reverse part is not relied upon.

Thanks for this suggestion. I have added this as a runtime parameter with the default value as true.

I will update the test PR to test both options.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -2414,6 +2414,70 @@ void nontrivial_run_length_encode(ExecutionPolicy &&policy,
...
+          typename OutputIterator, typename CountIterator,
+          typename UnaryPredicate>
+void partition_if(ExecutionPolicy &&policy, InputIterator input,
+                  OutputIterator output, CountIterator num_true, int num_items,
+                  UnaryPredicate pred, bool reverse_last = true) {
+  dpl::copy(policy, input, input + num_items, output);
+  auto beg_false =
+      dpl::stable_partition(policy, output, output + num_items, pred);
+  auto num_true_items = ::std::distance(output, beg_false);
+  dpl::fill(policy, num_true, num_true + 1, num_true_items);
```
danhoeflinger: Since we are using the `policy` for multiple calls to oneDPL, do we need to wrap this policy uniquely for each call similar to what is done inside of oneDPL when multiple patterns / kernels are launched inside an implementation?

(The same comment for each of these APIs)

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -2414,6 +2414,70 @@ void nontrivial_run_length_encode(ExecutionPolicy &&policy,
...
+  if (reverse_last)
+    dpl::reverse(policy, beg_false, output + num_items);
```
danhoeflinger: Should we be forwarding the policy for this last call? (this is probably impacted by the policy wrapping comment above)

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
mmichel11: Yes, that is a good point. This change is necessary.

mmichel11: Yes, that is a good point. This change is necessary. Would calling the oneDPL internal `make_wrapped_policy` to do so be acceptable here?

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
mmichel11: Yes, it is probably worth it then for me to make the fill kernel go last and forward this call since we know it will always be called.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: `oneapi::dpl::__par_backend_hetero::make_wrapped_policy` is an implementation detail from oneDPL so as a general rule here it should not be used.  For something like this, its not been a problem to duplicate the functionality here in the dpct namespace, with the appropriate comments indicating it and then use that version of the code.

mmichel11: I have added this as a utility in the `internal` namespace and wrapped around the policies of the different oneDPL calls.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
mmichel11: Done.

danhoeflinger (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1561, opened 2023-12-27
[CI][GHA] Update clangformat with exclude folder

DoyleLi:


DoyleLi:
From the debug github action run, we can see the clang-format will only check necessary files:
https://github.com/oneapi-src/SYCLomatic/actions/runs/7335459743/job/19973403020?pr=1561

I will create a new commit that only contains change for github action job.

intwanghao (approved):
LGTM

PR #1562, opened 2023-12-27
[SYClomatic] Add cub::RadixSort.SortBlockedToStriped header

abhilash1910:
PR to add SortBlockedToStriped sorting ,
depends on #1483 (scatter_to_striped added in exchange)
cc @mmichel11 @danhoeflinger @yihanwg


clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -460,6 +460,49 @@ class radix_sort {
...
+
+    uint32_t(&unsigned_keys)[VALUES_PER_THREAD] =
+        reinterpret_cast<uint32_t(&)[VALUES_PER_THREAD]>(keys);
+
+#pragma unroll
+    for (int i = 0; i < VALUES_PER_THREAD; ++i) {
+      unsigned_keys[i] = detail::traits<T>::twiddle_in(unsigned_keys[i]);
+    }
+
+    while (true) {
```
mmichel11: Is it possible to reorganize this loop a bit to have an explicit condition instead of `while (true)` with a break in the middle of the loop? It makes the code a bit more clear in my opinion.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -460,6 +460,49 @@ class radix_sort {
...
+      detail::radix_rank<RADIX_BITS, DESCENDING>(_local_memory)
+          .template rank_keys(item, unsigned_keys, ranks, begin_bit, pass_bits);
+      begin_bit += RADIX_BITS;
+
+      item.barrier(sycl::access::fence_space::local_space);
+
+      if (begin_bit >= end_bit) {
+
+        exchange<T, VALUES_PER_THREAD>(_local_memory)
+            .scatter_to_striped(item, keys);
```
mmichel11: Do we need to synchronize here before we exit the loop and call `twiddle_out` over the keys?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -460,6 +460,49 @@ class radix_sort {
...
+  __dpct_inline__ void
+  sortBlockedToStriped(const Item &item, T (&keys)[VALUES_PER_THREAD],
+                       int begin_bit = 0, int end_bit = 8 * sizeof(T)) {
+
+    uint32_t(&unsigned_keys)[VALUES_PER_THREAD] =
+        reinterpret_cast<uint32_t(&)[VALUES_PER_THREAD]>(keys);
+
+#pragma unroll
+    for (int i = 0; i < VALUES_PER_THREAD; ++i) {
+      unsigned_keys[i] = detail::traits<T>::twiddle_in(unsigned_keys[i]);
```
mmichel11: My understanding is that `twiddle_in` and `twiddle_out` throw a runtime exception when it is not a supported type. Maybe it is not worth addressing here, but would it be better to catch this at compile time instead based on the type of `T`?

mmichel11 (commented):
In addition to the comments here, I think a test case would also be nice to ensure correctness in the SYCLomatic-test repo.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: I do not think a local barrier is required here since we call a barrier outside the exchange loop before twiddling out.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yeah I thought so too, but I am not sure how to modify this.

abhilash1910:
Thanks @mmichel11  for the suggestions,  I think the modifications can be addressed later as well since this PR is for basic functionality . I am in progress of adding test suite for this sorting algos.  Since there is a feature freeze on 8th, if it looks ok I can put it in for merging .  Any modifications on this can be addressed on another PR as well . Let me know your thoughts. Thanks.

Also tagging @zhimingwang36  for review.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -460,6 +460,49 @@ class radix_sort {
...
 
+#pragma unroll
+    for (int i = 0; i < VALUES_PER_THREAD; ++i) {
+      unsigned_keys[i] = detail::traits<T>::twiddle_out(unsigned_keys[i]);
+    }
+  }
+
+  template <typename Item>
+  __dpct_inline__ void
+  sortBlockedToStriped(const Item &item, T (&keys)[VALUES_PER_THREAD],
```
yihwang-nv: Let's keep the same naming convention as elsewhere

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
mmichel11: I think this should be `.scatter_to_striped(item, keys, ranks);`

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -460,6 +460,49 @@ class radix_sort {
...
+                       int begin_bit = 0, int end_bit = 8 * sizeof(T)) {
+
+    uint32_t(&unsigned_keys)[VALUES_PER_THREAD] =
+        reinterpret_cast<uint32_t(&)[VALUES_PER_THREAD]>(keys);
```
yihwang-nv: Why should we convert to uint32_t instead of using T directly?

yihwang-nv:
@abhilash1910 Thank you working on this, can you please also provide a test patch in SYCLomatic-test repo?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Its just iterating from `begin_bit` to `end_bit` by the `RADIX_BITS` with a branch for the final iteration.  Should be doable with a for loop and an if/else statement.

danhoeflinger (commented):
At a high level, is the only difference between this and the `radix_sort` function that we are doing the final exchange from `scatter_to_striped` rather than `scatter_to_blocked`?  

If so, it seems like we shouldn't copy / rewrite the entire function to accomplish this.  We should be able to just adjust that final exchange and rely on the same code otherwise.  This can be done a number of ways, with a simple helper and a branch, or via some functor / lambda.

abhilash1910:
> At a high level, is the only difference between this and the `radix_sort` function that we are doing the final exchange from `scatter_to_striped` rather than `scatter_to_blocked`?
> 
> If so, it seems like we shouldn't copy / rewrite the entire function to accomplish this. We should be able to just adjust that final exchange and rely on the same code otherwise. This can be done a number of ways, with a simple helper and a branch, or via some functor / lambda.

Yeah thought so, reduced it to a helper + branch. Addressed other issues as well. 
Can this be reviewed on priority ? @danhoeflinger @mmichel11 @yihanwg .Thanks

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,22 +441,28 @@ class radix_sort {
...
+          .template rank_keys(item, unsigned_keys, ranks, i, pass_bits);
 
       item.barrier(sycl::access::fence_space::local_space);
 
+      if (is_striped && i >= end_bit) {
+
+        exchange<T, VALUES_PER_THREAD>(_local_memory)
+            .scatter_to_striped(item, keys, ranks);
+
+        break;
```
danhoeflinger: You should not need `break;` commands.  You now have explicit loop bounds.  All this break does in practice is terminate the final iteration early.  You should be able to do this with an `if/else` instead, and it will be cleaner code.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,22 +441,28 @@ class radix_sort {
...
-      begin_bit += RADIX_BITS;
+          .template rank_keys(item, unsigned_keys, ranks, i, pass_bits);
 
       item.barrier(sycl::access::fence_space::local_space);
 
+      if (is_striped && i >= end_bit) {
+
+        exchange<T, VALUES_PER_THREAD>(_local_memory)
+            .scatter_to_striped(item, keys, ranks);
+
```
danhoeflinger, line 457: Is there a reason that we need a barrier at the end for the blocked case, but not the striped case?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,22 +441,28 @@ class radix_sort {
...
+        break;
+      }
       exchange<T, VALUES_PER_THREAD>(_local_memory)
           .scatter_to_blocked(item, keys, ranks);
 
       item.barrier(sycl::access::fence_space::local_space);
 
-      if (begin_bit >= end_bit)
+      if (!is_striped && i >= end_bit)
```
mmichel11: I believe this was taken from the implementation of the non-striped sort. This is not needed. The loop termination condition in your for loop will handle when we exit the loop.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,22 +441,28 @@ class radix_sort {
...
 
       int ranks[VALUES_PER_THREAD];
       detail::radix_rank<RADIX_BITS, DESCENDING>(_local_memory)
-          .template rank_keys(item, unsigned_keys, ranks, begin_bit, pass_bits);
-      begin_bit += RADIX_BITS;
+          .template rank_keys(item, unsigned_keys, ranks, i, pass_bits);
 
       item.barrier(sycl::access::fence_space::local_space);
 
+      if (is_striped && i >= end_bit) {
```
mmichel11: With the way the for loop is written, this condition will never be met. When `i` is incremented to be `>= end_bit` in the loop expression, then the condition will not be met and we will terminate the loop.

What you want to do here is modify this check to catch the last iteration.

danhoeflinger: An **optional** suggestion which would improve the code in my opinion would be to just skip this part of the loop for the last iteration and instead of passing a bool `is_striped` flag to the helper, just implement the last exchange specialization in the public functions which call the helper.

mmichel11: I believe the fix here does not correctly resolve the issue and instead adds an extra iteration. For example, if RADIX_BITS is 4 and you are are sorting a 32-bit type, then you will have 8 iterations in your loop. The recent changes to the for loop will falsely cause 9 iterations with `for (int i = begin_bit; i <= end_bit; i += RADIX_BITS)`.

I would recommend you revert the changes to the for loop and follow what @danhoeflinger suggested for the last iteration.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,23 +441,26 @@ class radix_sort {
...
 
       item.barrier(sycl::access::fence_space::local_space);
 
+      if (is_striped && i == end_bit) {
+
+        exchange<T, VALUES_PER_THREAD>(_local_memory)
+            .scatter_to_striped(item, keys, ranks);
+
+        item.barrier(sycl::access::fence_space::local_space);
+      }
```
mmichel11, line 461: An `else` here would be needed for correctness as we would falsely perform a `scatter_to_striped` followed by a `scatter_to_blocked` without it.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -441,23 +441,34 @@ class radix_sort {
...
+          item.barrier(sycl::access::fence_space::local_space);
+        }
+      } else {
+        exchange<T, VALUES_PER_THREAD>(_local_memory)
+            .scatter_to_blocked(item, keys, ranks);
 
-      if (begin_bit >= end_bit)
-        break;
+        item.barrier(sycl::access::fence_space::local_space);
+      }
```
danhoeflinger: There are a couple things still wrong here,

- I don't believe we are guaranteed that `end-bit - begin_bit` is evenly divided by `RADIX_BITS`, so lets not assume it here.
- We have the same code in both else blocks, we can just change the logic and combine those.

We are trying to merge today so here is my suggestion (please test/format):

``` 
      bool last_iter = i + RADIX_BITS > end_bit;
      if ( last_iter && is_striped ) {
          exchange<T, VALUES_PER_THREAD>(_local_memory)
            .scatter_to_striped(item, keys, ranks);
      }
      else{
          exchange<T, VALUES_PER_THREAD>(_local_memory)
            .scatter_to_blocked(item, keys, ranks);
      } 
      item.barrier(sycl::access::fence_space::local_space); 
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Lets not worry about this for now.  It complicates the twiddle out and isn't necessarily better.

mmichel11:
@abhilash1910 I am seeing clang-format failing in CI. Could you resolve this?

abhilash1910:
@mmichel11 @danhoeflinger  fixed the issues. Lets wait for CI to run . Thanks.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: How was this resolved?  
(In general, I'd recommend letting the reviewer resolve threads when they are satisfied, rather than resolving them yourself.)

danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

PR #1569, opened 2024-01-03
[SYCLomatic] Fix the migration of mov instruction

yihwang-nv:
Fix the migration of mov instruction when one of the input/output is a vector expression
```
{
    uint32_t v;
    uint16_t lo, hi;
    asm volatile("mov.b32 {%0, %1}, %2;\n" : "=h"(lo), "=h"(hi) : "h"(v));
}
```


wenhuiNi (approved):

zhimingwang36 (approved):
LGTM

PR #1570, opened 2024-01-03
Resolve symlink while dealing with dpct binary path

the-slow-one:
Not resolving symbolic link to `dpct` might result in incorrect path to _dpct.cmake_. This PR resolves the issue.


tomflinda (approved):
LGTM

PR #1587, opened 2024-01-05
[SYCLomatic] Adjust parameter order of bindless_image_wrapper::attach.

tangjj11:

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
+++ b/clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
@@ -314,10 +314,10 @@ public:
 
   /// Attach linear data to bindless image.
   /// \param [in] data The linear data used to create bindless image.
-  /// \param [in] channel The image channel used to create bindless image.
   /// \param [in] size The size of linear data used to create bindless image.
+  /// \param [in] channel The image channel used to create bindless image.
   /// \param [in] q The queue where the image creation be executed.
-  void attach(void *data, const image_channel &channel, size_t size = UINT_MAX,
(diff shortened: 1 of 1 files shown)


clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -314,10 +314,10 @@ template <class T, int dimensions> class bindless_image_wrapper {
 
   /// Attach linear data to bindless image.
   /// \param [in] data The linear data used to create bindless image.
-  /// \param [in] channel The image channel used to create bindless image.
   /// \param [in] size The size of linear data used to create bindless image.
+  /// \param [in] channel The image channel used to create bindless image.
   /// \param [in] q The queue where the image creation be executed.
-  void attach(void *data, const image_channel &channel, size_t size = UINT_MAX,
+  void attach(void *data, size_t size, const image_channel &channel,
```
zhimingwang36, line 320: what is the reason for the reorder?

tangjj11: Because the old implementation in image.hpp has the same functionality, we need keep them consistent to reuse migration logic. For example:
```bash
without bindless image:
cudaBindTexture --> image_wrapper_base::attach(const void *ptr, size_t count, image_channel channel))
with bindless image:
cudaBindTexture --> bindless_image_wrapper::void attach(void *data, size_t size, const image_channel &channel, sycl::queue &q = get_default_queue())
```

ziranzha (approved):

zhimingwang36 (approved):
LGTM

PR #1589, opened 2024-01-05
[SYCLomatic] iterator_adaptor implementation

danhoeflinger:
Implements a `dpct::iterator_adaptor` which adapts boost's iterator adaptor minimally to achieve what we require for use in oneDPL algorithms.

I've put the implementation in a new header to better isolate the inclusion of boost.  I think that I've added the necessary changes to support that new header, but please take a look at that.  With the elimination of the .inc files, I'm not sure how to only include this if it is used (is this done automatically)?

I have a SYCLomatic-test PR which corresponds and tests this functionality with a simple example.  (https://github.com/oneapi-src/SYCLomatic-test/pull/587)

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_utils.hpp
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_utils.hpp
@@ -20,8 +20,12 @@
 #include "dpl_extras/algorithm.h"
 #include "dpl_extras/numeric.h"
 #include "dpl_extras/iterators.h"
-#include "dpl_extras/iterator_adaptor.h"
 #include "dpl_extras/vector.h"
 #include "dpl_extras/dpcpp_extensions.h"
+
+// Only include iterator adaptor (and therefore boost) if necessary
+#ifdef ITERATOR_ADAPTOR_REQUIRED
+#include "dpl_extras/iterator_adaptor.h"
+#endif //ITERATOR_ADAPTOR_REQUIRED
 
 #endif // __DPCT_DPL_UTILS_HPP__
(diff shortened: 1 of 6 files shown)


clang/runtime/dpct-rt/include/dpct/dpl_utils.hpp
```diff
@@ -20,6 +20,7 @@
 #include "dpl_extras/algorithm.h"
 #include "dpl_extras/numeric.h"
 #include "dpl_extras/iterators.h"
+#include "dpl_extras/iterator_adaptor.h"
```
zhimingwang36: I'd prefer to include this header file only when needed,  then not all thrust migration need boost installed.
 
option1 may be: 0
change L23 to :  
#ifdef ITERATOR_ADAPTOR_REQUIRED 
#include "dpl_extras/iterator_adaptor.h"
#endif  

option2:  remove L23,  have DPCT to generate "#include "dpl_extras/iterator_adaptor.h" on demand. 

one question:  will iterator_adaptor be used in other dpl_extras header files?

danhoeflinger: I agree, it should only be included when needed.  I think option 1 looks better to me.  I assume that the migration code would populate `ITERATOR_ADAPTOR_REQUIRED` if necessary then?

We don't have current plans to iterator_adaptor in other dpl_extras header files, however I can't say definitively it will never happen.

zhimingwang36: >>>I assume that the migration code would populate ITERATOR_ADAPTOR_REQUIRED if necessary then? 
Yes.

tomflinda: @danhoeflinger  This PR needs to be rebased to the latest SYCLomatic branch.

danhoeflinger: Sorry I missed this.  I just rebased it.

tomflinda (approved):
LGTM

zhiweij1 (approved):

PR #1596, opened 2024-01-08
Added Support for emitting Migration Percentage for NVIDIA Library APIs in a given Application Binary

mpanoop:
Added Support for reporting the percentage of NVIDIA library APIs which is migratable by SYCLomatic before initiating the actual code migration @tovinkere @zhimingwang36


clang/tools/symbol-scanner/BinarySymbolScanner.py
```diff
@@ -54,6 +54,61 @@
 import sys
 import subprocess
 from pathlib import Path
+import pandas as p
+import numpy as np
+# Class which creates the dictionary of API Migration Support Status
+class api_mapping_support:
+    def __init__(self):
+        # Array of URLs for the API Mapping CSV files in SYCLomatic repo
+        self.api_mapping_urls=["https://raw.githubusercontent.com/oneapi-src/SYCLomatic/SYCLomatic/docs/dev_guide/api-mapping-status/CUB_API_migration_status.csv",
```
zhimingwang36, line 63: 1.  ASM_API_migration_status.csv and wmma_API_migration_status.csv  are not used.
2.  Have introduced "Partial" keywords in the *csv file, which means not all kind usage of the API is migration supported.

zhimingwang36 (approved):

PR #1609, opened 2024-01-10
[SYCLomatic] Fix lit case c_source_file_test/main.cu and redefine_sycl_type_a.c failed issue

intwanghao:


clang/lib/DPCT/SaveNewFiles.cpp
```diff
@@ -158,7 +158,7 @@ void rewriteFileName(clang::tooling::UnifiedPath &FileName) {
 
 void rewriteFileName(clang::tooling::UnifiedPath &FileName,
                      const clang::tooling::UnifiedPath &FullPathName) {
-  std::string FilePath;
+  std::string FilePath = FileName.getPath().str();
```
tomflinda, line 161:
```suggestion
  std::string FilePath;
```
As FilePath is used to output by rewriteFileName in Line162.

tomflinda: As the new add code shown in Line179, FileName is use to get fileinfo with DpctGlobalInfo::getInstance().insertFile, so it cannot be empty string.

tomflinda (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1617, opened 2024-01-12
[SYCLomatic] Adding asserts for unsupported functionality with DPCT_USM_LEVEL_NONE

danhoeflinger:
Adding asserts for unsupported functionality when `DPCT_USM_LEVEL_NONE` is true.


timmiesmith (approved):
LGTM

danhoeflinger:
It was determined that adding these asserts will cause more trouble than they will help clarify.  I will open another PR which adds comments describing this functionality as broken, but without asserts or changing the code.

PR #1630, opened 2024-01-16
[SYCLomatic] Fix incorrect source range of cub type migration

yihwang-nv:
- Fix incorrect source range of cub type migration. The `cub` namespace has not been removed when migrate cub type in dependent context.
- Fix crash issue. `DpctGlobalInfo::findAncestor` will return a local variable's address, which already dead when `findAncestor` return, it may cause SIGSEGV.


AndyCHHuang (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1636, opened 2024-01-17
[SYCLomatic] adding missing qualified name for dpct::transform_if

danhoeflinger:
Fully qualified name was missing from transform_if, which has caused an ambiguous function call with new API provided by oneDPL.  

We should update to use that oneDPL function, but for now, we should at least fix this ambiguity.


mmichel11 (approved):
LGTM

zhimingwang36 (approved):
LGTM

PR #1305, opened 2023-09-15
[SYCLomatic] Support migration for  cub::{StoreDirectBlocked, StoreDirectStriped} API

abhilash1910:
Add Migration of cub::store  API . Linked to #1819   
cc @yihanwg  , @zhimingwang36

PR #1640, opened 2024-01-18
[SYCLomatic] Block Load headers core

abhilash1910:
In Progress PR for Load/Store header functions for Block API (related later to #1305 )
cc @yihanwg @danhoeflinger @mmichel11


clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,51 @@ class radix_sort {
...
 
+/// Load linear segment items into block format across threads
+/// Helper for Block Load
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
+          typename InputIteratorT>
+__dpct_inline__ void load_blocked(int linear_tid, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_THREAD]) {
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+
```
yihwang-nv, line 572: Lets's remove extra blank line.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,51 @@ class radix_sort {
   uint8_t *_local_memory;
 };
 
+/// Load linear segment items into block format across threads
+/// Helper for Block Load
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
+          typename InputIteratorT>
+__dpct_inline__ void load_blocked(int linear_tid, InputIteratorT block_itr,
```
yihwang-nv: I would like to use `size_t` as the type of linear_id

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,51 @@ class radix_sort {
...
+}
+
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
+          typename InputIteratorT>
+__dpct_inline__ void load_striped(int linear_tid, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_THREAD]) {
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+
+    if((linear_tid + (ITEM * ITEMS_PER_THREAD) < GROUP_THREADS){
```
yihwang-nv: Need reformat and I think the if-stmt can be removed.

yihwang-nv (changes requested):

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,51 @@ class radix_sort {
...
+__dpct_inline__ void load_warp_striped(int linear_tid, InputIteratorT block_itr,
+                                       InputT (&items)[ITEMS_PER_THREAD]) {
+
+  int tid = linear_tid & 1;
+  int wid = linear_tid >> 1;
+  int warp_offset = wid * ITEMS_PER_THREAD;
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+
+    new (&items[ITEM]) InputT(block_itr[warp_offset + tid + (ITEM * 1)]);
```
yihanwg, line 608: Can we use `oneapi::dpl::uninitialized_value_construct`?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Added

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: removed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: The if-stmt is put to guard against GROUP_THREAD bounds (as a range) . reformatted.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,47 @@ class radix_sort {
...
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+    if (linear_tid + (ITEM * ITEMS_PER_THREAD) < GROUP_THREADS) {
+      items[ITEM] = block_itr[linear_tid + (ITEM * ITEMS_PER_THREAD)];
+    }
+  }
+}
+
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
+          typename InputIteratorT>
+__dpct_inline__ void load_subgroup_striped(size_t linear_tid,
```
mmichel11: If possible, I think this function should accept an `nd_item` as the first argument instead of `size_t`. We can't rely on a hardcoded subgroup size as SYCL allows variable sub-group sizes, so we should query it from the `nd_item` in this function along with the current work item's offset in the sub-group.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,47 @@ class radix_sort {
...
+                                           InputIteratorT block_itr,
+                                           InputT (&items)[ITEMS_PER_THREAD]) {
+
+  // int tid = linear_tid & 1;
```
mmichel11: I think it is best to reintroduce these variables so the line where we access memory is a bit more clear. As a general note, I think it is best to stick with SYCL terminology. In this case, thread should be work item and warp should be sub-group.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,47 @@ class radix_sort {
...
+  // int wid = linear_tid >> 5;
+  // int warp_offset = wid * ITEMS_PER_THREAD;
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+    new (&items[ITEM]) InputT(block_itr[&items[ITEM].get_sub_group().get_local_range()[0]] + linear_tid + (ITEM * ITEMS_PER_THREAD)]);
```
mmichel11: My understanding is that `InputT` can really be any type, so we can't query `get_sub_group` on it which is an `nd_item` function.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,47 @@ class radix_sort {
...
+}
+
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
+          typename InputIteratorT>
+__dpct_inline__ void load_striped(size_t linear_tid, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_THREAD]) {
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+    if (linear_tid + (ITEM * ITEMS_PER_THREAD) < GROUP_THREADS) {
+      items[ITEM] = block_itr[linear_tid + (ITEM * ITEMS_PER_THREAD)];
```
mmichel11: I think the accesses to `block_itr` should have a stride of the work-group size in a striped load pattern. To me this looks like this line should be
```
items[ITEM] = block_itr[linear_tid + (ITEM * GROUP_THREADS)];
```
Do you agree here?

mmichel11: In `load_subgroup_striped` we are assuming the memory is uninitialized and here we just perform an assignment. Should we have a consistent assumption about whether `items` is initialized / uninitialized throughout all of these functions?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -564,6 +564,47 @@ class radix_sort {
   uint8_t *_local_memory;
 };
 
+/// Load linear segment items into block format across threads
+/// Helper for Block Load
+template <int GROUP_THREADS, typename InputT, int ITEMS_PER_THREAD,
```
mmichel11: Instead of `GROUP_THREADS`, I think something like `work_group_size` or `GROUP_WORK_ITEMS` is best. Threads is not commonly used when referring to SYCL work items.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
yihwang-nv: Agree, we can add a template argument between `GROUP_WORK_ITEMS` and `InputT`
template <std::size_t GROUP_WORK_ITEMS, size_t ELEMENT_PRE_WORK_ITEM, typename ItemT, typename InputT, typename InputIteratorT>
__dpct_inline__ void load_subgroup_striped(const ItemT &item, .....) { ..... }

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes I agree, I will change this.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes changed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Let me think on this, because item per thread is similar to the work items present in a work group. So loading in striped indicates work items should get the stride. Let me know if I am mistaken.

abhilash1910: Yes I have similar doubt , could you suggest a better resolution on this?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: I have made changes in the template args.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes the reason I commented it out as currently it is not required for the memory allocation. 
Also I think that the previous logic might not be applicable for us as we directly set the offset (for warp) as the subgroup size.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: I think ITEM * GROUP_THREADS is correct if we donot use the if condition. Since here I am using a range block load it implies the same thing . If we remove the if condition, then the correct way is ITEM * GROUP_THREADS

mmichel11: Maybe I am misunderstanding the algorithm being implemented here but my understanding based on the name `load_striped` is that we load directly into items with a striped arrangement:

Say each thread is to load 4 elements (ITEMS_PER_THREAD=4) into it's local `items` array, the work-group size is 64, and `block_itr` consists of [0, 1, 2, 3, ... 255]. I would expect work-item 0 `items` to be `[0, 64, 128 192]`, work-item 1 to have `[1, 65, 129, 193]` and so on. This would require offsets of `linear_tid + (ITEM * GROUP_WORK_ITEMS)`.

mmichel11: A consistent assumption is probably best, but I do not know what the appropriate assumption is for this API so someone else will probably need to be asked.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,49 @@ class radix_sort {
   uint8_t *_local_memory;
 };
 
+/// Load linear segment items into block format across threads
+/// Helper for Block Load
+template <size_t GROUP_WORK_ITEMS, typename InputT, int ITEMS_PER_THREAD,
```
mmichel11: Similar to `GROUP_WORK_ITEMS`, I think `ITEMS_PER_THREAD` should be `ITEMS_PER_WORK_ITEM`, so we stick with SYCL terminology.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
mmichel11: I think the subgroup ID will likely be stored in a register either way. I would recommend to have the `item.get_sub_group().get_local_range()[0]` stored in some variable such as `subgroup_id` for readability, but I am okay with how it is currently if you would prefer that.

We should remove the commented code here.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Do we need a signed `int` for `ITEMS_PER_THREAD`?

danhoeflinger: I'd prefer we group all template value parameters together and all template type parameters together unless there is a compelling reason to do otherwise.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,49 @@ class radix_sort {
...
+          typename InputIteratorT>
+__dpct_inline__ void load_blocked(size_t linear_tid, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_THREAD]) {
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
```
danhoeflinger: You should take a step back and consider the naming conventions for variables, template types, template values here.  

It is somewhat a mess already within this file (before this PR), but I'd recommend the following to match the compatibility headers in general:
`PascalCase` with for template type parameters
`snake_case`  for local variables
I'm not sure if there is a strong convention already for template values like `GROUP_WORK_ITEMS`.  This file seems to use `ALL_CAPS_WITH_UNDERSCORES`, so I suppose that is OK.  

Whatever we use, lets be consistent, it is confusing to see this local variable `ITEM` in all caps.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: I believe @yihanwg  means that we may want to rethink if statement **inside** the inner loop to avoid overflow for `GROUP_WORK_ITEMS`.  We should be able to do some math outside the look and then have special casing for "full" and "partial" loads or just simply have an upper bound of the loop controlled by the starting math.  Checking this at every iteration seems very slow, and we should be able to know how many iterations this work_item needs to run with a single check or min / max operation.

We may need to consider thread divergence here, but I think even with that in mind this wont be the optimal way of writing this.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,49 @@ class radix_sort {
...
+                                           InputT (&items)[ITEMS_PER_THREAD]) {
+
+  // int tid = linear_tid & 1;
+  // int wid = linear_tid >> 5;
+  // int warp_offset = wid * ITEMS_PER_THREAD;
+#pragma unroll
+  for (int ITEM = 0; ITEM < ITEMS_PER_THREAD; ITEM++) {
+    new (&items[ITEM])
+        InputT(block_itr[item.get_sub_group().get_local_range()[0] +
+                         linear_tid + (ITEM * ITEMS_PER_THREAD)]);
```
mmichel11: A similar comment to above with regards to the stride. In this case, I believe instead of `ITEMS_PER_THREAD` it should be the subgroup size.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes this is updated

abhilash1910: Yes updated

abhilash1910: Yes added

danhoeflinger: I think you maybe misunderstood my comment.  I'm suggesting that we will only have positive numbers for `ITEMS_PER_THREAD`, and so we do not need `int` (which is signed) for it or for `item`, but rather we could use `uint32_t`, `uint16_t` or something like that which is unsigned.

abhilash1910: Yes it was an error on my part.

mmichel11:
Thinking about the design in more depth, I do not think these APIs should be implemented as free functions. Rather, I think they should be implemented as classes which overload the function call operator or have a member function that invokes the load operation.

My reasoning for this is that certain applications during migration may need to pass a "load" object to a separate function 
 / routine in a kernel. If they are implemented as classes, then an object can be passed to the function where it could then be used:
```
load_striped<...> obj;
my_func(obj);
```
However, if they are implemented as free functions, then they would need to be passed as function pointers:
```
auto func_obj = &load_striped<...>;
my_func(func_obj);
```
which is not supported in the SYCL programming model. To make this work, `load_striped` would have to be called directly from `my_func` which I think may be difficult from the migration perspective.

What I think we should do is have two classes `work_group_load` and `sub_group_load` and then have some enum type dictating the strategy (ex: striped, blocked, etc). We could then have have specializations of this class to implement each strategy and maybe have some helpers to reduce code duplication.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: I see that the if statements have been removed, but they have not been replaced by anything.  I assume that we still ultimately need to protect overrunning the range. Also, it is important for unrolling that the loop bounds are known at compile time.  Therefore, we may look at a sort of branched approach.

Here is some pseudocode:

``` 
if  (last element will overrun range)
{
    // "partial" case
    upper_bound = calculate last element which fits in range
    // no unrolling because upper_bound is runtime variable
    for (loop from start to upper_bound)
        operation //no if statement guard is required
}
else
{
    // "full case
    #pragma unroll
    for (loop from start to end)
        operation //no if statement guard is required
}
```

This allows us to still guard our range bounds, but not check at each iteration of the inner loop, and enjoy unrolling for the "full" case.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,58 @@ class radix_sort {
...
+  workgroup_load
+
+};
+
+template <size_t GROUP_WORK_ITEMS,
+          size_t ITEMS_PER_WORK_ITEM,
+          load_algorithm ALGORITHM,
+          typename InputT,
+          typename InputIteratorT>
+class load {
```
mmichel11: In my opinion, we should have a separate `work_group_load` and `sub_group_load` class. I think the differences between work group level and sub-group level algorithms are significant enough to logically separate in this manner.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,58 @@ class radix_sort {
   uint8_t *_local_memory;
 };
 
+/// Load linear segment items into block format across threads
+/// Helper for Block Load
+enum load_algorithm{
```
mmichel11: My thoughts for this enum were to use it to specify a `blocked`, `striped`, or any other load strategy implemented in the future rather than to choose between a sub-group or work group load.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,58 @@ class radix_sort {
...
+
+template <size_t GROUP_WORK_ITEMS,
+          size_t ITEMS_PER_WORK_ITEM,
+          load_algorithm ALGORITHM,
+          typename InputT,
+          typename InputIteratorT>
+class load {
+public:
+  template <typename Item>
+  __dpct_inline__ void load_blocked(size_t linear_tid, InputIteratorT block_itr,
```
mmichel11: If we decide to separate this into work group and sub group load classes and use the enum type to select the load strategy, then we can expose a single public `load` function and invoke the correct implementation based on the value of `ALGORITHM`.

We can consolidate the logic between striped and blocked strategies in private helper functions.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,58 @@ class radix_sort {
...
+};
+
+template <size_t GROUP_WORK_ITEMS,
+          size_t ITEMS_PER_WORK_ITEM,
+          load_algorithm ALGORITHM,
+          typename InputT,
+          typename InputIteratorT>
+class load {
+public:
+  template <typename Item>
```
mmichel11: `Item` is unused here. This template argument should be removed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: yes made necessary changes.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Should the load function be a  standalone function or should it be inside a class ? I have made some changes as per the review , could you please take a look?

abhilash1910: Also the striped enum has 2 sub functions (warp and without warp) so is there a better way to add into the same load method ( I am using another enum for warp/subgroup striped now )

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Removed where it is not needed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: I have made some changes following this logic. Could you please take a look? Thanks

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: yes removed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,116 @@ class radix_sort {
...
+};
+
+ __dpct_inline__ void load(size_t GROUP_WORK_ITEMS,
+                           size_t ITEMS_PER_WORK_ITEM,
+                           load_algorithm ALGORITHM,
+                           InputT (&items)[ITEMS_PER_WORK_ITEM],
+                           InputIteratorT block_itr,
+                           Item &item){
+
+   if (ALGORITHM == blocked){
```
danhoeflinger: These `if`, `else if` will need to be `constexpr` to avoid a runtime branch for each load call.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,116 @@ class radix_sort {
...
+          typename InputIteratorT>
+class workgroup_load {
+public:
+  
+  __dpct_inline__ void load_blocked(size_t linear_tid, InputIteratorT block_itr,
+                                    InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+  
+  uint32_t workgroup_offset = linear_tid * ITEMS_PER_WORK_ITEM;
+  if (workgroup_offset > GROUP_WORK_ITEMS){
+      uint32_t final_idx = (GROUP_WORK_ITEMS - workgroup_offset);
```
danhoeflinger: `workgroup_offset` seems to be the base offset for the work item. You check if it is larger than `GROUP_WORK_ITEMS`, and if it is you subtract `GROUP_WORK_ITEMS - workgroup_offset`.  It seems like this can only be a negative number, but then you assign it to an unsigned int.

Please take another look at this, I think its wrong in a few places.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,116 @@ class radix_sort {
...
+/// Helper for Block Load
+enum load_algorithm{
+
+  blocked,
+  striped,
+  warp_striped
+
+};
+
+template <size_t GROUP_WORK_ITEMS,
```
danhoeflinger: What is `GROUP_WORK_ITEMS` meant to represent?
It is named as if it is the size of a sycl work group or subgroup (number of work items in a group), but it seems to be used as the total number of elements to be loaded by the load operation.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,116 @@ class radix_sort {
...
+          load_algorithm ALGORITHM,
+          typename InputT,
+          typename InputIteratorT,
+          typename Item>
+class subgroup_load {
+public:
+  __dpct_inline__ void load_striped(size_t linear_tid, InputIteratorT block_itr,
+                                    InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+    if (linear_tid + ((ITEMS_PER_WORK_ITEM - 1)*ITEMS_PER_WORK_ITEM) > GROUP_WORK_ITEMS){
```
danhoeflinger: If `GROUP_WORK_ITEMS` is the total elements to load, and we are trying to determine if we will overrun that range, I believe this needs to be `>=` rather than `>`.  (same in other places)

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,116 @@ class radix_sort {
...
+    else{
+    #pragma unroll
+      for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+        items[idx] = block_itr[linear_tid + (idx  * GROUP_WORK_ITEMS)];
+      }
+    }
+  }
+
+  
+  __dpct_inline__ void load_subgroup_striped(const Item &item, size_t linear_tid,
```
danhoeflinger: Can you explain the difference between `load_striped` and `load_subgroup_striped`?  I'm not following how these are meant to be used.

This one does a placement `new` operator, and incorporates the subgroup offset of a work_item into the indexing...

An overall description of each APIs semantically in the PR would help me to review this.

danhoeflinger (changes requested):
Do you have a matching test branch / PR for this that you are working with?
If so, please ensure that it has appropriate coverage and post it.
It would help to understand how this is meant to be used and to ensure correctness.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes I think for this  ```load_blocked``` function, there is no mandatory need for checking bounds . The ```GROUP_WORK_ITEMS```  indicates a valid range up to which the number of items to be loaded.

abhilash1910:
> Do you have a matching test branch / PR for this that you are working with? If so, please ensure that it has appropriate coverage and post it. It would help to understand how this is meant to be used and to ensure correctness.

Yes it is in PR https://github.com/oneapi-src/SYCLomatic-test/pull/619 . This needs to be updated with the changes in this PR.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes I agree

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,89 @@ class radix_sort {
...
+
+  // This implementation does not take in account range loading across
+  // workgroup items To-do: Decide whether range loading is required for group
+  // loading
+
+  size_t subgroup_offset = item.get_sub_group().get_local_range()[0];
+#pragma unroll
+  for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    new (&items[idx]) InputT(
+        block_itr[subgroup_offset + linear_tid + (idx * GROUP_WORK_ITEMS)]);
```
mmichel11: With my understanding that this is a work-group level algorithm, I believe the stride here should be the subgroup size instead of `GROUP_WORK_ITEMS`. 

I think we should also compute the initial offset for a subgroup to start as `(sg_id * ITERS_PER_WORK_ITEM * sg_sz + subgroup_offset)` where `sg_sz` is the number of elements in the subgroup and `sg_id` is the subgroup's index in the workgroup (can be queried with `item.get_sub_group().get_global_range()`). If we call this `start`, then I believe the offset within your loop would be computed as `start + idx * sg_sz`.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,89 @@ class radix_sort {
...
+    // loading
+
+    uint32_t workgroup_offset = linear_tid * ITEMS_PER_WORK_ITEM;
+#pragma unroll
+    for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+      items[idx] = block_itr[workgroup_offset + idx];
+    }
+  }
+  // loads a linear segment of workgroup items into a striped arrangement.
+  __dpct_inline__ void load_striped(size_t linear_tid, InputIteratorT block_itr,
```
mmichel11: Do we want to expose this publicly? If we do, it may be worth pulling it out as a a free function, so it can be invoked directly from here as well. It could still be called similarly within `load`.

If not, we should make it private. The same goes for `load_blocked`.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,89 @@ class radix_sort {
...
+  BLOCK_LOAD_STRIPED,
+  // To-do: BLOCK_LOAD_WARP_TRANSPOSE
+
+};
+
+// loads a linear segment of workgroup items into a subgroup striped
+// arrangement. Created as free function until exchange mechanism is
+// implemented.
+// To-do: inline this function with BLOCK_LOAD_WARP_TRANSPOSE mechanism
+template <size_t GROUP_WORK_ITEMS, size_t ITEMS_PER_WORK_ITEM, typename InputT,
```
mmichel11: Do we need `GROUP_WORK_ITEMS` as a template parameter or can we just query it from the work item?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Free function makes sense, can be used both ways.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,89 @@ class radix_sort {
...
+  workgroup_load(uint8_t *local_memory) : _local_memory(local_memory) {}
+
+  __dpct_inline__ void load(const Item &item, size_t linear_tid,
+                            InputIteratorT block_itr,
+                            InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+    if constexpr (ALGORITHM == BLOCK_LOAD_DIRECT) {
+      load_blocked(linear_tid, block_itr, (&items)[ITEMS_PER_WORK_ITEM]);
+    }
+    if constexpr (ALGORITHM == BLOCK_LOAD_STRIPED) {
```
danhoeflinger: Should be `else if`

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: I agree that free functions make the most sense.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,89 @@ class radix_sort {
...
+                      InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+  // This implementation does not take in account range loading across
+  // workgroup items To-do: Decide whether range loading is required for group
+  // loading
+
+  size_t subgroup_offset = item.get_sub_group().get_local_range()[0];
+#pragma unroll
+  for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    new (&items[idx]) InputT(
```
danhoeflinger: I'm not sure if this was answered before, but why does this function differ in semantics to create items with placement `new`, where the others just assign?
Is this the desired effect?  Will load_subgroup_striped always be loading into uninitialized memory where the others are not?
If so, do we need to adjust naming / comments to make this clear?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes changed

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: Yes updated.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,98 @@ class radix_sort {
...
+load_subgroup_striped(const Item &item, size_t linear_tid,
+                      InputIteratorT block_itr,
+                      InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+  // This implementation does not take in account range loading across
+  // workgroup items To-do: Decide whether range loading is required for group
+  // loading
+  // This implementation uses unintialized memory for loading linear segments
+  // into warp striped arrangement.
+  size_t subgroup_offset = item.get_sub_group().get_local_range()[0];
```
mmichel11: I had phrased my initial comment here poorly. This is what I believe the variables should be equal to:

```
auto subgroup_offset = item.get_sub_group().get_local_linear_id();
auto subgroup_size = item.get_sub_group().get_local_linear_range();
auto subgroup_idx = item.get_sub_group().get_group_linear_id();
```
The APIs also return `uint32_t`, so we should use this or just `auto` instead of converting to `size_t`.

mmichel11: Bumping this comment for visibility

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,98 @@ class radix_sort {
...
+}
+
+// loads a linear segment of workgroup items into a subgroup striped
+// arrangement. Created as free function until exchange mechanism is
+// implemented.
+// To-do: inline this function with BLOCK_LOAD_WARP_TRANSPOSE mechanism
+template <size_t ITEMS_PER_WORK_ITEM, typename InputT, typename InputIteratorT,
+          typename Item>
+__dpct_inline__ void
+uninitialized_load_subgroup_striped(const Item &item, size_t linear_tid,
```
mmichel11: We do not use `linear_tid`, so I think it can be removed. `item` is taking its place here.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,98 @@ class radix_sort {
...
+#pragma unroll
+  for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    items[idx] = block_itr[workgroup_offset + idx];
+  }
+}
+
+// loads a linear segment of workgroup items into a striped arrangement.
+template <size_t ITEMS_PER_WORK_ITEM, load_algorithm ALGORITHM, typename InputT,
+          typename InputIteratorT, typename Item>
+__dpct_inline__ void load_striped(const Item &item, size_t linear_tid,
```
mmichel11: I think we can just accept the `Item` parameter and query the work-group ID from there.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,98 @@ class radix_sort {
...
+  BLOCK_LOAD_DIRECT,
+  BLOCK_LOAD_STRIPED,
+  // To-do: BLOCK_LOAD_WARP_TRANSPOSE
+
+};
+
+// loads a linear segment of workgroup items into a blocked arrangement.
+template <size_t ITEMS_PER_WORK_ITEM, load_algorithm ALGORITHM, typename InputT,
+          typename InputIteratorT>
+__dpct_inline__ void load_blocked(size_t linear_tid, InputIteratorT block_itr,
```
mmichel11: I think this signature is confusing since in other functions we are accepting an `Item` parameter as well.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: Bumping this, as I think it still needs addressing for `GROUP_WORK_ITEMS` in `load_striped`.

mmichel11 (commented):
For a consistent API, I think all of these additions should just accept an `Item` parameter and query the workgroup ID internally instead of taking the `size_t linear_tid` parameter. 

In my opinion, it is a bit confusing when some of the functions accept `size_t linear_tid` and others accept both an `Item` and `size_t` parameter. Unless there are any performance reasons not to, I think accepting the `nd_item` as the first argument is the clearest.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
abhilash1910: addressed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
mmichel11: I also suggested changes of the SYCL APIs being called for `subgroup_offset` and `subgroup_idx`. Can you verify this?

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,98 @@ class radix_sort {
...
+
+    if constexpr (ALGORITHM == BLOCK_LOAD_DIRECT) {
+      load_blocked(item, block_itr, (&items)[ITEMS_PER_WORK_ITEM]);
+    } else if constexpr (ALGORITHM == BLOCK_LOAD_STRIPED) {
+      load_striped(item, block_itr, (&items)[ITEMS_PER_WORK_ITEM]);
+    }
+  }
+
+private:
+  uint8_t *_local_memory;
```
danhoeflinger, line 663: `_local_memory` is initialized in this class constructor, but then never used...

I assume it is meant to be used as the per-work-item local memory for the load calls, but it is not as the PR is currently written.

danhoeflinger (changes requested):

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
danhoeflinger: After closer inspection, I think that it is meant to be there as potential temporary memory for use in the load / rearrange operation.  In this case, my suggestion would be to return `0` from `get_local_memory_size()`, as no temporary data is used in the load algorithms so far, but keep the `_local_memory` variable / ctor.  This allows us the most freedom for the API going forward to support such temporary memory usage.

The other question is if we want to consider accepting temporary storage in the "free functions" (even if we don't use it right now).  I'm not sure the best answer there right now.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,93 @@ class radix_sort {
...
+#pragma unroll
+  for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    new (&items[idx]) InputT(block_itr[initial_offset + (idx * subgroup_size)]);
+  }
+}
+
+template <size_t ITEMS_PER_WORK_ITEM, load_algorithm ALGORITHM, typename InputT,
+          typename InputIteratorT, typename Item, typename T>
+class workgroup_load {
+  static size_t get_local_memory_size(size_t group_threads) {
```
danhoeflinger: Should use `group_work_items` rather than `group_threads` i think.

abhilash1910: yes will update , thanks

danhoeflinger (commented):
All my feedback is "resolved" and marked as much except for one remaining issue with naming conventions.  Please look above.  

Please check other unresolved feedback to see if anything else it outstanding from other reviewers.  

Also, we want to relocate this to the new header rather than place it in `dpl_extras/dpcpp_extensions.h`. So we are blocked by #1784

abhilash1910:
Yes made necessary changes, for now if looks good, we can go ahead with this PR , and I will add to #1784 following other apis. 
PR test https://github.com/oneapi-src/SYCLomatic-test/pull/619 needs some discussion and checks.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,96 @@ class radix_sort {
...
+__dpct_inline__ void load_striped(const Item &item, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+  // This implementation does not take in account range loading across
+  // workgroup items To-do: Decide whether range loading is required for group
+  // loading
+  size_t linear_tid = item.get_local_linear_id();
+  size_t group_work_items = item.get_global_range();
+#pragma unroll
+  for (uint32_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
```
yihwang-nv: Can type `size_t` be used here like `ITEMS_PER_WORK_ITEM`?

yihwang-nv (approved):
LGTM

mmichel11:
This PR looks to be in good shape for me. However, I think the outstanding comments in https://github.com/oneapi-src/SYCLomatic-test/pull/619 need to be addressed, so we can have confidence in the implementation's correctness.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,97 @@ class radix_sort {
...
+  BLOCK_LOAD_STRIPED,
+  // To-do: BLOCK_LOAD_WARP_TRANSPOSE
+
+};
+
+// loads a linear segment of workgroup items into a blocked arrangement.
+template <size_t ITEMS_PER_WORK_ITEM, load_algorithm ALGORITHM, typename InputT,
+          typename InputIteratorT, typename Item>
+__dpct_inline__ void load_blocked(const Item &item, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_WORK_ITEM]) {
```
danhoeflinger: Why are we accepting `ALGORITHM` as a template parameter, it is not used?  Isn't the algorithm "blocked" / "direct" which is indicated by the function name and the branch before calling it?
Same question below for `load_striped`.

abhilash1910: yes this is ALGORITHM parameter can be removed .

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,99 @@ class radix_sort {
...
+template <size_t ITEMS_PER_WORK_ITEM, typename InputT, typename InputIteratorT,
+          typename Item>
+__dpct_inline__ void load_striped(const Item &item, InputIteratorT block_itr,
+                                  InputT (&items)[ITEMS_PER_WORK_ITEM]) {
+
+  // This implementation does not take in account range loading across
+  // workgroup items To-do: Decide whether range loading is required for group
+  // loading
+  size_t linear_tid = item.get_local_linear_id();
+  size_t group_work_items = item.get_global_range().size();
```
mmichel11: Should this be `get_local_range()`? That returns the work-group size which is what I think we want for our stride.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,99 @@ class radix_sort {
...
+  uint32_t initial_offset =
+      (subgroup_idx * ITEMS_PER_WORK_ITEM * subgroup_size) + subgroup_offset;
+#pragma unroll
+  for (size_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    new (&items[idx]) InputT(block_itr[initial_offset + (idx * subgroup_size)]);
+  }
+}
+
+template <size_t ITEMS_PER_WORK_ITEM, load_algorithm ALGORITHM, typename InputT,
+          typename InputIteratorT, typename Item>
```
zhimingwang36, line 646: here are quite many item usage, ITEM/Item/iterms/ITERMS_PER_WORK_ITEM,  maybe it should give a better name for "typename Item", it is a little confusing.  or please add comments.

abhilash1910: thanks will add comments to clarify.

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,106 @@ class radix_sort {
...
+  uint32_t subgroup_idx = item.get_sub_group().get_group_linear_id();
+  uint32_t initial_offset =
+      (subgroup_idx * ITEMS_PER_WORK_ITEM * subgroup_size) + subgroup_offset;
+#pragma unroll
+  for (size_t idx = 0; idx < ITEMS_PER_WORK_ITEM; idx++) {
+    new (&items[idx]) InputT(block_itr[initial_offset + (idx * subgroup_size)]);
+  }
+}
+// template parameters :
+// ITEMS_PER_WORK_ITEM: size_t varaiable controlling the number of items per
```
zhimingwang36:
```suggestion
// ITEMS_PER_WORK_ITEM: size_t variable controlling the number of items per
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,106 @@ class radix_sort {
...
+// thread/work_item
+// ALGORITHM: load_algorithm variable controlling the type of load operation.
+// InputT: typename parameter controlled at runtime from input sequence.
```
zhimingwang36:
```suggestion
// InputT: type for input sequence.
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,106 @@ class radix_sort {
...
+// InputIteratorT: typename parameter for iterator pointer controlled at
```
zhimingwang36:
```suggestion
// InputIteratorT:  input iterator type 
```

clang/runtime/dpct-rt/include/dpct/dpl_extras/dpcpp_extensions.h
```diff
@@ -566,6 +566,106 @@ class radix_sort {
...
+// runtime.
+// Item : typename parameter resembling sycl::nd_item<3> .
```
zhimingwang36, line 644: can we rename to nd_item? (what is the diff b/w Item and sycl::nd_item)

abhilash1910: Yes it is same thing, but to maintain the same variable name across the file (similar for other apis), Item is used.

abhilash1910:
This PR is now tested with https://github.com/oneapi-src/SYCLomatic-test/pull/619 and all tests are passing. 
Tested on application side as well, working as expected.

danhoeflinger (approved):
LGTM, 
There are some changes needed in the testing PR still, but with some local changes to those tests and passing runs, I'm confident enough in this PR now to approve.

mmichel11 (approved):
LGTM

PR #1691, opened 2024-01-31
[SYCLomatic] Check option conflict for analysis mode

ziranzha:


clang/lib/DPCT/Error.h
```diff
@@ -64,6 +64,7 @@ enum ProcessStatus {
   MigarteBuildScriptIncorrectUse = -50,
   MigarteBuildScriptAndMigarteBuildScriptOnlyBothUse = -51,
   MigrationErrorNoExplicitInRootAndCMakeScript = -52,
+  MigrationErrorOptionConflict = -53,
```
tangjj11: There is already an option `MigrationErrorConflictOptions`. Can we use it directly? now it only use in query-api-mapping.

zhimingwang36 (approved):
LGTM

zhimingwang36:
need rebase the PR.

PR #1696, opened 2024-02-02
[SYCLomatic][Feature] Added categorized help support

TejaX-Alaghari:
Added categorical help support for SYCLomatic command line options

Example 1:
`dpct --help`

image

Example 2:
`dpct --help=advanced`

image


llvm/lib/Support/CommandLine.cpp
```diff
@@ -2351,6 +2351,20 @@ sortSubCommands(const SmallPtrSetImpl<SubCommand *> &SubMap,
 
 namespace {
 
+/// HelpCategory defines various category groups for dpct options
+enum class HelpCategory {
```
zhimingwang36, line 2358: this file is not only for DPCT, so guard the code change with macro: SYCLomatic_CUSTOMIZATION

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2462,6 +2476,7 @@ class HelpPrinter {
 
 class CategorizedHelpPrinter : public HelpPrinter {
 public:
+  HelpCategory helpCatEnum = HelpCategory::HC_All;
```
zhimingwang36: Same as above.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2473,6 +2488,46 @@ class CategorizedHelpPrinter : public HelpPrinter {
     return (*A)->getName().compare((*B)->getName());
   }
 
+  bool isSameCategory(OptionCategory *lhs, OptionCategory *rhs) {
```
zhimingwang36: same as above.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2473,6 +2488,46 @@ class CategorizedHelpPrinter : public HelpPrinter {
...
+    return (lhs->getName() == rhs->getName());
+  }
+
+  // Return the user requested category
+  OptionCategory &getReqCategory(HelpCategory reqCatEnumVal) {
+    if (reqCatEnumVal == HelpCategory::HC_All) {
```
zhimingwang36: pre to use switch case to make the code readable.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2657,11 +2716,61 @@ struct CommandLineCommonOptions {
...
+                                   "List options to manage warnings "
+                                   "generated by the tool",
+                                   false},
+              cl::OptionEnumValue{"help-info",
+                                   int(HelpCategory::HC_HelpInfo),
+                                   "List options to display tool information",
+                                   false},
+              cl::OptionEnumValue{"intercept-build",
+                                   int(HelpCategory::HC_InterceptBuild),
+                                   "Lists options of intercept-build tool",
```
zhimingwang36:
```suggestion
                                   "List options to generate a compilation database",
```

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2657,11 +2716,61 @@ struct CommandLineCommonOptions {
...
+                                   "how code is migrated",
+                                   false},
+              cl::OptionEnumValue{"report-gen",
+                                   int(HelpCategory::HC_ReportGen),
+                                   "List options to control report generation "
+                                   "during migration",
+                                   false},
+              cl::OptionEnumValue{"build-script",
+                                   int(HelpCategory::HC_BuildScript),
+                                   "List options to migrate build scripts",
```
zhimingwang36:
```suggestion
                                   "List options to migrate build script(s)",
```

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2657,11 +2716,61 @@ struct CommandLineCommonOptions {
...
+                                   int(HelpCategory::HC_All),
+                                   "List all options in alphabetical order",
+                                   true},
+              cl::OptionEnumValue{"basic",
+                                   int(HelpCategory::HC_Basic),
+                                   "List options for basic migration",
+                                   false},
+              cl::OptionEnumValue{"advanced",
+                                   int(HelpCategory::HC_Advanced),
+                                   "List options to control migrated source, "
```
zhimingwang36:
```suggestion
                                   "List option(s) to control migrated source, "
```

zhimingwang36: same for others.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2784,6 +2930,12 @@ void HelpPrinterWrapper::operator=(bool Value) {
     UncategorizedPrinter = true; // Invoke uncategorized printer
 }
 
+void HelpPrinterWrapper::operator=(HelpCategory Value) {
```
zhimingwang36, line 2953: guard the code with SYCLomatic_CUSTOMIZATION

zhimingwang36:
Aslo:  1. when run $dpct -help advance  ,  don't show the "Examples" section.  2.  how about add a new category for "Examples" section.   3.  We need more content for "Examples" section.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2503,13 +2558,16 @@ class CategorizedHelpPrinter : public HelpPrinter {
...
     }
 
+    OptionCategory &reqCat(getReqCategory(helpCatEnum));
+
     // Now do printing.
     for (OptionCategory *Category : SortedCategories) {
       // Hide empty categories for --help, but show for --help-hidden.
       const auto &CategoryOptions = CategorizedOptions[Category];
       bool IsEmptyCategory = CategoryOptions.empty();
+      bool IsReqCat = isSameCategory(&reqCat, Category);
```
ziranzha: Only comparing the pointer may be enough here.

tomflinda:
Suggest adding lit test for each output of categorized value, so that new option added or old option changed for some categorized value in the future, they can be tracked.
For checking the output of of categorized value, like output of 'dpct --help=advanced', you can follow the way of cmake-syntax lit test.

tomflinda (commented):
Suggest adding lit test for each output of categorized value, so that new option added or old option changed for some categorized value in the future, they can be tracked.
For checking the output of of categorized value, like output of 'dpct --help=advanced', you can follow the way of cmake-syntax lit test.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2403,6 +2421,87 @@ class HelpPrinter {
     StrSubCommandPairVector Subs;
     sortSubCommands(GlobalParser->RegisteredSubCommands, Subs);
 
+    const char *DPCTExamplesMsg =
```
ziranzha, line 2424: The logic should be same to output examples for enum::example and enum::all

ziranzha (approved):

clang/lib/DPCT/DPCT.cpp
```diff
@@ -96,18 +108,55 @@ const char *const CtHelpMessage =
...
-    "Migrate single source file with C++11 features:\n\n"
-    "  dpct --extra-arg=\"-std=c++11\" source.cpp\n\n"
-    "Migrate all files available in compilation database:\n\n"
-    "  dpct --compilation-database=<path to location of compilation database file>\n\n"
-    "Migrate one file in compilation database:\n\n"
-    "  dpct --compilation-database=<path to location of compilation database file>  source.cpp\n\n"
-#if defined(_WIN32)
-    "Migrate all files available in vcxprojfile:\n\n"
-    "  dpct --vcxprojfile=path/to/vcxprojfile.vcxproj\n"
+    "Prepare for Migration\n"
```
zhimingwang36: as the string is quite long now, suggest put into a single "migration_cmd_examples.inc" file, and include it.  also easy to maintain.  Aslo maybe consider removing the formats.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2403,6 +2421,87 @@ class HelpPrinter {
...
+        "Example commands:\n"
+#ifndef _WIN32
+        "       $ dpct --intercept-build make\n"
+#endif
+        "       $ dpct --intercept-build --parse-build-log <logfile>\n\n"
+        "Estimate the migration effect to complete a migration\n"
+        "Example commands:\n"
+        "       $ dpct --analysis-mode source.cu "
+        "--analysis-mode-output-file\n\n"
+        "Migrate your code\n"
```
zhimingwang36: Same as above.  Let's isolate the fixed help string into single inc file.

clang/lib/DPCT/DPCT.cpp
```diff
@@ -86,28 +86,30 @@ extern UnifiedPath VcxprojFilePath;
...
+llvm::cl::OptionCategory &DPCTWarningsCat = llvm::cl::getDPCTWarningsCategory();
+llvm::cl::OptionCategory &DPCTHelpInfoCat = llvm::cl::getDPCTHelpInfoCategory();
+llvm::cl::OptionCategory &DPCTInterceptBuildCat =
+    llvm::cl::getDPCTInterceptBuildCategory();
 void initWarningIDs();
 } // namespace dpct
 } // namespace clang
 
 // clang-format off
+// #include "clang/DPCT/migration_cmd_examples.inc"
```
zhimingwang36: remove the dead Line 106.

llvm/include/llvm/migration_cmd_examples.inc
```diff
@@ -0,0 +1,95 @@
+#ifdef _WIN32
```
zhimingwang36: suggest to make is simple, like const char *longString = R""""(
This is 
a very 
long 
string
)"""";

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2353,9 +2353,31 @@ sortSubCommands(const SmallPtrSetImpl<SubCommand *> &SubMap,
...
+  HC_BuildScript,
+  HC_QueryAPI,
+  HC_Warnings,
+  HC_HelpInfo,
+  HC_InterceptBuild,
+  HC_Examples
+};
+#endif // SYCLomatic_CUSTOMIZATION
+
+// clang-format off
```
zhimingwang36: L2374 should be protected by SYCLomatic_CUSTOMIZATION. 
L2373 "clang-format off " should be removed.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2353,9 +2353,31 @@ sortSubCommands(const SmallPtrSetImpl<SubCommand *> &SubMap,
...
+#include "llvm/migration_cmd_examples.inc"
+// clang-format on
+
 class HelpPrinter {
 protected:
   const bool ShowHidden;
+  HelpCategory helpCatEnum = HelpCategory::HC_All;
```
zhimingwang36, line 2379: for all SYCLomaitc specific code should be protected by "SYCLomatic_CUSTOMIZATION"

llvm/include/llvm/migration_cmd_examples.inc
```diff
@@ -0,0 +1,123 @@
...
+const char *const DPCTExamplesMsg = R""""(
```
zhimingwang36: what is the diff between two DPCTExamplesMsg? Avoid modifying 2 for each update. How about we String A  + String B + String C ;  (maybe String B is specific for windows).

TejaX-Alaghari: How about I maintain a single str with "#ifdef WIN32"?

Will that be okay?

zhimingwang36 (approved):
LGTM

PR #1709, opened 2024-02-07
[SYCLomatic] Fix coverity issue 0207

AndyCHHuang:


wenhuiNi (approved):

clang/lib/DPCT/DPCT.cpp
```diff
@@ -665,7 +665,7 @@ int runDPCT(int argc, const char **argv) {
         "suppress-warnings",
         "suppress-warnings-all",
     };
-    for (auto Opt : IgnoreOpts) {
+    for (auto &Opt : IgnoreOpts) {
```
wenhuiNi: I think we can add const for Opt,
```suggestion
    for (const auto &Opt : IgnoreOpts) {
```

ziranzha (approved):

PR #1783, opened 2024-03-13
[NFC][SYCLomatic] Use bitfield to reduce class CtTypeInfo  size from 200 bytes to 184 bytes

yihwang-nv:


clang/lib/DPCT/AnalysisInfo.h
```diff
@@ -1675,20 +1676,20 @@ class CtTypeInfo {
   void removeQualifier() { BaseName = BaseNameWithoutQualifiers; }
 
 private:
+  unsigned PointerLevel : 16;
+  unsigned IsReference : 1;
+  unsigned IsTemplate : 1;
+  unsigned TemplateDependentMacro : 1;
+  unsigned IsArray : 1;
+  unsigned IsPseudo : 1;
```
tangjj11: declare but not use?

tangjj11 (changes requested):

clang/lib/DPCT/AnalysisInfo.h
yihwang-nv: removed

tangjj11 (approved):

zhimingwang36 (approved):
LGTM

PR #1790, opened 2024-03-15
[DOC] Add last updated date

krisdale:
Add last updated date to footer and to top of api mapping status page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,8 +16,7 @@ sys.path.insert(1, '_static/nodes')
 # sys.path.insert(0, os.path.abspath('.'))
 
 from docutils import nodes
-
-
+from datetime import date
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -43,6 +42,8 @@
 
 # -- Options for HTML output -------------------------------------------------
 
+today = str(date.today().strftime('%d %B, %Y'))
```
zhimingwang36, line 45: should we pre-defined today to be the last update date for the *CSV file? If the value is not available, then try to use it today (runtime date for this Python script).

krisdale: The date that is currently in place is intended to be the last updated date of the published document (since the csv files in the repo may have been updated since then). Especially the date in the footer, the context is 'this page last updated on ...'. The footer date and the date on the mapping page should match, in order to avoid confusion. Glad to discuss further if needed!

zhimingwang36 (approved):
LGTM

PR #1792, opened 2024-03-18
[SYCLomatic] Fix pulldown failure

the-slow-one:
Resolve merge conflict due to refactored code.


the-slow-one:
Added uninteded workflow file. Replacing the PR with a new one.

PR #1794, opened 2024-03-18
[SYCLomatic] Fix pulldown fail 20240311

the-slow-one:
The merge conflict was due to refactored code.


clang/lib/Frontend/CompilerInstance.cpp
```diff
@@ -1067,30 +1067,7 @@ bool CompilerInstance::ExecuteAction(FrontendAction &Act) {
 // don't show the message "3 errors generated during parsing".
 // it's duplicate with dpct summary in command line msg.
 #else
-  if (getDiagnosticOpts().ShowCarets) {
```
the-slow-one, line 1070: Here is the change!

the-slow-one:
Pulldown fail is fixed here

https://github.com/oneapi-src/SYCLomatic/pull/1800

PR #1853, opened 2024-04-01
[SYCLomatic] Fix for ambiguous operations in device_vector

danhoeflinger:
After conversion operator was added to a normal reference #1538, many provided operators became ambiguous with the operators of the converted type.
This removes those ambiguous operators, in favor of converting to the base reference type and using the operators there.  

We can convert back to back to `device_reference<T>` from a `T&` already.


zhimingwang36 (approved):
LGTM

tomflinda (approved):
LGTM

mmichel11 (approved):
LGTM, thanks for the fix. I experimented with a few different operators as well and all behave as expected.

ziranzha (approved):

PR #1861, opened 2024-04-03
[SYCLomatic] return device_pointer from data()

danhoeflinger:
Fix for data() function return and `pointer` typedef for `device_vector` to use `device_pointer`

I don't know the motivation for originally using `T*` rather than `device_pointer<T>`.  This passes the tests for vector.  It seems that this should be a reasonable change.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
@@ -159,8 +159,8 @@ public:
   using reference = device_reference<T>;
   using const_reference = const reference;
   using value_type = T;
-  using pointer = T *;
-  using const_pointer = const T *;
+  using pointer = device_pointer<T>;
+  using const_pointer = const device_pointer<T>;
   using difference_type =
       typename ::std::iterator_traits<iterator>::difference_type;
   using size_type = ::std::size_t;
@@ -171,7 +171,7 @@ private:
   Allocator _alloc;
   size_type _size;
   size_type _capacity;
-  pointer _storage;
+  T* _storage;
 
   size_type _min_capacity() const { return size_type(1); }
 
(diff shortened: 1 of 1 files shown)


clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
```diff
@@ -462,8 +462,8 @@ class device_vector {
   reference front() { return *begin(); }
   const_reference back(void) const { return *(end() - 1); }
   reference back(void) { return *(end() - 1); }
-  pointer data(void) { return _storage; }
-  const_pointer data(void) const { return _storage; }
+  pointer data(void) { return pointer(_storage); }
+  const_pointer data(void) const { return pointer(_storage); }
```
timmiesmith:
```suggestion
  const_pointer data(void) const { return const_pointer(_storage); }
```

timmiesmith (changes requested):
A small change is needed. The rest looks good to me.

clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
danhoeflinger: done.  Thanks

timmiesmith (approved):
LGTM. Thank you.

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpct/dpl_extras/vector.h
```diff
@@ -159,8 +159,8 @@ class device_vector {
   using reference = device_reference<T>;
   using const_reference = const reference;
   using value_type = T;
-  using pointer = T *;
-  using const_pointer = const T *;
+  using pointer = device_pointer<T>;
+  using const_pointer = const device_pointer<T>;
```
ziranzha: Should `const_pointer` be defined as `device_pointer<const T>`?

danhoeflinger: Good catch.  Fixed, thanks!

PR #1900, opened 2024-04-22
[SYCLomatic] Update format for clang version

the-slow-one:
Change format of clang version to be in line with dpct version.


zhimingwang36 (approved):
LGTM

PR #1913, opened 2024-04-24
[SYCLomatic] Removed example string for --help=<arg> option and added autocompletion support

TejaX-Alaghari:
This PR adds following functionality to **dpct --help** option

- Removed default example string for --help=<arg> option and 
- Added autocompletion support for --help args


zhimingwang36 (approved):
LGTM

PR #1934, opened 2024-04-29
[SYCLomatic] Fix the issue that variables that has same name not merged into one variable

intwanghao:


tomflinda (commented):
Advise to add a lit test to verify this fixing so that the regression could be caught in time.

zhiweij1 (approved):

ziranzha (approved):

tomflinda (commented):
LGTM

PR #1951, opened 2024-05-08
[SYCLomatic] Relax the syntax restriction of asm builtin identifier

yihwang-nv:


clang/lib/DPCT/Asm/AsmLexer.cpp
```diff
@@ -196,6 +196,11 @@ bool InlineAsmLexer::lexTokenInternal(InlineAsmToken &Result) {
       ++CurPtr;
       Kind = asmtok::pipe;
     } else {
+      // PTX allows % builtin-identifier.
+      if (isIdentifierContinue(Char)) {
```
tangjj11: Should better merge this line to the pre line? `else if (isIdentifierContinue(Char))`

tangjj11 (approved):

clang/lib/DPCT/Asm/AsmLexer.cpp
yihanwg: they are on different paths

yihwang-nv: done

zhimingwang36 (approved):

PR #1996, opened 2024-05-16
[SYCLomatic] Use oneDPL provided sort_by_key 

danhoeflinger:
oneDPL has added a `sort_by_key` API, so its best to use that when we can.  This enables optimizations like radix sort for key types which allow it.


danhoeflinger:
It seems there is an issue with same mangled name for a kernel. I will see if it is something with the test, or with oneDPL...

TejaX-Alaghari (approved):
LGTM

the-slow-one (approved):

tomflinda (approved):
LGTM

PR #2003, opened 2024-05-21
Added migration support for 8 D3D11 APIs

TejaX-Alaghari:
Added migration support for 8 D3D11 APIs:

- cudaGraphicsD3D11RegisterResource
- cudaGraphicsResourceSetMapFlags
- cudaGraphicsMapResources
- cudaGraphicsResourceGetMappedMipmappedArray
- cudaGraphicsSubResourceGetMappedArray
- cudaGraphicsUnmapResources
- cudaGraphicsUnregisterResource

Note: `cudaGraphicsResourceGetMappedPointer` API migration is dropped from this PR as it is dependent on `syclexp::map_external_linear_memory` API to be merged by Codeplay.


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -2300,6 +2302,23 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
       return;
     }
 
+    if (CanonicalTypeStr == "cudaGraphicsRegisterFlags") {
+      if (!DpctGlobalInfo::useExtBindlessImages()) {
+        report(TL->getBeginLoc(), Diagnostics::TRY_EXPERIMENTAL_FEATURE, false,
+               "cudaGraphicsRegisterFlags",
+               "--use-experimental-features=bindless_images");
+      }
+    }
```
ziranzha: It is duplicated.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15005,18 +15037,12 @@ void AssertRule::registerMatcher(MatchFinder &MF) {
...
-    if (DpctGlobalInfo::isInAnalysisScope(SpellingLoc)) {
-      ExprAnalysis EA(CE);
-      emplaceTransformation(EA.getReplacement());
-      EA.applyAllSubExprRepl();
-    }
-  }
+  const CallExpr *CE = getNodeAsType<CallExpr>(Result, "FunctionCall");
+  ExprAnalysis EA(CE);
+  emplaceTransformation(EA.getReplacement());
+  EA.applyAllSubExprRepl();
```
ziranzha: Need to use left code when resolving conflicts.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+
+  MF.addMatcher(
+      callExpr(allOf(callee(functionDecl(graphicsInteropAPI())), parentStmt()))
+          .bind("call"),
+      this);
+
+  MF.addMatcher(
+      callExpr(allOf(callee(functionDecl(graphicsInteropAPI())), unless(parentStmt())))
+          .bind("callUsed"),
+      this);
```
ziranzha: Seems migration is same for these two cases, so don't distinguish them.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+    APIName = DC->getNameAsString();
+  } else {
+    return;
+  }
+
+  if (!CallExprRewriterFactoryBase::RewriterMap)
+    return;
+
+  auto Itr = CallExprRewriterFactoryBase::RewriterMap->find(APIName);
+  if (Itr != CallExprRewriterFactoryBase::RewriterMap->end()) {
```
ziranzha: Don't check whether in the `RewriterMap`, using `ExprAnalysis` unconditionally.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+    ExprAnalysis EA(CE);
+    emplaceTransformation(EA.getReplacement());
+    EA.applyAllSubExprRepl();
+    return;
+  }
+
+  if (auto TL = getNodeAsType<TypeLoc>(Result, "resType")) {
```
ziranzha: Seems like this `if` can not be enabled. If the node is matched, the function will exit at line 10107.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+    requestHelperFeatureForTypeNames(
+        DpctGlobalInfo::getUnqualifiedTypeName(TL->getType(), *Result.Context));
+    insertHeaderForTypeRule(
+        DpctGlobalInfo::getUnqualifiedTypeName(TL->getType(), *Result.Context),
+        TL->getBeginLoc());
+
+    if (!ReplType.empty())
+        emplaceTransformation(new ReplaceToken(TL->getBeginLoc(), TL->getEndLoc(),
+                                              std::string(ReplType)));
+    */
```
ziranzha: All the code inside the `if` is commented, it it expected?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
   image_mem_wrapper *_sub_wrappers{nullptr};
 };
 
+#ifdef _WIN32
+std::unordered_set<ID3D11Resource *> uniqueD3D11Res;
+#endif
+
+/// The wrapper class of graphics interop memory handle.
+class interop_mem_wrapper {
```
ziranzha: How about add `static_assert` to notify users that the class is only available on Windows.

clang/lib/DPCT/CallExprRewriterGraphicsInterop.cpp
```diff
@@ -0,0 +1,24 @@
+//===--------------- CallExprRewriterDriver.cpp ---------------------------===//
```
yihwang-nv: It should be CallExprRewriterGraphicsInterop.cpp.

clang/lib/DPCT/APINamesGraphicsInterop.inc
```diff
@@ -0,0 +1,96 @@
+//===--------------- APINamesDriver.inc -----------------------------------===//
```
yihwang-nv: It should be APINamesGraphicsInterop.inc

clang/lib/DPCT/CallExprRewriterGraphicsInterop.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "CallExprRewriterCommon.h"
+
+namespace clang {
+namespace dpct {
+
+void CallExprRewriterFactoryBase::initRewriterMapGraphicsInterop() {
+  RewriterMap->merge(
+      std::unordered_map<std::string,
+                         std::shared_ptr<CallExprRewriterFactoryBase>>({
+#include "APINamesGraphicsInterop.inc"
```
yihwang-nv, line 19: Can we inline clang/lib/DPCT/APINamesGraphicsInterop.inc here?

clang/lib/DPCT/ASTTraversal.h
```diff
@@ -1754,6 +1754,7 @@ class CudaArchMacroRule : public NamedMigrationRule<CudaArchMacroRule> {
   void registerMatcher(ast_matchers::MatchFinder &MF) override;
   void runRule(const ast_matchers::MatchFinder::MatchResult &Result);
 };
+
```
yihwang-nv: Let's remove extra blank line change.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+        "Images with multiple mip and array levels are not supported!");
+    
+    sycl::ext::oneapi::experimental::image_type img_type = 
+        (res_num_levels > 1)?
+            sycl::ext::oneapi::experimental::image_type::mipmap:
+            sycl::ext::oneapi::experimental::image_type::standard;
+
+    _img_desc = sycl::ext::oneapi::experimental::image_descriptor{
+                    sycl::range<3>{res_width, res_height, res_depth},
+                    res_num_channels, img_ch_type, img_type, res_num_levels,
```
tangjj11: Now the bindless image not support using channel num to construct, need add a macro judgment of `__SYCL_COMPILER_VERSION`.

clang/lib/DPCT/DPCT.cpp
```diff
@@ -826,6 +826,23 @@ int runDPCT(int argc, const char **argv) {
 
     Tool.mapVirtualFile(SourcePathList[0], SourceCode);
 
+    static const std::string UnsupportedStr{"// UNSUPPORTED:"};
```
yihwang-nv: Can we use constexpr StringRef UnsupportedStr{"// UNSUPPORTED:"}; instead of a static const string?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+
+      // NOTE: 'map_external_linear_memory' API is not yet merged by Codeplay
+      /*
+      _buf_ptr =
+        sycl::ext::oneapi::experimental::map_external_linear_memory(
+          get_interop_mem_handle(), _img_desc, q);
+      */
+      throw std::runtime_error("Mapping a buffer resoure is not yet supported!");
+    }
+    else {
```
yihwang-nv: Needs format.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+  void unmap_resource(sycl::queue q) {
+    if (_res_is_buffer) {
+      assert(_buf_ptr && "Resource is not mapped!");
+
+      _buf_ptr = nullptr;
+
+      throw std::runtime_error(
+          "Unmapping a buffer resoure is not yet supported!");
+    }
+    else {
```
yihwang-nv: Needs format.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+      }
+      case D3D11_RESOURCE_DIMENSION_TEXTURE1D: {
+        D3D11_TEXTURE1D_DESC desc;
+        ((ID3D11Texture1D*)resource)->GetDesc(&desc);
+
+        res_width = desc.Width;
+        res_arr_size = desc.ArraySize;
+        res_num_levels = (desc.MipLevels? desc.MipLevels: 1);
+
+        get_img_ch_info(
```
tangjj11: Can create a dpct::image_channel object to get the total bytes. Then need not add the `get_img_ch_info` function.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+    auto result = uniqueEntries.insert(entries[i]);
+    if (!result.second) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+static inline void map_resources(int count, interop_mem_wrapper **handles,
```
yihwang-nv: Why we need `static` in a header file?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+    sycl::queue q = get_default_queue()) {
+  assert(!check_duplicate_entries(count, handles) &&
+      "Duplicate handle entries found during resource mapping!");
+
+  for (int i = 0; i < count; i++ ) {
+    handles[i]->map_resource(q);
+  }
+}
+
+static inline void get_mapped_pointer(
```
yihwang-nv: Same as above.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+  assert(handle->is_buffer_res() &&
+      "Non buffer resouce cannot be accessed as a pointer!");
+  
+  *ptr = handle->get_buf_ptr();
+  assert(*ptr && "Resource is not mapped!");
+
+  *size = handle->get_res_size_bytes();
+}
+
+static inline void unmap_resources(int count, interop_mem_wrapper **handles,
```
yihwang-nv: Same as above.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15005,18 +15037,12 @@ void AssertRule::registerMatcher(MatchFinder &MF) {
           .bind("FunctionCall"),
       this);
 }
+
```
yihwang-nv: Extra blank line.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+    emplaceTransformation(EA.getReplacement());
+    EA.applyAllSubExprRepl();
+    return;
+  }
+
+  if (auto TL = getNodeAsType<TypeLoc>(Result, "resType")) {
+    if (isCapturedByLambda(TL))
+      return;
+
+    /*
```
yihwang-nv: Can we remove the comment code?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+
+  // Helper function to get the texture channel info of DX11 resource using DXGI
+  // format
+  void get_img_ch_info(DXGI_FORMAT format, unsigned &bytes,
+      unsigned &res_num_channels, sycl::image_channel_type &img_ch_type) {
+    switch (format) {
+      case DXGI_FORMAT_R16G16B16A16_FLOAT:
+        bytes = 8;
+        res_num_channels = 4;
+        img_ch_type = sycl::image_channel_type::fp16;
```
ziranzha: How about using `dpct::image_channel::create<sycl::half4>`, then get data from `dpct::image_channel`. It can be more readable.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -15042,3 +15068,108 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
...
+          loc(qualType(hasDeclaration(namedDecl(hasAnyName(
+              "cudaGraphicsResource", "cudaGraphicsResource_t"))))))
+          .bind("resType"),
+      this);
+}
+
+void GraphicsInteropRule::runRule(
+    const ast_matchers::MatchFinder::MatchResult &Result) {
+  std::string APIName;
+  std::ostringstream OS;
```
yihwang-nv: It looks like this variable was not used, can we remove it?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+                    res_num_channels, img_ch_type, img_type, res_num_levels,
+                    res_arr_size};
+  }
+
+  // Helper function to create shared handle for DXD11 resource
+  HANDLE create_shared_handle(ID3D11Resource *resource) {
+    HANDLE win_nt_handle = nullptr;
+
+    IDXGIResource1* pDXGIResource1;
+    resource->QueryInterface(__uuidof(IDXGIResource1), reinterpret_cast<void**>(&pDXGIResource1));
```
ziranzha: Does icpx/clang support the operator `__uuidof`?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+    } else {
+      pDXGIResource1->CreateSharedHandle(
+        nullptr,
+        DXGI_SHARED_RESOURCE_READ | DXGI_SHARED_RESOURCE_WRITE,
+        nullptr,
+        &win_nt_handle
+      );
+    }
+
+    pDXGIResource1->Release();
```
ziranzha, line 488: Does `win_nt_handle` still available after `pDXGIRewource1` released?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+        ext_mem_win_handle,
+        sycl::ext::oneapi::experimental::external_mem_handle_type::
+            win32_nt_dx12_resource,
+        _res_size_bytes};
+
+    _interop_mem_handle_ptr =
+      new sycl::ext::oneapi::experimental::interop_mem_handle;
+    *_interop_mem_handle_ptr =
+      sycl::ext::oneapi::experimental::import_external_memory(
+        ext_mem_desc, get_default_queue());
```
ziranzha:
```suggestion
      new (_interop_mem_handle_ptr) sycl::ext::oneapi::experimental::interop_mem_handle(
      sycl::ext::oneapi::experimental::import_external_memory(
        ext_mem_desc, get_default_queue()));
```
BTW, why using a pointer instead of an object for the handle here?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+      // NOTE: 'map_external_linear_memory' API is not yet merged by Codeplay
+      /*
+      _buf_ptr =
+        sycl::ext::oneapi::experimental::map_external_linear_memory(
+          get_interop_mem_handle(), _img_desc, q);
+      */
+      throw std::runtime_error("Mapping a buffer resoure is not yet supported!");
+    }
+    else {
+      assert(!_img_mem_handle_ptr && "Resource is already mapped!");
```
ziranzha: Why the resource cannot be mapped more than once?
Also why use a pointer instead of an object for the handler?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+#endif
+
+/// The wrapper class of graphics interop memory handle.
+class interop_mem_wrapper {
+public:
+#ifdef _WIN32
+  interop_mem_wrapper(ID3D11Resource *d3d11_res, unsigned reg_flags)
+      : _d3d11_res(d3d11_res), _reg_flags(reg_flags) {
+
+    assert(uniqueD3D11Res.find(_d3d11_res) == uniqueD3D11Res.end() &&
```
ziranzha: Why doing the check here?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+    assert(_img_mem_wrapper_ptr && "Resouce is not mapped!");
+
+    return _img_mem_wrapper_ptr;
+  }
+
+  image_mem_wrapper* get_sub_resource_mapped_array(
+      unsigned array_index, unsigned mip_level) {
+    assert(!_res_is_buffer &&
+        "Buffer resouce cannot be accessed as an array!");
+    assert(array_index == 0 &&
```
ziranzha: Same as above.

clang/lib/DPCT/CallExprRewriterGraphicsInterop.cpp
TejaX-Alaghari: Can you please explain your reason for suggesting the inclusion the contents of the .inc file here directly?
The other CallExpr files seem to include the .inc header instead.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+    else {
+      assert(!_img_mem_handle_ptr && "Resource is already mapped!");
+
+      _img_mem_handle_ptr =
+        new sycl::ext::oneapi::experimental::image_mem_handle;
+      *_img_mem_handle_ptr =
+        sycl::ext::oneapi::experimental::map_external_image_memory(
+          get_interop_mem_handle(), _img_desc, q);
+      
+      _img_mem_wrapper_ptr = new image_mem_wrapper(this);
```
ziranzha: `_img_mem_handle_ptr` doesn't need to store in the class. Using `_img_mem_wrapper_ptr` to manage it.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+    }
+  }
+
+  image_mem_wrapper* get_mapped_mipmapped_array() {
+    assert(!_res_is_buffer &&
```
ziranzha: Should throw exceptions for when encounter errors.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +137,341 @@ class image_mem_wrapper {
...
+    if (_res_is_buffer) {
+      assert(_buf_ptr && "Resource is not mapped!");
+
+      _buf_ptr = nullptr;
+
+      throw std::runtime_error(
+          "Unmapping a buffer resoure is not yet supported!");
+    }
+    else {
+      assert(_img_mem_handle_ptr && "Resource is not mapped!");
```
ziranzha: Same as above.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+      return true;
+    }
+  }
+
+  return false;
+}
+
+static inline void map_resources(int count, interop_mem_wrapper **handles,
+    sycl::queue q = get_default_queue()) {
+  assert(!check_duplicate_entries(count, handles) &&
```
ziranzha: Should throw exceptions for the errors.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: `static` keyword is added to prevent multiple redefinitions of the function when compiling multiple src files

If this is not a good enough reason, please let me know.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
 }
 } // namespace detail
 
+template <typename T>
+bool check_duplicate_entries(int count, T **entries) {
+  std::unordered_set<T *> uniqueEntries;
```
ziranzha: Using `std::set` maybe faster if `count` is not a big number.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +709,51 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+      "Duplicate handle entries found during resource mapping!");
+
+  for (int i = 0; i < count; i++ ) {
+    handles[i]->map_resource(q);
+  }
+}
+
+static inline void get_mapped_pointer(
+    void **ptr, size_t *size, interop_mem_wrapper *handle) {
+  assert(handle->is_buffer_res() &&
```
ziranzha: Same as above.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
ziranzha: Same as above.

ziranzha (changes requested):

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Yes, code ran as expected without errors on a Windows machine when tested with clang++ compiler

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Yes, handle will be alive as `resource` is still alive
And `pDXGIRewource1` is just a referenced interface of the `resource` which can be released after its usage.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: This check is added to honor CUDA's condition [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP_1gad8fbe74d02adefb8e7efb4971ee6322:~:text=If%20any%20of%20resources%20are%20presently%20mapped%20for%20access%20by%20CUDA%20then%20cudaErrorUnknown%20is%20returned.) for `cudaGraphicsMapResources`

clang/lib/DPCT/ASTTraversal.cpp
TejaX-Alaghari: I've removed the commented code now.
The file was not cleaned up properly.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: To prevent the resource from registering to multiple wrappers.
Similar to the behavior [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__D3D11.html#group__CUDART__D3D11_1g85d07753780643584b8febab0370623b:~:text=If%20pD3DResource%20is%20of%20incorrect%20type%20or%20is%20already%20registered%2C%20then%20cudaErrorInvalidResourceHandle%20is%20returned) of `cudaGraphicsD3D11RegisterResource`

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Implemented as a pointer to check if a resource is registered or not (if `_interop_mem_handle_ptr` is NULL or not) during unregistration (destructor).

Please suggest if there is a better mechanism to achieve this without pointers

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: This check is added to honor CUDA's condition [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP_1g0dd6b5f024dfdcff5c28a08ef9958031:~:text=resource%20is%20mapped.-,If%20resource%20is%20not%20a%20texture%20then%20it%20cannot%20be%20accessed%20via%20an%20array%20and%20cudaErrorUnknown%20is%20returned,-.%20If%20arrayIndex%20is) for `cudaGraphicsSubResourceGetMappedArray`

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: This class should be common to both Win and Lin platforms.

Only the constructor `interop_mem_wrapper(ID3D11Resource* d3d11_res, int reg_flags)` is available for Win.
And another constructor `interop_mem_wrapper(GLuint gl_res, int reg_flags)` to be added in the future for OpenGL support will be available for both platforms.

Similar logic applies for Vulkan and other Graphics libraries

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
ziranzha: This pointer is initialized in constructor, and I see the check is done at the beginning of the constructor. So I think it cannot be used for the check.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
ziranzha: The initialization seems very different for the platforms. I suggest put the common logic in a base class, and the implementation for different platforms can be put in derived classes, or classes have a same name but separated by macros. I think it can be more readable.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
...
+    }
+
+    pDXGIResource1->Release();
+
+    return win_nt_handle;
+  }
+
+  // Helper function to get the texture channel info of DX11 resource using DXGI
+  // format
+  void get_img_ch_info(DXGI_FORMAT format, image_channel &channel) {
```
ziranzha:
```suggestion
  image_channel get_img_ch_info(DXGI_FORMAT format) {
```

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
...
+          _img_mem_wrapper_ptr->get_mip_level(mip_level);
+  }
+
+  void unmap_resource(sycl::queue q) {
+    if (_res_is_buffer) {
+      if (!_buf_ptr) {
+        throw std::runtime_error("Resource is not mapped!");
+      }
+
+      _buf_ptr = nullptr;
```
ziranzha: Should `_but_ptr` be freed before set null?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
...
+  }
+
+  image_mem_wrapper* get_mapped_mipmapped_array() {
+    if(_res_is_buffer) {
+      throw std::runtime_error(
+          "Buffer resouce cannot be accessed as an array!");
+    }
+    if(!_img_mem_wrapper_ptr) {
+      throw std::runtime_error("Resouce is not mapped!");
+    }
```
ziranzha, line 225: Should it throw an exception if `_img_mem_wrappter_ptr` is not a mipmapped array?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
   image_mem_wrapper *_sub_wrappers{nullptr};
 };
 
+#ifdef _WIN32
+std::unordered_set<ID3D11Resource *> uniqueD3D11Res;
```
ziranzha: Suggest to move into `class interop_mem_wrapper` as a static member.

clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -3286,6 +3298,18 @@ void EnumConstantRule::runRule(const MatchFinder::MatchResult &Result) {
...
     return;
+  } else if ((EnumName == "cudaGraphicsRegisterFlagsNone" ||
+              EnumName == "cudaGraphicsRegisterFlagsReadOnly" ||
+              EnumName == "cudaGraphicsRegisterFlagsWriteDiscard" ||
+              EnumName == "cudaGraphicsRegisterFlagsSurfaceLoadStore" ||
+              EnumName == "cudaGraphicsRegisterFlagsTextureGather" ||
+              EnumName == "cudaGraphicsMapFlagsNone" ||
+              EnumName == "cudaGraphicsMapFlagsReadOnly" ||
+              EnumName == "cudaGraphicsMapFlagsWriteDiscard") &&
+             !DpctGlobalInfo::useExtBindlessImages()) {
```
ziranzha: Suggest move `!DpctGlobalInfo::useExtBindlessImages()` to the beginning.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: I wanted a mechanism to check if the resource was registered or not before releasing its memory in the wrapper destructor ([here](https://github.com/oneapi-src/SYCLomatic/blob/55de4c380edbdf6a04de6d7b13430848b92a8622/clang/runtime/dpct-rt/include/dpct/bindless_images.hpp#L193)).
That's the main motivation for making it a pointer.

The check at the beginning of the constructor is done with the D3D11 resource itself

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: The `_buf_ptr` obtained during mapping should not be freed because the underlying memory is created by DX11 and should be available to be used for DX11 APIs after unmapping of the resource CUDA/SYCL.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: There is no mention of this check in the CUDA [doc](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP_1g84c3772d2ed06cda8c92bc43cdc893d0:~:text=If%20resource%20is%20not%20a%20texture%20then%20it%20cannot%20be%20accessed%20via%20an%20array%20and%20cudaErrorUnknown%20is%20returned.%20If%20resource%20is%20not%20mapped%20then%20cudaErrorUnknown%20is%20returned.) for `cudaGraphicsResourceGetMappedMipmappedArray `.

Besides, there are only 2 ways to retrieve the underlying image data of resource in CUDA - 
- `cudaGraphicsResourceGetMappedMipmappedArray(array, resource)`
- `cudaGraphicsSubResourceGetMappedArray(array, resource, arrayIndex, mipLevel)`

As any non-buffer (texture) resource qualifies as a mipmap image with 1 level, I believe that's the reason why the image returned by `cudaGraphicsResourceGetMappedMipmappedArray` is not checked for having multiple levels

Instead, this check is done for `cudaGraphicsSubResourceGetMappedArray` ([ref](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP_1g84c3772d2ed06cda8c92bc43cdc893d0:~:text=If%20mipLevel%20is%20not%20a%20valid%20mipmap%20level%20for%20resource%20then%20cudaErrorInvalidValue%20is%20returned.)), which is used to get a specific level of a mipmap image.

Please correct/suggest changes for my thoughts on this.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
yihwang-nv: `inline` will get rid of linking errors.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -67,6 +71,21 @@ class image_mem_wrapper {
     auto q = get_default_queue();
     _handle = alloc_image_mem(_desc, q);
   }
+  /// Create bindless image memory wrapper.
+  /// \param [in] res_interop_mem_wrapper The interop mem wrapper of a
```
tangjj11: The comment is not match the function.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -67,6 +71,21 @@ class image_mem_wrapper {
...
+      sycl::ext::oneapi::experimental::image_mem_handle img_mem_handle)
+      : _desc(desc), _handle(img_mem_handle) {
+    _channel.set_channel_type(_desc.channel_type);
+  #if (__SYCL_COMPILER_VERSION && __SYCL_COMPILER_VERSION >= 20240725)
+    _channel.set_channel_num(_desc.num_channels);
+  #endif
+    auto q = get_default_queue();
+    init_mip_level_wrappers(q);
+  }
+  image_mem_wrapper(const interop_mem_wrapper *);
```
tangjj11: Cannot find the definition of this ctor.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -77,6 +96,21 @@ class image_mem_wrapper {
     }
     free_image_mem(_handle, _desc.type, get_default_queue());
   }
+  /// Initialize image mem wrappers for all the levels of mipmap image
+  /// @param q The Queue to be used to query the mip levels 
+  void init_mip_level_wrappers(sycl::queue q) {
```
tangjj11: This method should be private, if no migrated code use it directly.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
tangjj11: If user need not use it directly, the member can be private.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
...
+      }
+
+      delete _img_mem_wrapper_ptr;
+      _img_mem_wrapper_ptr = nullptr;
+    }
+  }
+
+#ifdef _WIN32
+  // Helper function to query the properties of DX11 resource
+  void query_res_info(ID3D11Resource *resource) {
```
tangjj11, line 381: Can be private? Because I can not find it usage in migrated code.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -117,6 +151,349 @@ class image_mem_wrapper {
...
+  get_interop_mem_handle() const noexcept {
+    return *_interop_mem_handle_ptr;
+  }
+
+  sycl::ext::oneapi::experimental::image_mem_handle
+  get_img_mem_handle() const noexcept {
+    return _img_mem_wrapper_ptr->get_handle();
+  }
+
+  void *
```
tangjj11: it seem like this file not formated properly.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -354,6 +731,57 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
 }
 } // namespace detail
 
+template <typename T>
+bool check_duplicate_entries(int count, T **entries) {
```
tangjj11: If migrated code will not use it, should put in detail namespace.

clang/lib/DPCT/DPCT.cpp
```diff
@@ -826,6 +826,23 @@ int runDPCT(int argc, const char **argv) {
...
+    bool win_unsupported = false;
+    bool lin_unsupported = false;
+    if (SourceCode.starts_with(UnsupportedStr)) {
```
tangjj11, line 840: I think we need not these logic. Because the query mapping need keep consistent with that we use dpct to migrate a example code. It seem like we can migrate the `cudaGraphicsD3D11RegisterResource` APIs with experimental option on both win and lin. So need not report warning when querying.

clang/lib/DPCT/ASTTraversal.h
```diff
@@ -1800,6 +1800,13 @@ class AssertRule : public NamedMigrationRule<AssertRule> {
   void runRule(const ast_matchers::MatchFinder::MatchResult &Result);
 };
 
+/// Graphics Interoperability migraiton rule
```
tangjj11: The comment is unnecessary because other rule has not comment too.

clang/lib/DPCT/APINamesGraphicsInterop.inc
```diff
@@ -0,0 +1,96 @@
...
+        checkIsCallExprOnly(),
+        WARNING_FACTORY_ENTRY("cudaGraphicsResourceSetMapFlags",
+            TOSTRING_FACTORY_ENTRY("cudaGraphicsResourceSetMapFlags", LITERAL("")),
+            Diagnostics::FUNC_CALL_REMOVED,
+            std::string("cudaGraphicsResourceSetMapFlags"), getRemovedAPIWarningMessage("cudaGraphicsResourceSetMapFlags")),
+        WARNING_FACTORY_ENTRY("cudaGraphicsResourceSetMapFlags",
+            TOSTRING_FACTORY_ENTRY("cudaGraphicsResourceSetMapFlags", LITERAL("0")),
+            Diagnostics::FUNC_CALL_REMOVED_0,
+            std::string("cudaGraphicsResourceSetMapFlags"), getRemovedAPIWarningMessage("cudaGraphicsResourceSetMapFlags"))),
+    UNSUPPORT_FACTORY_ENTRY("cudaGraphicsResourceSetMapFlags", Diagnostics::TRY_EXPERIMENTAL_FEATURE,
```
tangjj11: It seems like for the removed API rule, we need not check whether the bindless images extension is enabled. Because se not use the bindless image API at all.

clang/lib/DPCT/APINamesGraphicsInterop.inc
```diff
@@ -0,0 +1,96 @@
...
+                            ARG("--use-experimental-features=bindless_images"))))
+
+ASSIGNABLE_FACTORY(CONDITIONAL_FACTORY_ENTRY(
+    UseExtBindlessImages,
+    DELETER_FACTORY_ENTRY("cudaGraphicsUnregisterResource", ARG(0)),
+    UNSUPPORT_FACTORY_ENTRY("cudaGraphicsUnregisterResource", Diagnostics::TRY_EXPERIMENTAL_FEATURE,
+                            ARG("cudaGraphicsUnregisterResource"),
+                            ARG("--use-experimental-features=bindless_images"))))
+
+// clang-format on
```
tangjj11, line 106: The file without clang-format seems like not easier to read compared with non-format. Please try format and not format the file, and compare which one is better.

tangjj11 (changes requested):

clang/lib/DPCT/DPCT.cpp
TejaX-Alaghari: The `cudaGraphicsD3D11RegisterResource` API is only migrated on Windows. 
On Linux, dpct returns below output:

`dpct exited with code: 1 (Migration not necessary; no CUDA code detected.)`

This is because this API is present in the header `cuda_d3d11_interop.h` which is only available on Windows platform.

Hence, I've devised this logic to parse supported and unsupported plaatforms.
Please let me know your thoughts on the same.

tangjj11: ok, currently I have no better solution, just keep this solution is ok to me.

tangjj11 (approved):

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
ziranzha: But it is always initialized in the constructor. In destructor, there is no check needed, since it cannot be uninitialized.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
ziranzha: Sounds good.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
   const sycl::ext::oneapi::experimental::image_descriptor _desc;
   sycl::ext::oneapi::experimental::image_mem_handle _handle;
   image_mem_wrapper *_sub_wrappers{nullptr};
 };
 
+/// The base wrapper class of graphics interop memory handle.
+class interop_mem_wrapper_base {
+public:
+  /// Creates interop memory wrapper for a given resource.
+  interop_mem_wrapper_base() = default;
```
ziranzha:
```suggestion
  interop_mem_wrapper_base(sycl::ext::oneapi::experimental::interop_mem_handle interop_mem_handle); // Initialize the interop_mem_handl
```

Put it to `protected:`

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
+      break;
+    }
+
+    return channel;
+  }
+};
+#else
+/// Stub interop_mem_wrapper implementation for non-Windows platforms
+class interop_mem_wrapper : public interop_mem_wrapper_base {
+public:
```
ziranzha: Add `static_assert` for unsupported platforms.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
+  }
+
+private:
+  static std::set<ID3D11Resource *> _res_D3D11_set;
+
+  ID3D11Resource *_res_D3D11 = nullptr;
+  HANDLE _res_win_nt_handle = nullptr;
+  unsigned _res_reg_flags = 0;
+  sycl::ext::oneapi::experimental::interop_mem_handle
+      *_res_interop_mem_handle_ptr = nullptr;
```
ziranzha: Why have this member in both derived and base classes?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
+
+    image_channel channel;
+
+    // Get the dimension info of DX11 resource
+    D3D11_RESOURCE_DIMENSION dimension;
+    resource->GetType(&dimension);
+
+    switch (dimension) {
+    case D3D11_RESOURCE_DIMENSION_BUFFER: {
+      _res_is_buffer = true;
```
ziranzha, line 399: Cannot access the private member in the base class.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
+
+      D3D11_BUFFER_DESC desc;
+      ((ID3D11Buffer *)resource)->GetDesc(&desc);
+
+      _res_size_bytes = desc.ByteWidth;
```
ziranzha, line 404: Same as above

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,401 @@ class image_mem_wrapper {
...
+
+      _res_size_bytes = res_width * res_height * res_depth * res_arr_size *
+                        res_num_levels * channel.get_total_size();
+
+      sycl::ext::oneapi::experimental::image_type img_type =
+          (res_num_levels > 1)
+              ? sycl::ext::oneapi::experimental::image_type::mipmap
+              : sycl::ext::oneapi::experimental::image_type::standard;
+
+      _res_img_desc = sycl::ext::oneapi::experimental::image_descriptor{
```
ziranzha, line 462: Same as above

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: It was a faulty commit. I've corrected it now.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Made it a protected member.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Made it a protected member.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Made it a protected member.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Addressed - no need to insert static_assert

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -352,8 +765,53 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
+}
+
+/// Unmap the resource memories to mem handles
+/// \param [in] count The count of resources to unmap.
+/// \param [in] handles The external mem wrappers used to unmap the resources.
+/// \param [in] q The queue used to unmap the resource with.
+inline void unmap_resources(int count, external_mem_wrapper **handles,
+                            queue_ptr q_ptr = &get_default_queue()) {
+  if (detail::check_duplicate_entries(count, handles) &&
+      "Duplicate handle entries found during resource unmapping!")
```
zhimingwang36, line 783: L807,  change to L791?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -352,8 +765,53 @@ dpct_memcpy(pitched_data src, const sycl::id<3> &src_id,
...
 } // namespace detail
 
+/// Map the resource memories to mem handles
+/// \param [in] count The count of resources to map.
+/// \param [in] handles The external mem wrappers used to map the resources.
+/// \param [in] q The queue used to map the resource with.
+inline void map_resources(int count, external_mem_wrapper **handles,
+                          queue_ptr q_ptr = &get_default_queue()) {
+  if (detail::check_duplicate_entries(count, handles)) {
+    throw std::runtime_error(
```
zhimingwang36, line 767: how about do like this,  only map one time for duplicated, and return same handle for duplicated items.  then no warnning needed.

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,410 @@ class image_mem_wrapper {
...
+    resource->QueryInterface(__uuidof(IDXGIKeyedMutex),
+                             reinterpret_cast<void **>(&_res_keyed_mutex));
+
+    // Use the keyed mutex to synchronize access to the shared resource
+    _res_keyed_mutex->AcquireSync(0, INFINITE);
+
+    return win_nt_handle;
+  }
+
+  /// Helper function to get the texture channel info of DX11 resource using
```
zhimingwang36, line 500:
```suggestion
  /// Get the texture channel info of DX11 resource using
```

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
```diff
@@ -111,12 +126,410 @@ class image_mem_wrapper {
...
+  /// \param [in] d3d11_res Pointer to the ID3D11Resource to be wrapped.
+  /// \param [in] reg_flags Registration flags used for resource registration.
+  external_mem_wrapper(ID3D11Resource *d3d11_res, unsigned reg_flags)
+      : _res_D3D11(d3d11_res),
+        external_mem_wrapper_base(
+            sycl::ext::oneapi::experimental::external_mem(), reg_flags) {
+    // Check if the resource is already registered to prevent double
+    // registration.
+    if (_res_D3D11_set.find(_res_D3D11) != _res_D3D11_set.end()) {
+      throw std::runtime_error("This D3D11 resource is already registered!");
```
zhimingwang36, line 318: can we return the registered one directly without runtime_error?

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: Written this logic to emulate CUDA's behavior

Ref: [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP_1gad8fbe74d02adefb8e7efb4971ee6322:~:text=in%20stream%20begins.-,If%20resources%20contains%20any%20duplicate%20entries%20then%20cudaErrorInvalidResourceHandle%20is%20returned.,-If%20any%20of)

clang/runtime/dpct-rt/include/dpct/bindless_images.hpp
TejaX-Alaghari: This logic is added for emulating the CUDA behavior here ([ref](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__D3D11.html#group__CUDART__D3D11_1g85d07753780643584b8febab0370623b:~:text=If%20pD3DResource%20is%20of%20incorrect%20type%20or%20is%20already%20registered%2C%20then%20cudaErrorInvalidResourceHandle%20is%20returned))

zhimingwang36 (approved):

PR #2009, opened 2024-05-23
[SYCLomatic] Add support for `cuFuncSetAttribute` migration

the-slow-one:
`cuFuncSetAttribute` will be removed from the CUDA source files as DPC++ don't support setting attribute for kernel function.


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -12329,6 +12333,18 @@ void KernelFunctionInfoRule::runRule(const MatchFinder::MatchResult &Result) {
     if (NameMap != AttributesNamesMap.end())
       emplaceTransformation(new ReplaceToken(MemberName.getBeginLoc(),
                                              std::string(NameMap->second)));
+
+  } else if (auto *CallNode =
+                 getNodeAsType<CallExpr>(Result, "cuFuncSetAttribute")) {
+    std::string FuncName =
+        CallNode->getDirectCallee()->getNameInfo().getName().getAsString();
+    auto Msg = MapNames::RemovedAPIWarningMessage.find(FuncName);
+    // TODO: Handle assignments
```
tangjj11: it is similar  to line 12312, can add a `if (isAssigned(C)) {`.

tangjj11 (approved):

clang/lib/DPCT/ASTTraversal.cpp
tomflinda: I think so, cuFuncSetAttribute has a return value, it may be checked in the user's code.

tomflinda (approved):
LGTM

PR #2018, opened 2024-05-28
[SYCLomatic] Change std numeric  to oneapi dpl numeric impl

abhilash1910:
Changes numeric.h to support oneapi::dpl:: primitives. 
cc @danhoeflinger @mmichel11  @yihanwg @ShengchenJ


danhoeflinger:
Why is this PR needed?  These wrappers don't seem to add anything to the oneDPL function they wrap.  Can the calling code just call oneDPL directly (or interchangeably, the std API with an execution policy)?

abhilash1910:
Yes the main intent is to use the dpl invocations directly from numeric than the std:: variants for this numeric file. Also increased the added feature set of reduce ops over the transform_reduce to the code . Doesnt change any functionality of the apis.

abhilash1910:
Close as not required.

PR #2024, opened 2024-05-30
[SYCLomatic] Add support for `cuDeviceCanAccessPeer`

the-slow-one:
If the peer-to-peer support extension is enabled, migrate the API. Otherwise add a comment to enable extension. The API mapping is same as `cudaDeviceCanAccessPeer`.


TejaX-Alaghari (approved):
LGTM

clang/lib/DPCT/APINamesMemory.inc
```diff
@@ -613,6 +613,23 @@ ASSIGNABLE_FACTORY(FEATURE_REQUEST_FACTORY(
...
 
+CONDITIONAL_FACTORY_ENTRY(
+    UsePeerAccess(),
+    ASSIGNABLE_FACTORY(ASSIGN_FACTORY_ENTRY(
+        "cuDeviceCanAccessPeer", DEREF(ARG_WC(0)),
+        MEMBER_CALL(CALL(MapNames::getDpctNamespace() +
+                             "dev_mgr::instance().get_device",
+                         ARG_WC(1)),
+                    false, "ext_oneapi_can_access_peer",
+                    CALL(MapNames::getDpctNamespace() +
```
AndyCHHuang: this should be another member call

the-slow-one: We only want to get the second device and pass it to `ext_oneapi_can_access_peer`. And we don't want to access any members from the device returned from `get_device()`.

Therefore, it's just a call.

AndyCHHuang: "dpct::dev_mgr::instance().get_device(device).ext_oneapi_can_access_peer(dpct::dev_mgr::instance().get_device(peerDevice));"

The first dpct::dev_mgr::instance().get_device(device) is a member call.
ext_oneapi_can_access_peer() is another membercall.
The arg "dpct::dev_mgr::instance().get_device(peerDevice)" is also a member call.

AndyCHHuang (approved):
LGTM

clang/lib/DPCT/APINamesMemory.inc
the-slow-one: Replaced CALL with MEMBER_CALL template.

AndyCHHuang (approved):
LGTM

PR #2032, opened 2024-06-04
[SYCLomatic] Introduce `dpct::argmax` and `dpct::argmin` functors

mmichel11:
This PR introduces `dpct::argmax` and `dpct::argmin` functors for determining the arg max / arg min of two `dpct::key_value_pair` types.

These functors have the following behavior:

`dpct::argmin`

- If the two `key_value_pair` objects have different values, then return a copy of the `key_value_pair` with the lower value.
- If the two `key_value_pair` objects the same values, then return a copy of the `key_value_pair` with the lower key.

`dpct::argmax`

- If the two `key_value_pair` objects have different values, then return a copy of the `key_value_pair` with the greater value.
- If the two `key_value_pair` objects the same values, then return a copy of the `key_value_pair` with the **lower** key.

The corresponding test PR is https://github.com/oneapi-src/SYCLomatic-test/pull/724

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/functional.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/functional.h
@@ -58,6 +58,40 @@ template <typename _Op> struct mark_functor_const {
   }
 };
 
+// Forward declare key_value_pair to avoid creating cyclic dependency between
+// iterators.h and functional.h.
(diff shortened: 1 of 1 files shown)


timmiesmith (approved):
LGTM

tomflinda (approved):
LGTM

PR #2033, opened 2024-06-04
[SYCLomatic] Fix a crash in case for `cuEvent*` API with explicist case parameter

the-slow-one:
The following code would result in SIGABRT.

```cpp
#include <cstdint>
#include <cuda.h>

void foo(CUevent hEvent, CUstream hStream ) {
  CUresult result = cuEventRecord((CUevent)hEvent, (CUstream)hStream);
}
```

Issue is with the `isa<>` check of first argument's type, specifically when it's typecasted as shown in example.
Fix is to handle the explicit cast expression of the first argument to `cuEvent*` APIs.


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -7367,15 +7369,19 @@ void EventAPICallRule::handleEventRecordWithProfilingDisabled(
   // Insert the helper variable right after the event variables
   static std::set<std::pair<const Decl *, std::string>> DeclDupFilter;
   auto &SM = DpctGlobalInfo::getSourceManager();
-  const ValueDecl *MD = getDecl(CE->getArg(0));
-  std::string InsertStr;
+
+  const ValueDecl *MD = nullptr;
+  if((MD = getDecl(CE->getArg(0))) == nullptr)
```
the-slow-one:
```suggestion
  if ((MD = getDecl(CE->getArg(0))) == nullptr)
```

clang/test/dpct/driver-stream-and-event.cu
```diff
@@ -123,3 +123,9 @@ void test_stream() {
   // CHECK: dpct::get_current_device().destroy_queue(hStream);
   cuStreamDestroy(hStream);
 }
+
+void test_cuEventRecord_crash(uint64_t hEvent,uint64_t hStream )
+{
+  // CHECK: int result = DPCT_CHECK_ERROR(*(CUevent)hEvent = (CUstream)hStream->ext_oneapi_submit_barrier());
```
zhiweij1: The type of the C-style cast also needs to be migrated.
I think it should be something like:
```
// CHECK: int result = DPCT_CHECK_ERROR(*(dpct::event_ptr)hEvent = (dpct::queue_ptr)hStream->ext_oneapi_submit_barrier());
```

zhiweij1 (changes requested):

clang/test/dpct/driver-stream-and-event.cu
the-slow-one: Done

zhiweij1 (approved):

AndyCHHuang (approved):
LGTM

PR #2038, opened 2024-06-06
[DOC] Add feedback link to a second place in Migration Workflow

lapel22:
Adding link to provide feedback in an extra spot in the Migration Workflow.


zhimingwang36 (approved):
LGTM

PR #1970, opened 2024-05-10
[SYCLomatic] Refine migration of cudaEventQuery

tomflinda:
Signed-off-by: chenwei.sun <chenwei.sun@intel.com>

PR #2042, opened 2024-06-06
[SYCLomatic] Add support for cuStreamQuery

the-slow-one:
`cuStreamQuery()` is migrated to `ext_oneapi_empty()` if correspoinding DPC++ extension isn't disabled.


clang/lib/DPCT/ASTTraversal.cpp
```diff
@@ -8299,12 +8299,13 @@ void StreamAPICallRule::runRule(const MatchFinder::MatchResult &Result) {
...
-             FuncName == "cudaStreamQuery") {
+             FuncName == "cudaStreamQuery" ||
+             FuncName == "cuStreamQuery") {
 
     // if extension feature sycl_ext_oneapi_queue_empty is used, member
     // functions "ext_oneapi_empty" in SYCL queue is used to map
-    // cudaStreamQuery.
-    if (FuncName == "cudaStreamQuery" && DpctGlobalInfo::useQueueEmpty()) {
+    // cudaStreamQuery or cuStreamQuery.
+    if ((FuncName == "cudaStreamQuery" || FuncName == "cuStreamQuery") && DpctGlobalInfo::useQueueEmpty()) {
```
daiyaan-ahmed6: Please check if it meets clang format

the-slow-one: Fixed the format! CLI passes.

daiyaan-ahmed6 (approved):
LGTM 👍🏼

clang/test/dpct/disable-all-extensions.cu
```diff
@@ -402,6 +402,16 @@ static void func() {
   // CHECK-NEXT: MY_ERROR_CHECKER(0);
   MY_ERROR_CHECKER(cudaStreamQuery(s0));
 
+  // CHECK: /*
+  // CHECK-NEXT: DPCT1026:{{[0-9]+}}: The call to cuStreamQuery was removed because SYCL currently does not support query operations on queues.
+  // CHECK-NEXT: */
+  cuStreamQuery(s0);
```
zhiweij1, line 411: Could you give an example of how this API is used in the real code?

The return value of APIs like `cu[da]EventQuery/cu[da]StreamQuery` is a little different from other APIs.
It can return a non-zero value (such as `CUDA_ERROR_NOT_READY`) which does not meaning an error.
E.g., for cudaEventQuery in below code we cannot let the migrated API always return 0.
```
    unsigned long int counter = 0;
    while (cudaEventQuery(stop) == cudaErrorNotReady) {
        counter++;
    }
```

the-slow-one: Here is the snippent of a code from a project.

```cuda
CUresult result = cuStreamQuery((CUstream)hStream );
ASSERT( result==CUDA_SUCCESS || result==CUDA_ERROR_NOT_READY, "Failed to query stream");
```

The function containing the above code doesn't contain any loop on the API! But I do see your point. Do you have any suggestion for handling the `cuEventQuery` in loops?

zhiweij1: I'm OK with replacing this API with 0 if it is not "really" used in the project, like in your example.
If you want to support the in-loop scenario, you can refer to PR https://github.com/oneapi-src/SYCLomatic/pull/1970.

clang/runtime/dpct-rt/include/dpct/util.hpp
```diff
@@ -1107,6 +1107,10 @@ class args_selector<n_nondefault_params, n_default_params, R(Ts...)> {
 #define DPCT_EXPORT
 #endif
 
+inline bool sycl_stream_query(dpct::queue_ptr queue) {
+  return false;
+}
+
```
the-slow-one:
```suggestion
```

zhiweij1 (approved):

PR #2078, opened 2024-06-19
[SYCLomatic] Add Block Shuffle Up/Down

abhilash1910:
PR for block shuffle support (up/down) for application enabling

cc @danhoeflinger @mmichel11 @yihanwg @zhimingwang36

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/group_utils.hpp
+++ b/clang/runtime/dpct-rt/include/dpct/group_utils.hpp
@@ -559,7 +559,7 @@ public:
     size_t group_work_items = item.get_local_range().size();
     temp_storage[linear_tid] = input[ITEMS_PER_WORK_ITEM - 1];
 
-    item.barrier(sycl::access::fence_space::local_space);
+    sycl::group_barrier(item.get_group());
 
 #pragma unroll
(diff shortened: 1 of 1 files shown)


clang/runtime/dpct-rt/include/dpct/group_utils.hpp
```diff
@@ -527,6 +527,94 @@ class workgroup_load {
...
+  __dpct_inline__ void shuffle_up(const Item &item,
+                                  T (&input)[ITEMS_PER_WORK_ITEM],
+                                  T (&prev)[ITEMS_PER_WORK_ITEM],
+                                  T &block_suffix) {
+
+    size_t linear_tid = item.get_local_linear_id();
+    size_t group_work_items = item.get_local_range().size();
+    temp_storage[linear_tid] = input[ITEMS_PER_WORK_ITEM - 1];
+
+    item.barrier(sycl::access::fence_space::local_space);
```
mmichel11: The `nd_item::barrier` function does not exist in the SYCL2020 spec and will be removed sometime in the future. Instead, we should use `sycl::group_barrier` here.

The switch may also negatively impact performance as `sycl::group_barrier` does not accept an address space and may synchronize the global space as well.

abhilash1910:
Closed as implemented.

PR #2109, opened 2024-07-01
[SYCLomatic] Add support for `cuCtxPushContext` and `cuCtxPopContext`

the-slow-one:
* Use stack to store the current context (DPCT device extension)
* Add support for `cuCtxPushContext` and `cuCtxPopContext` based on above design
* Update the support for `cuCreateCtx` based on the above design


clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -662,7 +663,14 @@ class dev_mgr {
...
     check_id(id);
-    _thread2dev_map[get_tid()] = id;
+    auto curr_tid = get_tid();
+
+    // make sure the device stack exist
+    auto it = _thread2dev_map.find(curr_tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map[curr_tid] = std::stack<unsigned int>();
+
+    _thread2dev_map[curr_tid].push(id);
```
zhiweij1: There are two unnecessary search operations for the map introduced by the operator[].
How about changing to:
```suggestion
    if (it != _thread2dev_map.end())
      it->second.push(id);
    else {
      std::stack<unsigned int> s;
      s.push(id);
      _thread2dev_map.insert(id);
    }
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +740,31 @@ class dev_mgr {
     select_device(selected_device_id);
   }
 
+  /// Update the device stack for the current thread id
+  void push_device(unsigned int id) {
+    auto tid = get_tid();
+    auto it = _thread2dev_map.find(tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map[tid] = std::stack<unsigned int>();
```
zhiweij1: Same as above.

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +740,31 @@ class dev_mgr {
...
+
+  /// Remove the device from top of the stack if it exist
+  unsigned int pop_device() {
+    auto tid = get_tid();
+    auto it = _thread2dev_map.find(tid);
+    if (it == _thread2dev_map.end())
+      return -1;
+
+    if (it->second.empty())
+      return -1;
```
zhiweij1: Does CUDA return a default context even if the stack is empty?

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Agreed. Code modified.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Code modified

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: I could find any clear statement about what happens when the stack is empty. I tried an example though and ***it seems like any pop is a no-op when the stack is empty.*** Comments?

```
#include <cuda.h>
#include <iostream>


int main() {
  CUdevice  dev;
  CUcontext ctx;

  cuInit(0);

  cudaGetDevice(&dev);

  cuCtxGetCurrent(&ctx);
  std::cout << ctx << "\n";

  cuCtxCreate(&ctx, 0, dev);
  std::cout << "After create " << ctx << "\n";

  cuCtxPushCurrent(ctx);
  std::cout << "After push " << ctx << "\n";

  cuCtxPopCurrent(&ctx);
  std::cout << ctx << "\n";

  cuCtxPopCurrent(&ctx);
  std::cout << ctx << "\n";

  ctx = 0;

  cuCtxPopCurrent(&ctx);
  std::cout << ctx << "\n";

  cuCtxPopCurrent(&ctx);
  std::cout << ctx << "\n";

  cuDevicePrimaryCtxRetain(&ctx, dev);

  cuCtxDestroy(ctx);

  return 0;
}
```

The output is as the following
```
shliclel672:/export/users/dhr/cuda_samples/cuCtx
$ ./a.out
0
After create 0x557a84f83b10
After push 0x557a84f83b10
0x557a84f83b10
0x557a84f83b10
0
0
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -662,7 +663,16 @@ class dev_mgr {
...
+    auto curr_tid = get_tid();
+
+    // make sure the device stack exist
+    auto it = _thread2dev_map.find(curr_tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map[curr_tid] = std::stack<unsigned int>({id});
+    else
+      it->second.push(id);
+
+    _thread2dev_map[curr_tid].push(id);
```
zhiweij1:
```suggestion
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -662,7 +663,16 @@ class dev_mgr {
   void select_device(unsigned int id) {
     std::lock_guard<std::recursive_mutex> lock(m_mutex);
     check_id(id);
-    _thread2dev_map[get_tid()] = id;
+    auto curr_tid = get_tid();
+
+    // make sure the device stack exist
+    auto it = _thread2dev_map.find(curr_tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map[curr_tid] = std::stack<unsigned int>({id});
```
zhiweij1:
```suggestion
      _thread2dev_map.insert({curr_tid, std::stack<unsigned int>({id})});
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +742,32 @@ class dev_mgr {
     select_device(selected_device_id);
   }
 
+  /// Update the device stack for the current thread id
+  void push_device(unsigned int id) {
+    auto tid = get_tid();
+    auto it = _thread2dev_map.find(tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map[curr_tid] = std::stack<unsigned int>({id});
```
zhiweij1:
```suggestion
      _thread2dev_map.insert({curr_tid, std::stack<unsigned int>({id})});
```

clang/runtime/dpct-rt/include/dpct/device.hpp
zhiweij1: I updated your code like below:
```
#include <cuda.h>
#include <iostream>


#define checkCUDAStatus(CALL) do {\
    cudaError_t Res = CALL;\
    if (Res != cudaSuccess) {\
        printf("Failed at %s:%d, error: %d\n", __FILE__, __LINE__, (int)Res);\
    }\
} while(0)

#define checkCUStatus(CALL) do {\
    CUresult Res = CALL;\
    if (Res != CUDA_SUCCESS) {\
        printf("Failed at %s:%d, error: %d\n", __FILE__, __LINE__, (int)Res);\
    }\
} while(0)

int main() {
  CUdevice  dev;
  CUcontext ctx;

  checkCUStatus(cuInit(0));

  checkCUDAStatus(cudaGetDevice(&dev));

  checkCUStatus(cuCtxGetCurrent(&ctx));
  std::cout << ctx << "\n";

  checkCUStatus(cuCtxCreate(&ctx, 0, dev));
  std::cout << "After create " << ctx << "\n";

  checkCUStatus(cuCtxPushCurrent(ctx));
  std::cout << "After push " << ctx << "\n";

  checkCUStatus(cuCtxPopCurrent(&ctx));
  std::cout << ctx << "\n";

  checkCUStatus(cuCtxPopCurrent(&ctx));
  std::cout << ctx << "\n";

  ctx = 0;

  checkCUStatus(cuCtxPopCurrent(&ctx));
  std::cout << ctx << "\n";

  checkCUStatus(cuCtxPopCurrent(&ctx));
  std::cout << ctx << "\n";

  checkCUStatus(cuDevicePrimaryCtxRetain(&ctx, dev));

  checkCUStatus(cuCtxDestroy(ctx));

  return 0;
}
```
```
zhiweij1@shliclel672:/export/users/zhiweij1/workspace$ nvcc a.cu -lcuda
zhiweij1@shliclel672:/export/users/zhiweij1/workspace$ ./a.out 
0
After create 0x55fd1eda3b30
After push 0x55fd1eda3b30
0x55fd1eda3b30
0x55fd1eda3b30
Failed at a.cu:44, error: 201
0
Failed at a.cu:47, error: 201
0
Failed at a.cu:52, error: 201
zhiweij1@shliclel672:/export/users/zhiweij1/workspace$ 
```
It seems that the cuda API will return a non-zero error code if we try to pop from an empty stack.

zhiweij1: So I agree you can return -1 here.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Done

zhiweij1 (approved):

AndyCHHuang (approved):
LGTM

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +741,31 @@ class dev_mgr {
...
   }
 
+  /// Update the device stack for the current thread id
+  void push_device(unsigned int id) {
+    auto tid = get_tid();
+    auto it = _thread2dev_map.find(tid);
+    if (it == _thread2dev_map.end())
+      _thread2dev_map.insert({tid, std::stack<unsigned int>({id})});
+    else
+      it->second.push(id);
```
ziranzha:
```suggestion
    _thread2dev_map[tid].push(id);
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -662,8 +663,16 @@ class dev_mgr {
   void select_device(unsigned int id) {
     std::lock_guard<std::recursive_mutex> lock(m_mutex);
     check_id(id);
-    _thread2dev_map[get_tid()] = id;
+    auto curr_tid = get_tid();
+
+    // make sure the device stack exist
```
ziranzha: `select_device` shouldn't push current device into the stack

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -778,7 +812,7 @@ class dev_mgr {
   /// for the current thread.
   const unsigned int DEFAULT_DEVICE_ID = 0;
   /// thread-id to device-id map.
-  std::map<unsigned int, unsigned int> _thread2dev_map;
+  std::map<unsigned int, std::stack<unsigned int>> _thread2dev_map;
```
ziranzha:
```suggestion
  static thread_local unsigned _curr_id;
  static thread_local std::stack<unsigned> _dev_stack;
```

clang/test/dpct/driver_context.cu
```diff
@@ -35,7 +35,7 @@ int main(){
   // CHECK: ctx = dpct::select_device(device);
```
ziranzha: Should use push_device here because this API will push the context into the stack.

clang/runtime/dpct-rt/include/dpct/device.hpp
ziranzha: The current ID and stack should be different. Some operations only change the current ID, but don't push/pop the stack.
Can we use the `thread_local` variable to replace the map? Then, we won't need to use mutex in `select_device` or `push_device`.

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +741,31 @@ class dev_mgr {
...
+      _thread2dev_map.insert({tid, std::stack<unsigned int>({id})});
+    else
+      it->second.push(id);
+  }
+
+  /// Remove the device from top of the stack if it exist
+  unsigned int pop_device() {
+    auto it = _thread2dev_map.find(get_tid());
+    if (it == _thread2dev_map.end())
+      return -1;
```
ziranzha: Should throw exception instead of returning `-1`.

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -732,6 +741,31 @@ class dev_mgr {
...
+
+    if (it->second.empty())
+      return -1;
```
ziranzha: Same as above.

ziranzha (changes requested):

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: `cuCtxPopContext` returns as non-zero error code but let the program continue. A sample code is [here](https://github.com/oneapi-src/SYCLomatic/pull/2109#discussion_r1662329017).

Shouldn't we maintain the same behavior?

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Please see the comment above.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Using two thread scope variables is a cleaner. But where do you suggest we initialise the `_curr_id` with `DEFAULT_DEVICE_ID`. We need this initised value to implement `current_device_id()`. Having a std::map (`thread2dev_map`) is handy here.

About the mutex, it is also supposed to prevent race condition for std::vector type variable `_devs`. I came to this conclusion because it's used in `get_device` where there is no access of `thread2dev_map`. Note: `_devs` is written in the ctor `dev_mgr()` and in method `filter()` and read in multiple places.

Does it make sense to also make the `_devs` thread_local? IMO no.

ziranzha: I think `static thread_local unsigned _curr_id = 0;` can work for the initialization.

About the mutex, yes, just remove the used for avoiding race condition of `thread2dev_map`

clang/runtime/dpct-rt/include/dpct/device.hpp
ziranzha: The returned error code can be checked in the CUDA code, but there is no such mechanism in SYCL. So, I suggest throwing the exception.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Now we are only using a device stack for all CUDA context operation as discussed.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: `_dev_stack` is now `thread_local` and `_curr_id` is no longer used and we will use mutex for access to `_devs`.

clang/test/dpct/driver_context.cu
the-slow-one: Done

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Done

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Done

clang/test/dpct/query_api_mapping/Driver/test.cu
```diff
@@ -59,7 +59,7 @@
 // CUDEVICEPRIMARYCTXRETAIN: CUDA API:
 // CUDEVICEPRIMARYCTXRETAIN-NEXT:   cuDevicePrimaryCtxRetain(pc /*CUcontext **/, d /*CUdevice*/);
 // CUDEVICEPRIMARYCTXRETAIN-NEXT: Is migrated to:
-// CUDEVICEPRIMARYCTXRETAIN-NEXT:   *pc = dpct::select_device(d);
+// CUDEVICEPRIMARYCTXRETAIN-NEXT:   *pc = d;
```
the-slow-one:
```suggestion
// CUDEVICEPRIMARYCTXRETAIN-NEXT:   *pc = dpct::select_device(d);
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -648,22 +649,23 @@ class dev_mgr {
...
   /// \param [in] id The id of the device which can
   /// be obtained through get_device_id(const sycl::device).
   void select_device(unsigned int id) {
-    std::lock_guard<std::recursive_mutex> lock(m_mutex);
-    check_id(id);
-    _thread2dev_map[get_tid()] = id;
+    /// Replace the top of the stack with the given device id
+    if (!_dev_stack.empty())
+      pop_device();
+    push_device(id);
```
ziranzha:
```suggestion
    if (_dev_stack.empty()) {
      push_device(id);
    } else {
      check_id(id);
      _dev_stack.top() = id;
    }
```

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -717,7 +719,7 @@ class dev_mgr {
         break;
       }
     }
-    _thread2dev_map.clear();
+    _dev_stack = std::stack<unsigned int>();
```
ziranzha: Only clear the stack of the current thread, other threads' stacks are unchanged.
Should we clear other stacks together or do nothing?

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -768,17 +786,18 @@ class dev_mgr {
 #endif
   }
   void check_id(unsigned int id) const {
+    std::lock_guard<std::recursive_mutex> lock(m_mutex);
```
ziranzha, line 787: For read-only, maybe we can assume the safety.

clang/runtime/dpct-rt/include/dpct/device.hpp
```diff
@@ -768,17 +786,18 @@ class dev_mgr {
...
     }
   }
   std::vector<std::shared_ptr<device_ext>> _devs;
-  /// DEFAULT_DEVICE_ID is used, if current_device_id() can not find current
-  /// thread id in _thread2dev_map, which means default device should be used
-  /// for the current thread.
+  /// stack of devices resulting from CUDA context change;
+  static inline thread_local std::stack<unsigned int> _dev_stack;
+  /// DEFAULT_DEVICE_ID is used, if current_device_id() finds an empty
+  /// _dev_stack, which means the last set device should be used for the current
```
ziranzha:
```suggestion
  /// _dev_stack, which means the default device should be used for the current
```

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: Aren't they both are doing the same thing?

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: `_dev_stack` is `thread_local` and will take care of this issue.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: The existing implementation always gets the mutex lock before calling `check_id`. This mod is in line with the existing implementation.

clang/runtime/dpct-rt/include/dpct/device.hpp
ziranzha: pop and then push will do more things. Integer assignment is much simpler.

clang/runtime/dpct-rt/include/dpct/device.hpp
ziranzha: It means only the `_dev_stack` of the current thread can be accessed. `_dev_stack` of other threads are not updated.

clang/runtime/dpct-rt/include/dpct/device.hpp
ziranzha: Agreed.

clang/runtime/dpct-rt/include/dpct/device.hpp
the-slow-one: At this moment we don't have a way to clear all threads. And I can't find a place where this method is used. Shall we do nothing and leave a comment here?

ziranzha: Agreed.

ziranzha (approved):

PR #2145, opened 2024-07-09
[SYCLomatic] Add support for passing `const` qualified argument to `set_data_ptr`

the-slow-one:
## Issue
In the following source, the line `cpy.srcHost = pSrc` is migrated to `cpy.from.pitched.set_data_ptr(pSrc)`. This fails to compile with _icpx_ as the defination of `set_data_ptr` expects a non-const qualified argument.
```
$ cat src.cu
#include <cstring>
#include <cuda.h>

CUDA_MEMCPY2D get(const void *pSrc) {
  CUDA_MEMCPY2D cpy;
  cpy.srcHost = pSrc;
  return cpy;
}
```
Migrated code by DPCT is as following. _icpx_ fails to compile this code.
```
$ cat dpct_output/src.dp.cpp
#include <sycl/sycl.hpp>
#include <dpct/dpct.hpp>
#include <cstring>

dpct::memcpy_parameter get(const void *pSrc) {
  dpct::memcpy_parameter cpy;
  cpy.from.pitched.set_data_ptr(pSrc);  
  return cpy;
}
```

## Solution
Currently, the argument to `set_data_ptr` is implicitly type-casted to `void *`. Add an explicit type casting to pointer to void. This will also ensure the helper functions are able to use the private member `_data` as per the current implementation.

This PR implements the above solution.

Signed off by: [Deepak Raj H R](deepak.raj.h.r@intel.com)


tangjj11:
I have a PR #2326 to handle this problem. Please consider which one is better solution?

the-slow-one:
Withdrawing PR in the favour of https://github.com/oneapi-src/SYCLomatic/pull/2326

PR #2146, opened 2024-07-09
[SYCLomatic] Fix incorrect typecast of `sycl::malloc` return type in case of CUDA type redefination

the-slow-one:
The following migrated code in DPCT have incompatiable cast operation. This isssue is only relavent on Windows. The sample below only compiles on Windows.

```
// issue.dp.cpp
#include <sycl/sycl.hpp>
#include <dpct/dpct.hpp>
#include <cstdint>

typedef uint64_t CUdeviceptr;

void foo(dpct::device_ptr ptr) {
  ptr = DPCT_CHECK_ERROR(
      DPCT_CHECK_ERROR(ptr = (unsigned long long)sycl::malloc_device(
                           1024, dpct::get_in_order_queue())));
}
```

Orginal CUDA code is
```
#include <cstdint>
#include <cuda.h>

typedef uint64_t CUdeviceptr;

void foo(CUdeviceptr ptr) {
  ptr = cuMemAlloc(&ptr, 1024);
}
```

Solution is to stop the underlying type lookup if the type is a defined by CUDA. (Existing implementation only looks if the source of defination is cuda.h but it's not sufficent as evident in the CUDA example above)


AndyCHHuang (approved):
Implementation LGTM.
Please add a lit test.

AndyCHHuang (approved):

clang/test/dpct/cu_mem_alloc_typedef.cu
```diff
@@ -0,0 +1,11 @@
+// UNSUPPORTED: system-linux
+// RUN: dpct --format-range=none -out-root %T/cu_mem_alloc_typedef %s --cuda-include-path="%cuda-path/include" -- -std=c++14 -x cuda --cuda-host-only
+// RUN: FileCheck %s --match-full-lines --input-file %T/cu_mem_alloc_typedef/cu_mem_alloc_typedef.dp.cpp
+#include <cstdint>
+#include <cuda.h>
+
+typedef uint64_t CUdeviceptr;
+
+void foo(CUdeviceptr ptr) {
+  ptr = cuMemAlloc(&ptr, 1024);
```
the-slow-one:
```suggestion
  // CHECK: ptr = DPCT_CHECK_ERROR(DPCT_CHECK_ERROR(ptr = (dpct::device_ptr)sycl::malloc_device(1024, dpct::get_in_order_queue())));
  ptr = cuMemAlloc(&ptr, 1024);
```

zhiweij1:
@the-slow-one 
The test case has syntax error, it cannot be compiled by nvcc:
```
zhiwei@zhiwei-ubuntu:~/testfolder$ nvcc test.cu -c
test.cu(4): error: invalid redeclaration of type name "CUdeviceptr" (declared at line 262 of /usr/local/cuda-12.4/bin/../targets/x86_64-linux/include/cuda.h)
  typedef uint64_t CUdeviceptr;
                   ^

test.cu(7): error: argument of type "CUdeviceptr *" is incompatible with parameter of type "CUdeviceptr *"
    ptr = cuMemAlloc_v2(&ptr, 1024);
                        ^

2 errors detected in the compilation of "test.cu".
zhiwei@zhiwei-ubuntu:~/testfolder$ 
```

zhiweij1 (approved):

the-slow-one:
> @the-slow-one The test case has syntax error, it cannot be compiled by nvcc:
> 
> ```
> zhiwei@zhiwei-ubuntu:~/testfolder$ nvcc test.cu -c
> test.cu(4): error: invalid redeclaration of type name "CUdeviceptr" (declared at line 262 of /usr/local/cuda-12.4/bin/../targets/x86_64-linux/include/cuda.h)
>   typedef uint64_t CUdeviceptr;
>                    ^
> 
> test.cu(7): error: argument of type "CUdeviceptr *" is incompatible with parameter of type "CUdeviceptr *"
>     ptr = cuMemAlloc_v2(&ptr, 1024);
>                         ^
> 
> 2 errors detected in the compilation of "test.cu".
> zhiwei@zhiwei-ubuntu:~/testfolder$ 
> ```

This is a Windows-only test. `nvcc` compiles this file on Windows.

zhimingwang36 (approved):

clang/test/dpct/cu_mem_alloc_typedef.cu
```diff
@@ -0,0 +1,12 @@
+// UNSUPPORTED: system-linux
+// RUN: dpct --format-range=none -out-root %T/cu_mem_alloc_typedef %s --cuda-include-path="%cuda-path/include" -- -std=c++14 -x cuda --cuda-host-only
+// RUN: FileCheck %s --match-full-lines --input-file %T/cu_mem_alloc_typedef/cu_mem_alloc_typedef.dp.cpp
+#include <cstdint>
+#include <cuda.h>
+
+typedef uint64_t CUdeviceptr;
```
zhimingwang36, line 7: CUdeviceptr is user user-defined type, in L9 and L11,  DPCT should keep it without migration.  (As it is user-defined type).

clang/lib/DPCT/CallExprRewriterCommon.h
```diff
@@ -792,7 +792,9 @@ inline std::function<std::string(const CallExpr *C)> getDerefedType(size_t Idx)
       DerefQT = ET->getNamedType();
       if (const auto *TDT = dyn_cast<TypedefType>(DerefQT)) {
         auto *TDecl = TDT->getDecl();
-        if (dpct::DpctGlobalInfo::isInCudaPath(TDecl->getLocation()))
+        if (dpct::DpctGlobalInfo::isInCudaPath(TDecl->getLocation()) ||
```
zhiweij1: How about using `TDecl->redecls()` to get all decls.
If one of the decl is in cuda folder, then break.
Is it better?

clang/test/dpct/cu_mem_alloc_typedef.cu
zhimingwang36: can we find all the other typedefs?  (like the typedef in cuda header file). If so, we can migrate it..

the-slow-one: If there is a typedef and it's present in CUDA headers, and if nvcc can compile the code -- it mean the underlying types is same everywhere. I think it will be redundant to look for the type in CUDA header as we rely on the fact that for a successful migration the CUDA source should be compilable. Thoughts?

clang/lib/DPCT/CallExprRewriterCommon.h
```diff
@@ -792,7 +792,9 @@ inline std::function<std::string(const CallExpr *C)> getDerefedType(size_t Idx)
...
+            MapNames::TypeNamesMap.find(TDecl->getNameAsString()) !=
+                MapNames::TypeNamesMap.end())
           break;
```
zhiweij1:
```suggestion
        bool IsFromCuda = false;
        for (const auto& D : TDecl->redecls())
           if (dpct::DpctGlobalInfo::isInCudaPath(TDecl->getLocation())) {
             IsFromCuda = true;
             break;
           }
        if (IsFromCuda)
          break;
```

zhiweij1 (approved):

zhiweij1 (approved):

clang/test/dpct/cu_mem_alloc_typedef.cu
```diff
@@ -0,0 +1,12 @@
...
+
+void foo(CUdeviceptr ptr) {
+  // CHECK: ptr = DPCT_CHECK_ERROR(DPCT_CHECK_ERROR(ptr = (dpct::device_ptr)sycl::malloc_device(1024, dpct::get_in_order_queue())));
```
zhimingwang36: two DPCT_CEHCK_ERROR here?

PR #2147, opened 2024-07-09
[SYCLomatic] Guard std min/max from expanding into a min/max macro on Windows

the-slow-one:
`icpx` fails to compile the following code on Windows

```
#include <windows.h>
#include <dpct/dpct.hpp>
int main() { 
  return 0;
}
```
The compile error indicates the min/max marco expansion (defined in winmindef.h) in DPCT runtime headers.

This PR fixes the issue by preventing preprocessing of min/max using parentheses.


AndyCHHuang (approved):
LGTM

zhiweij1 (approved):

zhimingwang36 (approved):

PR #2166, opened 2024-07-15
[SYCLomatic][CMake]Add test for try_compile and try_run

the-slow-one:
Migration of CUDA source files in `try_compile` and `try_run` is already supported in DPCT. This PR adds test.


tomflinda (approved):
LGTM

TejaX-Alaghari (approved):
LGTM`

PR #2167, opened 2024-07-15
[NFC][SYCLomatic] Cleanup build warnings

yihwang-nv:
```
clang/lib/Lex/InitHeaderSearch.cpp:524:53: warning: ‘bool llvm::StringRef::equals(llvm::StringRef) const’ is deprecated: Use == instead [-Wdeprecated-declarations]
  524 |     if (Entry.Lookup.getName().equals("/usr/include")) {
      |                                                     ^

clang/lib/DPCT/Diagnostics.cpp:105:33: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  105 |       if (*Cur == '\0' && Begin < End) {
      |                           ~~~~~~^~~~~

llvm/lib/Support/CommandLine.cpp:2444:5: warning: default label in switch which covers all enumeration values [-Wcovered-switch-default]
    default:
    ^
```

### DIFF
--- a/clang/lib/DPCT/Diagnostics.cpp
+++ b/clang/lib/DPCT/Diagnostics.cpp
@@ -94,7 +94,7 @@ void initWarningIDs() {
         reportInvalidWarningID(ID);
       return Value;
     };
-    int Begin = ParseNumber();
+    auto Begin = ParseNumber();
 
     if (*Cur == '\0') {
       WarningIDs.insert(Begin);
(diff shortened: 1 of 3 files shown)


clang/lib/DPCT/Diagnostics.cpp
```diff
@@ -94,7 +94,7 @@ void initWarningIDs() {
         reportInvalidWarningID(ID);
       return Value;
     };
-    int Begin = ParseNumber();
+    unsigned Begin = ParseNumber();
```
tangjj11: Can use `auto`?

tangjj11 (approved):

clang/lib/DPCT/Diagnostics.cpp
yihwang-nv: Done.

zhimingwang36 (approved):

PR #2176, opened 2024-07-16
[SYCLomatic] Weekly pulldown from origin/sycl to origin/SYCLomatic 20240715

yihwang-nv:


clang/include/clang/AST/Decl.h
```diff
@@ -3054,7 +3054,6 @@ class FunctionDecl : public DeclaratorDecl,
   /// computed and stored.
   unsigned getODRHash() const;
 
-<<<<<<< HEAD
 #ifdef SYCLomatic_CUSTOMIZATION
```
tangjj11: Need handle some github action

clang/lib/AST/TypePrinter.cpp
```diff
@@ -1688,9 +1687,6 @@ void TypePrinter::printTemplateId(const TemplateSpecializationType *T,
   else
 #endif // SYCLomatic_CUSTOMIZATION
   // FIXME: Null TD never excercised in test suite.
```
tangjj11: Should we apply the change? `// FIXME: Null TD never exercised in test suite.`

tangjj11 (changes requested):

clang/lib/Parse/ParseExprCXX.cpp
```diff
@@ -3030,23 +3029,13 @@ bool Parser::ParseUnqualifiedId(CXXScopeSpec &SS, ParsedType ObjectType,
           SS, ObjectType, ObjectHadErrors,
           TemplateKWLoc ? *TemplateKWLoc : SourceLocation(), Id, IdLoc,
           EnteringContext, Result, TemplateSpecified);
+    else if (TemplateSpecified &&
```
tangjj11: These revert code do not have `SYCLomatic` tag, and will conflict once llvm or sycl touch these code. Can we just modify the core logic that cause the problem?

tangjj11 (approved):

zhimingwang36:
LGTM

clang/lib/Parse/ParseExprCXX.cpp
ziranzha: `SYCLomatic_CUSTOMIZATION` is needed here to mark the difference with clang main trunk.

ziranzha (approved):

PR #2215, opened 2024-08-01
[SYCLomatic] [stable_]partition performance improvement

danhoeflinger:
Performance improvement for in-place `partition` and `stable_partition`.  

This allows us to avoid the extra copy of the mask, and reduce the number of kernels.  We rely internally upon `[stable_]partition_copy` now instead of `stable_partition` which internally does a partion_copy followed by a copy.  This allows us to avoid writing to the mask list without allocating a tmp mask buffer and copying to it.  

The change for `partition` reverses the "false" list, so it breaks stability.  Stability is not required for this API though, and we can skip one copy kernel this way.

It would be possible to write a specialized sycl copy kernel which copies two different sequences into an output, but for now we just call copy twice.

Requires https://github.com/oneapi-src/SYCLomatic-test/pull/752 to be merged first for passing tests (relaxing stability requirement of `partition` test).

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
@@ -19,6 +19,35 @@
 
 namespace dpct {
 
+namespace internal {
+// This function is ported from oneDPL with the check for an FPGA policy
+// removed. This function should be used to wrap a provided policy when multiple
+// oneDPL calls are made to ensure unique kernel names.
+template <template <typename> class NewKernelName, typename Policy>
+auto make_wrapped_policy(Policy &&policy)
+    -> decltype(oneapi::dpl::execution::make_device_policy<
+                NewKernelName<typename ::std::decay_t<Policy>::kernel_name>>(
+        ::std::forward<Policy>(policy))) {
+  return oneapi::dpl::execution::make_device_policy<
+      NewKernelName<typename ::std::decay_t<Policy>::kernel_name>>(
+      ::std::forward<Policy>(policy));
+}
+
+template <typename Name> class partition_call1;
+
+template <typename Name> class partition_call2;
+
+template <typename Name> class copy_before_partition;
+
+template <typename Name> class reverse_partition;
+
+template <typename Name> class copy_call1;
+
+template <typename Name> class copy_call2;
+
+}; // namespace internal
+
+
 template <typename Policy, typename Iter1, typename Iter2, typename Pred,
           typename T>
 void replace_if(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p,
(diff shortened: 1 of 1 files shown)


clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -978,7 +1018,41 @@ stable_partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
...
+                            typename std::decay<Policy>::type>::value,
+                        Iter1>::type
+partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
+  static_assert(
+      std::is_same<typename std::iterator_traits<Iter1>::iterator_category,
+                   std::random_access_iterator_tag>::value &&
+          std::is_same<typename std::iterator_traits<Iter2>::iterator_category,
+                       std::random_access_iterator_tag>::value,
+      "Iterators passed to algorithms must be random-access iterators.");
+  typedef typename std::decay<Policy>::type policy_type;
```
mmichel11: I don't think this line is used anywhere.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -937,18 +966,29 @@ stable_partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
...
       "Iterators passed to algorithms must be random-access iterators.");
   typedef typename std::decay<Policy>::type policy_type;
-  internal::__buffer<typename std::iterator_traits<Iter1>::value_type> _tmp(
-      std::distance(first, last));
+  auto _partition_call1 =
+      internal::make_wrapped_policy<internal::partition_call1>(policy);
+  auto _copy_call1 =
+      internal::make_wrapped_policy<internal::copy_call1>(policy);
+  auto _copy_call2 =
+      internal::make_wrapped_policy<internal::copy_call2>(policy);
```
mmichel11: Can we forward `policy` the last time we use it?

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -978,7 +1018,41 @@ stable_partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
...
+      std::is_same<typename std::iterator_traits<Iter1>::iterator_category,
+                   std::random_access_iterator_tag>::value &&
+          std::is_same<typename std::iterator_traits<Iter2>::iterator_category,
+                       std::random_access_iterator_tag>::value,
+      "Iterators passed to algorithms must be random-access iterators.");
+  typedef typename std::decay<Policy>::type policy_type;
+  auto _partition_call1 =
+      internal::make_wrapped_policy<internal::partition_call1>(policy);
+  auto _copy_call1 =
+      internal::make_wrapped_policy<internal::copy_call1>(policy);
```
mmichel11: Same comment as above regarding forwarding `policy`.

mmichel11 (commented):
Only minor comments from my side.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: fixed

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: fixed

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: fixed

mmichel11 (approved):
LGTM

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -937,18 +966,29 @@ stable_partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
...
-  return std::get<0>(ret_val.base());
+  auto _tmp_first = _tmp.get();
+  auto _tmp_last = _tmp_first + _n;
+  auto _tmp_last_rev = std::reverse_iterator(_tmp_last);
+
+  auto _end_pair =
+      stable_partition_copy(std::move(_partition_call1), first, last, mask,
+                            _tmp_first, _tmp_last_rev, p);
+  auto _first_n = std::distance(_tmp_first, std::get<0>(_end_pair));
+  std::copy(std::move(_copy_call1), _tmp_first, std::get<0>(_end_pair), first);
```
ziranzha: How about using this function's return value instead of `first + _first_n` to avoid computing the distance?

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -978,7 +1018,41 @@ stable_partition(Policy &&policy, Iter1 first, Iter1 last, Iter2 mask, Pred p) {
...
+
+  auto _tmp_first = _tmp.get();
+  auto _tmp_last = _tmp_first + _n;
+  auto _tmp_last_rev = std::reverse_iterator(_tmp_last);
+
+  auto _end_pair =
+      stable_partition_copy(std::move(_partition_call1), first, last, mask,
+                            _tmp_first, _tmp_last_rev, p);
+  auto _first_n = std::distance(_tmp_first, std::get<0>(_end_pair));
+  std::copy(std::move(_copy_call1), _tmp_first, _tmp_last, first);
```
ziranzha, line 1047: Same as above.

tomflinda (approved):
LGTM

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: Thanks, this is a good suggestion.  Fixed.

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: I don't think this is possible in this one, because we only have one copy call, which covers both parts of the partitioned result.  (This is what breaks the stability, because we are copying in a reversed second part).

I may be missing something, but I think we need the distance call and addition in this version.

ziranzha: I make a mistake here. You are right.

ziranzha (approved):

PR #2266, opened 2024-08-22
[SYCLomatic][CMake] Enable warning msg framework to the remove YAML rules which may have some side effects

tomflinda:

### DIFF
--- a/clang/lib/DPCT/DiagnosticsCMakeScriptMigration.inc
+++ b/clang/lib/DPCT/DiagnosticsCMakeScriptMigration.inc
@@ -9,4 +9,4 @@
 // FORMAT: {NAME, ID, MESSAGE}
 DEF_COMMENT(CMAKE_NOT_SUPPORT_WARNING, 3000, "Migration of syntax \"{0}\" is not supported. You may need to adjust the code.")
 DEF_COMMENT(CMAKE_CONFIG_FILE_WARNING, 3001, "\"{0}\" is used to copy an input file to an output file while performing transformations of the input file content. You may need to update the name of the input and output file if the input file contains CUDA syntax, and make sure the input file exists by copying or migrating the original input file from in-root folder.")
-DEF_COMMENT(WARNING_FOR_SYNTAX_REMOVED, 3002, "{0}")
+DEF_COMMENT(WARNING_FOR_SYNTAX_REMOVED, 3002, "\"{0}\" is removed because it is not supported to be used together with option \"-fsycl\" of SYCL compiler. You may need to check the build options.")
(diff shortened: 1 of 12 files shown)


clang/lib/DPCT/PatternRewriter.cpp
```diff
@@ -719,8 +728,50 @@ void setFileTypeProcessed(enum SourceFileType FileType) {
...
+      break;
+    }
+
+    Count += Line.size() + 1;
+    LineNumber++;
+  }
+
+  if (Warning.size() >= 2 && Warning.front() == '"' && Warning.back() == '"') {
+    Warning = Warning.substr(1, Warning.size() - 2);
+  }
```
tomflinda: Not needed, will remove it in the next PR.

zhiweij1 (approved):

clang/lib/DPCT/PatternRewriter.cpp
```diff
@@ -719,8 +728,50 @@ void setFileTypeProcessed(enum SourceFileType FileType) {
   SrcFileType = FileType;
 }
 
+static void constructWaringMsg(const std::string &Input, size_t index,
+                               const std::string &FileName,
+                               const std::string &FrontPart,
```
AndyCHHuang, line 733: instead of passing frontpart as a string, can we pass the line number of frontpart instead?

AndyCHHuang (approved):
LGTM
the comment is optional

clang/lib/DPCT/PatternRewriter.cpp
tomflinda: Yes, we can. Passing the Frontpart string makes it easy to identify the root cause of a possible incorrect line number. So, I suggest keeping the current style.

clang/lib/DPCT/DiagnosticsCMakeScriptMigration.inc
```diff
@@ -9,3 +9,4 @@
 // FORMAT: {NAME, ID, MESSAGE}
 DEF_COMMENT(CMAKE_NOT_SUPPORT_WARNING, 3000, "Migration of syntax \"{0}\" is not supported. You may need to adjust the code.")
 DEF_COMMENT(CMAKE_CONFIG_FILE_WARNING, 3001, "\"{0}\" is used to copy an input file to an output file while performing transformations of the input file content. You may need to update the name of the input and output file if the input file contains CUDA syntax, and make sure the input file exists by copying or migrating the original input file from in-root folder.")
+DEF_COMMENT(WARNING_FOR_SYNTAX_REMOVED, 3002, "{0}")
```
zhimingwang36: it's not clear for the 3002 here, please move the message body define to here from the Yaml rule.

clang/test/dpct/cmake_migration/case_041/expected.txt
```diff
@@ -11,7 +11,7 @@ target_link_options( quda PUBLIC )
 target_link_options(quda PUBLIC  )
 
 target_link_options(nvcv_util_compat PUBLIC
-    
+    # DPCT3002:1: '-static-libstdc++' is not supported together with '-fsycl' when using icpx compiler. You may need to manually fix it.
```
zhimingwang36:
```suggestion
    # DPCT3002:1: '-static-libstdc++' is removed because it is not supported to be used together with option '-fsycl' of SYCL compiler. You may need to check the build options.
```

clang/lib/DPCT/DiagnosticsCMakeScriptMigration.inc
tomflinda: Okay

clang/test/dpct/cmake_migration/case_041/expected.txt
tomflinda: Okay

PR #2283, opened 2024-08-27
[SYCLomatic] Fix compile error of auto-generated Makefile for macros defined for string directedly used in printf() and std::cout()

tomflinda:

### DIFF
--- a/clang/lib/DPCT/GenMakefile.cpp
+++ b/clang/lib/DPCT/GenMakefile.cpp
@@ -286,14 +286,14 @@ static void getCompileInfo(
         IsSystemInclude = true;
       } else if (llvm::StringRef(Option).starts_with("-D")) {
         // Parse macros defined.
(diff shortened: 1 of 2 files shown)


clang/lib/DPCT/GenMakefile.cpp
```diff
@@ -286,18 +286,32 @@ static void getCompileInfo(
         IsSystemInclude = true;
       } else if (llvm::StringRef(Option).starts_with("-D")) {
         // Parse macros defined.
+        bool NeedSigleQuote = false;
```
zhiweij1:
```suggestion
        bool NeedSingleQuote = false;
```

tomflinda: Good catch, will update it in the next PR.

zhiweij1 (approved):

zhimingwang36 (approved):

PR #2287, opened 2024-08-28
[SYCLomatic][CMake] Add new match mode "StrictFull" to fix full match not work when '.' follows the matched string

tomflinda:


zhiweij1 (approved):

clang/lib/DPCT/PatternRewriter.cpp
```diff
@@ -481,106 +473,53 @@ static std::optional<MatchResult> findFullMatch(const MatchPattern &Pattern,
...
-      }
-
-      std::string ElementContents = Input.substr(Index, Next - Index);
-      if (SrcFileType == SourceFileType::SFT_CMakeScript) {
-        if (Code.Name == "empty" && !ElementContents.empty() &&
-            ElementContents.find_first_not_of(' ') != std::string::npos) {
-          // For reversed variable ${empty}, it should be empty string or string
-          // only including spaces.
+      if (Mode == RuleMatchMode::Full) {
+        if (PatternIndex == 0 && Index - 1 >= 0 &&
```
AndyCHHuang: the code is similar to full mode can we extract it to a function?

AndyCHHuang (approved):

clang/lib/DPCT/Rules.h
```diff
@@ -28,7 +28,7 @@ enum RuleKind {
 };
 
 enum RulePriority { Takeover, Default, Fallback };
-enum RuleMatchMode { Partial , Full };
+enum RuleMatchMode { Partial, Full, RawStringMatch };
```
zhiweij1: How about
```suggestion
enum RuleMatchMode { Partial, Full, StrictFull};
```
?

clang/lib/DPCT/PatternRewriter.cpp
tomflinda: Okay, will update it in the next update.

clang/lib/DPCT/Rules.h
tomflinda: okay

AndyCHHuang (approved):
LGTM

zhimingwang36 (approved):

PR #2333, opened 2024-09-06
Add requirements.txt to list python script dependency.

jiejanezhang:
Add requirements.txt to list python script dependency.


intwanghao (approved):
LGTM

requirements.txt
```diff
@@ -0,0 +1 @@
+graphviz==0.20.3
```
zhimingwang36, line 1: Should this file be added into install folder?  (If so, need to update the install target to include it).

zhimingwang36, line 1: can other version of graphviz also works?

jiejanezhang: Feedback from Sergey is that it is better to have explicit version and maintain it up-to-date manually.

jiejanezhang: Yes.

zhimingwang36 (approved):
LGTM

PR #2350, opened 2024-09-10
[SYCLomatic]Remove reverse_iterator from partition implementations

danhoeflinger:
Partition implementations should not use `std::reverse_iterator`, because this disallows usage of sycl buffer iterator-like inputs returned from `oneapi::dpl::[begin,end]`.  MSVC STL runs into compiler errors for reverse_iterator of this iterator-like type.

Without a reverse iterator, we no longer can gain advantage from providing a non-stable partition implementation, so we go back to having a unified implementation (these were separated in #2215).  If in the future oneDPL provides a reverse_iterator which is compatible with sycl buffer iterator-like types, we can improve this implementation by removing this extra copy call, and extra tmp data allocation.  However, for now I think this is the best we can do.


timmiesmith (approved):
LGTM. Thank you.

mmichel11 (approved):
LGTM. We may want to create an issue somewhere about this problem if we want to consider adding a reverse iterator to oneDPL in the future.

danhoeflinger:
> LGTM. We may want to create an issue somewhere about this problem if we want to consider adding a reverse iterator to oneDPL in the future.

https://github.com/oneapi-src/oneDPL/issues/1836

tomflinda (approved):
LGTM

PR #2373, opened 2024-09-23
[SYCLomatic] Remove 'using namespace cub;' directive

yihwang-nv:
Remove the `using namespace cub;` directive.
```
#include <cub/cub.cuh>
#include <cuda_runtime.h>

using namespace std;
using namespace cub;

__global__ void foo(Sum sum) {}
int main() {
  Sum sum;
  foo<<<1, 1>>>(sum);
  return 0;
}
```


intwanghao (approved):
LGTM

zhimingwang36 (approved):

PR #2484, opened 2024-11-12
[SYCLomatic]Asm instr sub.f16x2 , ld.volatile.global/st.volatile.global

abhilash1910:
WIP for other asms
cc @tomflinda @zhimingwang36


clang/test/dpct/asm/sub.cu
```diff
@@ -42,21 +43,33 @@ __global__ void sub() {
...
+  // CHECK: f16x2 = fa - sycl::half2{1, 1};
+  asm("sub.f16x2 %0, %1, {1, 1};" : "=r"(f16x2) : "r"(fa));
+  
   // CHECK: u16x2 = ua - sycl::ushort2{1, 1};
   asm("sub.u16x2 %0, %1, {1, 1};" : "=r"(u16x2) : "r"(ua));
 
   // CHECK: s16x2 = sycl::short2{1, 1} - sa;
   asm("sub.s16x2 %0, {1, 1}, %1;" : "=r"(s16x2) : "r"(sa));
 
+  // CHECK: f16x2 = sycl::float2{1, 1} - fa;
```
tomflinda:
```suggestion
  // CHECK: f16x2 = sycl::half2{1, 1} - fa;
```

tomflinda: Current migrated result is incorrect, I have provided the suggestion above.

abhilash1910: Thanks

clang/test/dpct/asm/ld.cu
```diff
@@ -8,7 +8,7 @@
 #include <cuda_runtime.h>
 
 /*
-.ss =                       { .const, .global, .local, .param, .shared };
+.ss =                       { .const, .global, .local, .param, .shared, .volatile.global };
```
tomflinda: According to the PTX spec: https://docs.nvidia.com/cuda/parallel-thread-execution/#data-movement-and-conversion-instructions-ld
The .ss field for the ld instruction does not contain the ".volatile.global" value; it is invalid here; pls do a further investigation.

clang/test/dpct/asm/ld.cu
```diff
@@ -27,6 +27,11 @@ __global__ void ld(int *arr) {
   asm volatile ("ld.global.u32 %0, [%1 + 4];" : "=r"(c) : "l"(arr));
   // CHECK: d = *((uint64_t *)((uintptr_t)arr + 8));
   asm volatile ("ld.global.u64 %0, [%1 + 8];" : "=l"(d) : "l"(arr));
+  // CHECK: a = *arr;
+  asm volatile ("ld.volatile.global.s32 %0, [%1];" : "=r"(a) : "l"(arr));
+  // CHECK: b = *((uint32_t *)(uintptr_t)arr);
+  asm volatile ("ld.volatile.global.u32 %0, [%1];" : "=r"(b) : "l"(arr));
+  // CHECK: c = *((uint32_t *)((uintptr_t)arr + 4));
```
tomflinda, line 34: Could you provide the real-use scenario for these two ASM instructions?

clang/lib/DPCT/RulesAsm/Parser/AsmTokenKinds.def
```diff
@@ -278,6 +278,7 @@ STATE_SPACE(reg,    ".reg")
 STATE_SPACE(sreg,   ".sreg")
 STATE_SPACE(const,  ".const")
 STATE_SPACE(global, ".global")
+STATE_SPACE(volatile, ".volatile")
```
tomflinda: According to the value range of .ss in https://docs.nvidia.com/cuda/parallel-thread-execution/#data-movement-and-conversion-instructions-ld,  .volatile is not a ss value, should not be added here.

abhilash1910: Added as a modifier. As per docs it is a qualifier.

abhilash1910:
Closing as need to be reimplemented- @tomflinda

PR #2498, opened 2024-11-19
[SYCLomatic] Avoid migration of non-CUDA type "half"

the-slow-one:
Problem: `half` defined in a location outside of source under investigation and outside of CUDA header is migrated to `sycl::half`.

This PR prevents such migration.


clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1392,7 +1392,7 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
           return;
 
         // When a CUDA type is redefined in the files under analysis we
-        // want to migrate them.
+        // don't want to migrate them.
```
zhiweij1: The original comment is correct. We need to migrate redeclaration of CUDA type.

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1753,11 +1758,15 @@ void VectorTypeNamespaceRule::runRule(const MatchFinder::MatchResult &Result) {
     if (TL->getBeginLoc().isInvalid())
       return;
 
-    // To skip user-defined type.
+    // To skip user-defined type (found in in-root and from third party includes outside of in-root.
```
zhiweij1:
```suggestion
    // To skip user-defined type (found in in-root and from third party includes outside of in-root)
```

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1540,6 +1541,10 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
           return;
         }
       } else if (NTL.getTypeLocClass() == clang::TypeLoc::Record) {
+        if (!DpctGlobalInfo::isInCudaPath(NTL.getBeginLoc())) {
+            return;
```
zhiweij1, line 1545:
```suggestion
          return;
```

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1522,7 +1522,7 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
       auto ETC = TL->getAs<ElaboratedTypeLoc>();
       auto NTL = ETC.getNamedTypeLoc();
 
-      if (const auto *RD = NTL.getType().getCanonicalType()->getAsRecordDecl())
+      if (const auto *RD = NTL.getType().getCanonicalType()->getAsRecordDecl()) {
```
zhiweij1:
```suggestion
      if (const auto *RD = NTL.getType().getCanonicalType()->getAsRecordDecl())
```

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1531,6 +1531,7 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
             if (const auto *Init = VarD->getInit())
               if (const auto *VarInitExpr = dyn_cast<InitListExpr>(Init))
                 emplaceTransformation(new ReplaceStmt(VarInitExpr, "{}"));
+      }
```
zhiweij1:
```suggestion
```

clang/lib/DPCT/RulesInclude/InclusionHeaders.cpp
```diff
@@ -299,4 +299,4 @@ DpctInclusionHeadersMap::DpctInclusionHeadersMapInitializer
     DpctInclusionHeadersMap::Initializer;
 
 } // namespace dpct
-} // namespace clang
\ No newline at end of file
+} // namespace clang
```
zhiweij1: Please do not touch unrelated file.

clang/lib/DPCT/RulesLang/RulesLang.cpp
zhiweij1: How about add
```
      } else if (!isTypeInCUDAScope(TypePtr)) {
        return;
      }
```
after line 1399?

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -1540,6 +1541,10 @@ void TypeInDeclRule::runRule(const MatchFinder::MatchResult &Result) {
           return;
         }
       } else if (NTL.getTypeLocClass() == clang::TypeLoc::Record) {
+        if (!DpctGlobalInfo::isInCudaPath(NTL.getBeginLoc())) {
```
zhiweij1: How about changing at line 1402? (See above comment)

zhiweij1 (changes requested):
Please add a lit test.

clang/lib/DPCT/RulesLang/RulesLang.cpp
the-slow-one: The "half" is outside the analysis scope. A test is added now which will make this clearer. At L1378 there is a check `isTypeInAnalysisScope`.

But I do get your point! Ideally, we should perform this check for all types declared in CUDA header and exit this method early. I will try this and see what break.

zhiweij1: I mean the `else if (!isTypeInCUDAScope(TypePtr))` is added after the block of `if (isTypeInAnalysisScope(TypePtr)) { ... }`.

the-slow-one: Done

zhiweij1 (approved):

TejaX-Alaghari (approved):
LGTM

PR #2516, opened 2024-11-27
Added support for more torch.cuda APIs

TejaX-Alaghari:
Added support for below torch.cuda APIs

- from torch import cuda
- torch.cuda.device_count()
- torch.cuda.get_device_capability()
- torch.cuda.get_device_capability(int)
- torch.cuda.get_arch_list()
- torch.version.cuda

### DIFF
--- a/clang/test/dpct/python_migration/case_006/expected.py
+++ b/clang/test/dpct/python_migration/case_006/expected.py
@@ -1,0 +1,14 @@
+from torch import xpu
+
+devs = torch.xpu.device_count()
+devs = xpu.device_count()
+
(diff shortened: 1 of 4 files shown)


clang/test/dpct/python_migration/case_006/expected.py
```diff
@@ -0,0 +1,14 @@
+from torch import xpu
```
tomflinda, line 1: Do we need to import  intel_extension_for_pytorch , when using xpu device?

zhimingwang36 (approved):

tomflinda (approved):

PR #2530, opened 2024-12-02
[Py Build Script][Bug] Minor bug fixes for Python script migration

TejaX-Alaghari:
This PR introduces below changes to Py build script migration
1. Change **cpp** & **cu** file extension names based on HasCudaSyntax attr of MainSourceFiles.yaml
2. Fix skipped file extension replacements bug when \" (double-quotes) are used
3. Skip single line comments

### DIFF
--- a/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
+++ b/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
@@ -271,7 +271,8 @@ static int parseCodeElement(const MatchPattern &Suffix,
   const int Size = Input.size();
   while (Index >= 0 && Index < Size) {
 
-    if (SrcFileType == SourceFileType::SFT_CMakeScript) {
+    if (SrcFileType == SourceFileType::SFT_CMakeScript ||
+        SrcFileType == SourceFileType::SFT_PySetupScript) {
       if (Input[Index] == '#') {
         for (; Index < Size && Input[Index] != '\n'; Index++) {
         }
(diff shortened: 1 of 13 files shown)


zhiweij1 (approved):

clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
```diff
@@ -280,8 +281,10 @@ static int parseCodeElement(const MatchPattern &Suffix,
     }
 
     const auto Character = Input[Index];
-    if(Suffix.size() == 0 && Character =='"') {
-      return Index;
+    if (SrcFileType != SourceFileType::SFT_PySetupScript) {
```
tomflinda, line 284: Pls explain why  line285~line287 are skipped for SourceFileType::SFT_PySetupScript?

TejaX-Alaghari: Because that condition results in a failure to parse file names for Python syntax migration scenario
`f(["a.cu"])`

Where, parsing [...] in PatternRewriter results in `parseCodeElement()` call with empty suffix when the next element is `\"`

[From here:](https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp#L305)
image
[To here:](https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp#L255)
image

TejaX-Alaghari: If possible, can you please advice any other better way to do it
Or explain more about that code block

tomflinda (approved):
LGTM

PR #2541, opened 2024-12-04
[Bug][Help] Fixed helper options printing for pattern-rewriter & dpct

TejaX-Alaghari:
This PR enables correct command line help options to be printed for each binary (dpct/ pattern-rewriter)

**DPCT:**
image

**Pattern Rewriter:**
image

### DIFF
--- a/clang/tools/pattern-rewriter/src/driver.cpp
+++ b/clang/tools/pattern-rewriter/src/driver.cpp
@@ -113,18 +113,37 @@ static std::string fixLineEndings(const std::string &Input) {
   return OutputStream.str();
 }
 
+#ifdef SYCLomatic_CUSTOMIZATION
+llvm::cl::OptionCategory &getPatReCategory() {
+  static llvm::cl::OptionCategory PatReCategory("Pattern Rewriter options");
+  return PatReCategory;
+}
+#endif
+
 int main(int argc, char *argv[]) {
-  llvm::cl::opt<std::string> InputFilename(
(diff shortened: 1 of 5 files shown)


clang/tools/pattern-rewriter/src/driver.cpp
```diff
@@ -113,18 +113,37 @@ static std::string fixLineEndings(const std::string &Input) {
   return OutputStream.str();
 }
 
+#ifdef SYCLomatic_CUSTOMIZATION
```
zhiweij1: I think this file is not from llvm. So this mark can be removed.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2453,9 +2468,14 @@ class HelpPrinter {
     sortSubCommands(GlobalParser->RegisteredSubCommands, Subs);
 
 #ifdef SYCLomatic_CUSTOMIZATION
-    if (helpCatEnum == HelpCategory::HC_Examples) {
-      outs() << DPCTExamplesMsg;
-      return;
+    isDPCT = !strcmp((GlobalParser->ProgramName).c_str(), "dpct") ||
+             !strcmp((GlobalParser->ProgramName).c_str(), "c2s");
```
zhiweij1: We can define `IsDPCT` as a global variable with init value `false` in this file. And set its value to true in `runDPCT` function. Is it better?

zhiweij1 (approved):

zhimingwang36 (approved):

PR #2596, opened 2025-01-03
[SYCLomatic] Bugfix in non-trivial run length encode's usage of oneDPL's reduce-by-segment

mmichel11:
The current non-trivial run length encode implementation is dependent on an internal implementation detail in oneDPL. The defined operator (named `op`) in the `reduce_by_segment` call is dependent on segments being processed serially by a single work-item. In particular, the `get<2>(lhs) += get<0>(rhs);` update is dependent on `get<2>(rhs)` having no run length information that needs to be propagated through the reduction. When the segment is reduced serially, this is always the case.

oneDPL's new reduce-by-segment performance improvements does not process segments serially but rather distributes work evenly throughout work items. Information regarding lengths of runs are lost in `get<2>(rhs)` when oneDPL's new sub-group scan is performed. To resolve this issue, the flag element in the tuple is changed from a bool to an integral type and is used to compute the length of the run instead of separating the flag from the run-length count. As a result, partial computations of the run-length in `get<0>(rhs)` are propagated through the reduction. Additional logic is required when defining the mask to ensure that all elements of the run are flagged, and the padded end case is properly handled.

Please note that this PR is dependent on https://github.com/oneapi-src/oneDPL/pull/1987 to compile.

### DIFF
--- a/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
+++ b/clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
@@ -2422,9 +2422,10 @@ void nontrivial_run_length_encode(ExecutionPolicy &&policy,
   auto zipped_vals_beg = make_zip_iterator(tr_nontrivial_flags, count_beg);
   auto pred = [](bool lhs, bool rhs) { return !rhs; };
   auto op = [](auto lhs, const auto &rhs) {
-    using ::std::get;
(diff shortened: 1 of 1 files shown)


zhiweij1 (approved):

tomflinda (approved):
LGMT

zhimingwang36 (approved):

PR #2600, opened 2025-01-07
[SYCLomatic] Add migration for cudaGraphGetNodes, cudaGraphGetRootNodes

daiyaan-ahmed6:
Signed-off-by: Daiyaan Ahmed <daiyaan.ahmed@intel.com>


clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -133,5 +133,41 @@ static void add_dependencies(dpct::experimental::command_graph_ptr graph,
...
+static void get_nodes(dpct::experimental::command_graph_ptr graph,
+                      dpct::experimental::node_ptr *nodesArray,
+                      std::size_t *numberOfNodes) {
+  auto nodes = graph->get_nodes();
+  *numberOfNodes = nodes.size();
+  if (!nodesArray) {
+    return;
+  }
+  for (std::size_t i = 0; i < *numberOfNodes; i++) {
+    nodesArray[i] = &nodes[i];
```
zhiweij1: I have 1 question here.
`nodes` is `std::vector`, if more nodes are pushed into that vector, the underlying memory may reallocated and elements will be copied to the new memory. So the address assigned to `nodesArray[i]` will be invalid at that time. Will that case happen?

zhiweij1 (approved):

ziranzha (approved):

zhimingwang36 (approved):

PR #2623, opened 2025-01-16
[SYCLomatic] Add support for `cuMemGetAllocationPropertiesFromHandle`

the-slow-one:
The API populates the `CUmemAllocationProp` based on the value of `CUmemGenericAllocationHandle`.

In the experimental support for virtual memory, the only property which is used for creating `sycl::ext::oneapi::experimental::physical_mem` is `location.id`. Also, the `CUmemAllocationProp` is used in API `cuMemCreate` and `cuMemGetAllocationGranularity` and in both the cases `location.id` seems to be the only useful field for the API migration (among other two i.e. `location.type` and `type`).

This PR migrates the API call to an assignment to set the `mem_prop.location.id` to the `device_id` used while creating the object of `physical_mem`.

Also fixes the warning message. The experimental feature option name in _virtual_mem_.  


clang/lib/DPCT/RulesLang/APINamesMemory.inc
```diff
@@ -717,7 +717,21 @@ CONDITIONAL_FACTORY_ENTRY(
...
+
+CONDITIONAL_FACTORY_ENTRY(
+    UseExpVirtualMemory,
+    ASSIGNABLE_FACTORY(FEATURE_REQUEST_FACTORY(
+        HelperFeatureEnum::device_ext,
+        ASSIGN_FACTORY_ENTRY(
+            "cuMemGetAllocationPropertiesFromHandle",
+            MEMBER_EXPR(DEREF(0), false, LITERAL("location.id")),
+            CALL(MapNames::getDpctNamespace() + "get_device_id",
+                MEMBER_CALL(ARG(1), true, "get_device"))))),
```
the-slow-one:
```suggestion
                 MEMBER_CALL(ARG(1), true, "get_device"))))),
```

ziranzha (approved):

clang/test/dpct/virtual_memory.cu
```diff
@@ -38,14 +38,17 @@ int main() {
     cuMemCreate(&allocHandle, POOL_SIZE, &prop, 0);
     cuMemMap(reserved_addr, POOL_SIZE, 0, allocHandle, 0);
 
+    // CHECK: prop.location.id = dpct::get_device_id(allocHandle->get_device());
+    cuMemGetAllocationPropertiesFromHandle(&prop, allocHandle);
```
intwanghao, line 42: I think prop.location.type and prop.type should also set value 1, even these two members not used in allocation handle creation. But the user may do some condition check on these two members in origin code, so need to set proper value for them.

the-slow-one: To prevent having the garbage value for these fields I think we can set them in the constructor.

clang/runtime/dpct-rt/include/dpct/memory.hpp
```diff
@@ -680,11 +680,16 @@ struct mem_location {
   int id;
   int type; // Location type. Value 1 means device location, and thus, id is a
             // device id.
+  public:
```
the-slow-one:
```suggestion
public:
```

clang/runtime/dpct-rt/include/dpct/memory.hpp
```diff
@@ -680,11 +680,16 @@ struct mem_location {
...
+  mem_location() : id{0}, type{1} {}
 };
 
 struct mem_prop {
   mem_location location;
   int type; // Memory type. Value 1 means default device memory.
+
+  public:
```
the-slow-one:
```suggestion
public:
```

clang/runtime/dpct-rt/include/dpct/memory.hpp
```diff
@@ -680,11 +680,16 @@ struct mem_location {
...
+public:
+  mem_location() : id{0}, type{1} {}
 };
 
 struct mem_prop {
   mem_location location;
   int type; // Memory type. Value 1 means default device memory.
+
+public:
+  mem_location() : location{}, type{1} {}
```
intwanghao: May be could write simply as:
struct mem_location {
  int id = 0;
  int type = 1;
};
struct mem_prop {
  mem_location location;
  int type = 1;
  }

clang/runtime/dpct-rt/include/dpct/memory.hpp
```diff
@@ -677,14 +677,14 @@ namespace experimental {
 typedef sycl::ext::oneapi::experimental::physical_mem *physical_mem_ptr;
 
 struct mem_location {
-  int id;
-  int type; // Location type. Value 1 means device location, and thus, id is a
+  int id = 0;
+  int type = 1; // Location type. Value 1 means device location, and thus, id is a
             // device id.
```
the-slow-one:
```suggestion
  int type = 1; // Location type. Value 1 means device location, and thus, id is a
                // device id.
```

intwanghao (approved):
LGTM

PR #2624, opened 2025-01-17
[SYCLomatic] Add support for `cuStreamGetCtx`

the-slow-one:
The API is migrated to a fetch and assign the device identifier (used to replace the `CUcontext`) using the given `sycl::queue` instance (representing `CUstream`).


clang/test/dpct/driver-stream-and-event.cu
```diff
@@ -183,3 +183,9 @@ unsigned getEventFlags(bool enabledSyncBlock) {
 
   return flags;
 }
+
+// CHECK: void test_stream_and_context(dpct::queur_ptr stream, int& context);
```
the-slow-one:
```suggestion
// CHECK: void test_stream_and_context(dpct::queue_ptr stream, int& context) {
```

clang/lib/DPCT/RulesLang/APINamesStream.inc
```diff
@@ -31,6 +31,12 @@ ASSIGNABLE_FACTORY(HEADER_INSERT_FACTORY(
                                    MEMBER_CALL(ARG(0), true, "wait"),
                                    CALL(ARG(1), ARG(0), ARG("0"), ARG(2)))))))
 
+ASSIGNABLE_FACTORY(FEATURE_REQUEST_FACTORY(
+    HelperFeatureEnum::device_ext,
+    ASSIGN_FACTORY_ENTRY("cuStreamGetCtx", DEREF(1),
+                         CALL(MapNames::getDpctNamespace() + "get_device_id",
+                            MEMBER_CALL(ARG(0), true, "get_device")))))
```
the-slow-one:
```suggestion
                         CALL(MapNames::getDpctNamespace() + "get_device_id",
                              MEMBER_CALL(ARG(0), true, "get_device")))))
```

ziranzha (approved):

clang/test/dpct/driver-stream-and-event.cu
```diff
@@ -183,3 +183,9 @@ unsigned getEventFlags(bool enabledSyncBlock) {
 
   return flags;
 }
+
+// CHECK: void test_stream_and_context(dpct::queue_ptr stream, int& context) {
+void test_stream_and_context(CUstream stream, CUcontext& context) {
+  // CHECK: context = dpct::get_device_id(stream->get_device());
+  cuStreamGetCtx(stream, &context);
```
ziranzha, line 190: Does it need to check the `stream` is a default stream?

tomflinda (approved):
LGTM

clang/test/dpct/driver-stream-and-event.cu
the-slow-one: I think we don't need to check for default stream because the `sycl::queue::get_device()` (pointed by `dpct::queue_ptr`) returns the associated `sycl::device`. We use the same device to identify the `device_id`. Please let me know, if I have missed something here.

ziranzha: Sounds good.

PR #2643, opened 2025-01-24
[SYCLomatic] Support the migration of cusparse<T>csrgemm2 related API

zhiweij1:


clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1400,191 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+
+  std::int64_t ws_size = 0;
+  oneapi::mkl::sparse::omatadd_buffer_size(
+      queue, oneapi::mkl::transpose::nontrans, oneapi::mkl::transpose::nontrans,
+      info.matrix_handle_c1.get_handle(), info.matrix_handle_d.get_handle(),
+      info.matrix_handle_c.get_handle(),
+      oneapi::mkl::sparse::omatadd_alg::default_alg,
+      info.omatadd_desc.get_handle(), ws_size);
+
+  void *ws = ::dpct::cs::malloc(ws_size, queue);
```
zhiweij1: it can be free-ed in this function.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1400,191 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+
+  __MATMAT(oneapi::mkl::sparse::matmat_request::compute, nullptr);
+
+  int nnz_c1_int = 0;
+#ifdef DPCT_USM_LEVEL_NONE
+  sycl::buffer<std::int64_t, 1> nnz_buf_c1(1);
+  __MATMAT(oneapi::mkl::sparse::matmat_request::get_nnz, &nnz_buf_c1);
+  nnz_c1_int = nnz_buf_c.get_host_access(sycl::read_only)[0];
+#else
+  std::int64_t *nnz_c1 = sycl::malloc_host<std::int64_t>(1, queue);
```
zhiweij1: it can be free-ed in this function

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1400,191 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+                  const std::shared_ptr<matrix_info> info_c, int *row_ptr_c,
+                  int *nnz_ptr) {
+  using Ty = typename ::dpct::detail::lib_data_traits_t<T>;
+  sycl::queue &queue = desc->get_queue();
+  detail::csrgemm2_args_info args(
+      m, n, k, info_a, val_a, row_ptr_a, col_ind_a, info_b, val_b, row_ptr_b,
+      col_ind_b, info_d, val_d, row_ptr_d, col_ind_d, info_c, row_ptr_c);
+  auto &info = desc->get_csrgemm2_info_map()[args];
+
+  int *row_ptr_c1 = (int *)::dpct::cs::malloc((m + 1) * sizeof(int), queue);
```
zhiweij1: it needs to be free-ed in the 2nd function

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1400,191 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  nnz_c1_int = nnz_buf_c.get_host_access(sycl::read_only)[0];
+#else
+  std::int64_t *nnz_c1 = sycl::malloc_host<std::int64_t>(1, queue);
+  __MATMAT(oneapi::mkl::sparse::matmat_request::get_nnz, nnz_c1);
+  queue.wait();
+  nnz_c1_int = *nnz_c1;
+#endif
+
+  int *col_ind_c1 = (int *)::dpct::cs::malloc(nnz_c1_int * sizeof(int), queue);
+  Ty *val_c1 = (Ty *)::dpct::cs::malloc(nnz_c1_int * sizeof(Ty), queue);
```
zhiweij1: they need to be free-ed in the 2nd function

clang/runtime/dpct-rt/include/dpct/detail/sparse_utils_detail.hpp
```diff
@@ -47,18 +82,26 @@ template <typename handle_t> class handle_manager {
...
     if (!_q || !_h)
       return;
-    sycl::event e = _rel_func(*_q, _h, _deps);
-    _q->submit([&](sycl::handler &cgh) {
-      cgh.depends_on(e);
-      cgh.host_task([_hh = _h] { delete _hh; });
-    });
+    release();
     _h = nullptr;
     _q = nullptr;
```
zhiweij1: remove

zhiweij1:
@gajanan-choudhary, could you help review this PR? Thanks.

clang/lib/DPCT/RulesMathLib/SpBLASAPIMigration.cpp
```diff
@@ -146,12 +151,20 @@ void SPBLASFunctionCallRule::runRule(const MatchFinder::MatchResult &Result) {
...
     return;
   }
-  if (FuncName == "cusparseXcsrgemmNnz") {
+  if (FuncName == "cusparseXcsrgemmNnz" || FuncName == "cusparseXcsrgemm2Nnz") {
+    auto consumerFuncNames = [&]() {
+      if (FuncName == "cusparseXcsrgemmNnz")
+        return hasAnyName("cusparseScsrgemm", "cusparseDcsrgemm",
+                          "cusparseCcsrgemm", "cusparseZcsrgemm");
+      return hasAnyName("cusparseScsrgemm2", "cusparseDcsrgemm2",
+                        "cusparseCcsrgemm2", "cusparseZcsrgemm2");
```
gajanan-choudhary:
```suggestion
      if (FuncName == "cusparseXcsrgemmNnz")
        return hasAnyName("cusparseScsrgemm", "cusparseDcsrgemm",
                          "cusparseCcsrgemm", "cusparseZcsrgemm");
      else /* if (FuncName == "cusparseXcsrgemm2Nnz") */
          return hasAnyName("cusparseScsrgemm2", "cusparseDcsrgemm2",
                            "cusparseCcsrgemm2", "cusparseZcsrgemm2");
```

clang/runtime/dpct-rt/include/dpct/detail/sparse_utils_detail.hpp
```diff
@@ -79,16 +120,33 @@ template <typename handle_t> class handle_manager {
...
+    handle_manager<oneapi::mkl::sparse::omatadd_descr_t>::_init_func =
+        oneapi::mkl::sparse::init_omatadd_descr;
+template <>
+inline handle_manager<oneapi::mkl::sparse::omatadd_descr_t>::rel_func_t
+    handle_manager<oneapi::mkl::sparse::omatadd_descr_t>::_rel_func =
+        [](sycl::queue &queue, oneapi::mkl::sparse::omatadd_descr_t *p_desc,
+           const std::vector<sycl::event> &dependencies) -> sycl::event {
+  return oneapi::mkl::sparse::release_omatadd_descr(queue, *p_desc,
+                                                    dependencies);
+};
```
gajanan-choudhary, line 108: Going forward, release functions for our API descriptors (e.g. `sparse::release_omatadd_descr` for `omatadd_descr_t`) will be taking in the object as the argument rather than pointer-to-object as argument (e.g., `omatadd_descr_t` object as argument instead of `omatadd_descr_t *`). Unfortunately we are in a state where `release_matrix_handle` and `release_matmat_descr` take pointer-to-object as arguments for freeing which is an oversight and not required.

I don't think this piece of information _requires_ you to change your code here, but maybe it could simplify things for you for `omatadd_descr_t`, I don't know.

clang/runtime/dpct-rt/include/dpct/detail/sparse_utils_detail.hpp
```diff
@@ -79,16 +120,33 @@ template <typename handle_t> class handle_manager {
...
+
 template <>
 inline handle_manager<oneapi::mkl::sparse::matmat_descr_t>::init_func_t
     handle_manager<oneapi::mkl::sparse::matmat_descr_t>::_init_func =
-        oneapi::mkl::sparse::init_matmat_descr;
+        [](sycl::queue &queue, oneapi::mkl::sparse::matmat_descr_t *p_desc) {
+          oneapi::mkl::sparse::init_matmat_descr(p_desc);
+        };
```
gajanan-choudhary, line 115: It's good you have kept the option to use a queue here, even though `init_matmat_descr` doesn't use it as an arg (which was an oversight on our end, but oneMKL sparse BLAS's future new SYCL APIs will have `queue` in their arguments for the foreseeable future).

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -559,6 +559,33 @@ class descriptor {
...
+    void *val_c1 = nullptr;
+    void *ws = nullptr;
+    bool is_empty() { return omatadd_desc.is_empty(); }
+    void init(sycl::queue *q_ptr) {
+      omatadd_desc.init(q_ptr);
+      matrix_handle_c1.init(q_ptr);
+      matrix_handle_d.init(q_ptr);
+      matrix_handle_c.init(q_ptr);
+    }
+  };
```
gajanan-choudhary: Very interesting! Am I reading this correctly that you are performing `C = alpha * A * B + beta * D` with the help of `sparse::matmat` and `sparse::omatadd` and calling it `matmat2`, to match with the `cusparseXcsrgemm2` APIs (that were removed in CUDA 12.0.0) that used to perform that operation?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -559,6 +559,33 @@ class descriptor {
                      detail::csrgemm_args_info_hash>
       _csrgemm_info_map;
 
+  struct matmat2_info {
+    detail::matrix_handle_manager matrix_handle_c1;
+    detail::matrix_handle_manager matrix_handle_d;
+    detail::matrix_handle_manager matrix_handle_c;
+    detail::handle_manager<oneapi::mkl::sparse::omatadd_descr_t> omatadd_desc;
```
gajanan-choudhary: Why don't you need to store `matmat_descr_t` object here too?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -559,6 +559,33 @@ class descriptor {
...
+    void *col_ind_c1 = nullptr;
+    void *val_c1 = nullptr;
+    void *ws = nullptr;
+    bool is_empty() { return omatadd_desc.is_empty(); }
+    void init(sycl::queue *q_ptr) {
+      omatadd_desc.init(q_ptr);
+      matrix_handle_c1.init(q_ptr);
+      matrix_handle_d.init(q_ptr);
+      matrix_handle_c.init(q_ptr);
+    }
```
gajanan-choudhary: Don't you need to add a `release` function too here in this `struct`?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  oneapi::mkl::sparse::matmat_descr_t matmat_desc = nullptr;
+  oneapi::mkl::sparse::init_matmat_descr(&matmat_desc);
+  oneapi::mkl::sparse::set_matmat_data(
+      matmat_desc, oneapi::mkl::sparse::matrix_view_descr::general,
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::sparse::matrix_view_descr::general,
+      oneapi::mkl::transpose::nontrans,
+      oneapi::mkl::sparse::matrix_view_descr::general);
+
+#ifdef DPCT_USM_LEVEL_NONE
```
gajanan-choudhary, line 1519: What does this `DPCT_USM_LEVEL_NONE` do? Does this mean that if that macro is defined, then you use `sycl::buffer` code-path/APIs versus if it is not defined, you use USM APIs? (At least that's what it is looking like to me based on the `matmat` calls in the `#ifdef` and `#else` blocks.)

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  sycl::buffer<std::int64_t, 1> nnz_buf_c1(1);
+  __MATMAT(oneapi::mkl::sparse::matmat_request::get_nnz, &nnz_buf_c1);
+  nnz_c1_int = nnz_buf_c1.get_host_access(sycl::read_only)[0];
+#else
+  std::int64_t *nnz_c1 = sycl::malloc_host<std::int64_t>(1, queue);
+  __MATMAT(oneapi::mkl::sparse::matmat_request::get_nnz, nnz_c1);
+  queue.wait();
+  nnz_c1_int = *nnz_c1;
+  sycl::free(nnz_c1, queue);
+#endif
```
gajanan-choudhary: Having a `queue.wait()` for USM code-path is overkill. The USM API returns a `sycl::event`. It would be preferable to wait on only the returned event rather than blocking everything with a `queue.wait()`. You may want to modify your __MATMAT macro so that you can avoid unnecessary `queue.wait()` calls and instead wait on individual events only when needed. My suggestion (may need some tweaks below!):

```suggestion
#ifdef DPCT_USM_LEVEL_NONE
#define __MATMAT(_IGNORED_EVENT, STEP, NNZ_C1, _IGNORED_DEPS) \
  oneapi::mkl::sparse::matmat(queue, a, b, info.matrix_handle_c1.get_handle(), \
                              STEP, matmat_desc, NNZ_C1, nullptr)
#else
#define __MATMAT(EVENT, STEP, NNZ_C1, DEPS) \
  EVENT = oneapi::mkl::sparse::matmat(queue, a, b, info.matrix_handle_c1.get_handle(), \
                              STEP, matmat_desc, NNZ_C1, nullptr, DEPS)
#endif

  sycl::event ev_work_est, ev_compute, ev_get_nnz;
#ifdef DPCT_USM_LEVEL_NONE
  sycl::buffer<std::int64_t, 1> *nnz_buf_c1 = new sycl::buffer<std::int64_t, 1>(1);
#else
  std::int64_t *nnz_c1 = sycl::malloc_host<std::int64_t>(1, queue);
#endif

  __MATMAT(ev_work_est, oneapi::mkl::sparse::matmat_request::work_estimation, nullptr, {});
  __MATMAT(ev_compute, oneapi::mkl::sparse::matmat_request::compute, nullptr, {ev_work_est});
  __MATMAT(ev_get_nnz, oneapi::mkl::sparse::matmat_request::get_nnz, nnz_buf_c1, {ev_compute});

#ifdef DPCT_USM_LEVEL_NONE
  int nnz_c1_int = nnz_buf_c1->get_host_access(sycl::read_only)[0];
  delete nnz_buf_c1;
#else
  ev_get_nnz.wait();
  int nnz_c1_int = *nnz_c1;
  sycl::free(nnz_c1, queue);
#endif
```

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+
+  std::int64_t ws_size = 0;
+  oneapi::mkl::sparse::omatadd_buffer_size(
+      queue, oneapi::mkl::transpose::nontrans, oneapi::mkl::transpose::nontrans,
+      info.matrix_handle_c1.get_handle(), info.matrix_handle_d.get_handle(),
+      info.matrix_handle_c.get_handle(),
+      oneapi::mkl::sparse::omatadd_alg::default_alg,
+      info.omatadd_desc.get_handle(), ws_size);
+
+  void *ws = ::dpct::cs::malloc(ws_size, queue);
```
gajanan-choudhary: Is this a combined allocation function that works for both `sycl::buffer`s as well as USM? I see above you have different things for `sycl::buffer` and USM in case of `nnz_c1`, but you only have 1 combined call here.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+    ::dpct::cs::memcpy(::dpct::cs::get_default_queue(), nnz_ptr, &c_nnz_int,
+                       sizeof(int))
+        .wait();
+  }
+  if (info_c->get_index_base() == oneapi::mkl::index_base::one) {
+    c_nnz_int++;
+  }
+  ::dpct::cs::memcpy(::dpct::cs::get_default_queue(), row_ptr_c + m, &c_nnz_int,
+                     sizeof(int))
+      .wait();
```
gajanan-choudhary: I think there is one thing missing here, unless you are doing it somewhere else. [cuSPARSE 11.8.0 documentation for cusparseXcsrgemm2](https://docs.nvidia.com/cuda/archive/11.8.0/cusparse/index.html) says:
> In the second step, the user gathers nnzC (the number of nonzero elements of matrix C) from either (nnzC=*nnzTotalDevHostPtr) or (nnzC=csrRowPtrC(m)-csrRowPtrC(0)) 

This means that not only do you need to set `row_ptr_c[m] = c_nnz_int + indexBase`, but you **also** need to set `row_ptr_c[0] = indexBase`. If you haven't already done that somewhere else, please add that here.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  queue.submit([&](sycl::handler &cgh) {
+    cgh.depends_on(events);
+    cgh.host_task([_p1 = info.row_ptr_c1, _p2 = info.col_ind_c1,
+                   _p3 = info.val_c1, _p4 = info.ws, _q = queue] {
+      ::dpct::cs::free(_p1, _q);
+      ::dpct::cs::free(_p2, _q);
+      ::dpct::cs::free(_p3, _q);
+      ::dpct::cs::free(_p4, _q);
+    });
+  });
```
gajanan-choudhary: Maybe keep this as a separate `release` function in `struct matmat2_info` in `sparse_utils.hpp`?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+}
+
```
gajanan-choudhary, line 1776: Overall, this is fantastic! I think adding an example for performing `C = alpha * op(A) * op(B) + beta * op(D)` as part of oneMKL library will also benefit us!

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
gajanan-choudhary: Also, is this the right place to free these items? I believe these only need to be freed when `matmat2_info` object is being destroyed, isn't it?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1402,274 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+/// \param [in] col_ind_d An array containing the column indices in index-based
+/// numbering.
+/// \param [in] info_c Matrix info of the matrix C.
+/// \param [out] val_c An array containing the non-zero elements of the matrix
+/// C.
+/// \param [in] row_ptr_c An array of length row number + 1.
+/// \param [out] col_ind_c An array containing the column indices in index-based
+/// numbering.
+template <typename T>
+void csrgemm2(descriptor_ptr desc, int m, int n, int k, const T *alpha,
```
gajanan-choudhary, line 1708: Don't you also need an implementation/translation for the cuSPARSE's `cusparseXcsrgemm2_bufferSizeExt` API?

clang/test/dpct/cusparse.cu
```diff
@@ -471,3 +471,177 @@ void foo7_1(int c_nnz,
...
+            int nnzD,
+            int nnzC,
+            csrgemm2Info_t info_s,
+            size_t ws_s) {
+#ifndef NO_BUILD_TEST
+  // CHECK: /*
+  // CHECK-NEXT: DPCT1134:{{[0-9]+}}: The tool cannot deduce the consumer API ("dpct::sparse::csrgemm") of this API, and this API has 2 arguments depending on the 8th and the 12th parameters of the consumer API. Please replace the 2 arguments tagged as "dpct_placeholder" with the corresponding value.
+  // CHECK-NEXT: */
+  // CHECK-NEXT: dpct::sparse::csrgemm2_nnz(handle, 3, 4, 3, descrA, nnzA, dpct_placeholder, row_ptr_a, col_ind_a, descrB, nnzB, dpct_placeholder, row_ptr_b, col_ind_b, descrD, nnzD, dpct_placeholder, row_ptr_d, col_ind_d, descrC, row_ptr_c, &nnzC);
+  cusparseXcsrgemm2Nnz(handle, 3, 4, 3, descrA, nnzA, row_ptr_a, col_ind_a, descrB, nnzB, row_ptr_b, col_ind_b, descrD, nnzD, row_ptr_d, col_ind_d, descrC, row_ptr_c, &nnzC, info_s, &ws_s);
```
gajanan-choudhary: Why is there no call to `cusparseXcsrgemm2_bufferSizeExt` API?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
gajanan-choudhary: Although your current implementation should work in simple scenarios (e.g. single calls to perform operations, no repeated calls/requests), it may run into issues in more complicated cases. I believe the following division is ideally needed, but may be complicated to implement:

1. `cusparseXcsrgemm2_bufferSizeExt` translation layer:
    * Create `A`, `B`, `C1` matrix handles
    * Create `matmat_descr_t` object,
    * Perform `matmat`'s `work estimation`, `compute_structure` (**not** `compute`!), and `finalize_structure` into `C1` matrix.
    * Create `omatadd_descr_t` object and `C` matrix handle
    * Call `omatadd_get_buffer_size()` API to get temporary workspace size. **This** size is returned to the user to allocate and is maintained on the user-side.
    * Do not free `matmat_descr_t` object. It must be stored and reused in future `csrgemm2` calls
    * Do not free the A/B matrix handles (or if you do, recreate them later in future API calls)
1. `cusparseXcsrgemm2Nnz` translation layer:
    * Call `omatadd_analyze()` and `omatadd_get_nnz()` functions
    * Copy over `nnz` value appropriately
1. `cusparseXcsrgemm2` translation layer:
    * Call `matmat`'s `compute` and `finalize` stages
    * Then call `omatadd()`
    * Which objects are freed and which ones are not freed here needs careful consideration depending on who owns the objects. Maybe some must be freed here and others must be stored in the `info` object to be freed later only when the user destroys the `info` object.

It's a difficult task, and you need not do it right away (or may never need to do it), but please make a note of this somewhere in the code itself as comments so that you don't forget this in the long run.

gajanan-choudhary (commented):
Although I have many questions and have requested some changes, I believe this is fantastic work given how complicated the operation is and the APIs are!

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: Yes

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: Because in current impl, matmul operations are all finished in the `matmul2_nnz()`, so it is unnecessary to save it and pass into the next `matmul2()` (for omatadd) function.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: The default dtor of this struct will call the dtor of `detail::matrix_handle_manager`. In that dtor, MKL release function is exectued.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: You are right. By default, SYCLomatic generates USM-based code. But we have an option to generate buffer-based code. And since this file is a header (not generated by SYCLomatic), we use that macro to split different code path.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: Yes. In that function there are 2 code paths for USM and buffer.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: > Maybe keep this as a separate release function in struct matmat2_info in sparse_utils.hpp?

I think keeping it here is okay because the malloc is not in the `matmat2_info ` ctor, either.

> Also, is this the right place to free these items? I believe these only need to be freed when matmat2_info object is being destroyed, isn't it?

Yes. At this time `matmat2_info` is already destroyed. It is destroyed in this line:
```
desc->get_csrgemm2_info_map().erase(args);
```
BTW, I find I need to use 4 stack variables to save the 4 pointers since the data behind `info` is already destroyed (`info` is a reference).

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: Actually, event.wait() and queue.wait() are the same here because the queue created by SYCLomatic is in-ordered (in the USM case). The in-ordered queue can have more similar behavior as CUDA's cudaStream_t.

clang/test/dpct/cusparse.cu
zhiweij1: Test updated

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: Hi @gajanan-choudhary, thanks for the suggestions.
I have updated the PR. By using the `compute_structure` and `finalize_structure`, I can eliminate inserting the value pointers in the `csrgemm2_nnz`. The code logic is simplified.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1192,8 +1192,7 @@ inline void csr2csc(sycl::queue queue, int m, int n, int nnz,
 /// Calculate the non-zero elements number of the result of a
 /// sparse matrix (CSR format)-sparse matrix (CSR format) product:
 /// C = op(A) * op(B)
-/// \param [in] queue The queue where the routine should be executed. It must
-/// have the in_order property when using the USM mode.
+/// \param [in] desc The descriptor of this calculation.
```
zhimingwang36, line 1195: is it wrong update?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1352,426 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
   info.matmat_desc.add_dependency(e);
   desc->get_csrgemm_info_map().erase(args);
 }
+
+class csrgemm2_info {
```
zhimingwang36, line 1358: please add comments for the public interface class.

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1352,426 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  detail::matrix_handle_manager matrix_handle_a;
+  detail::matrix_handle_manager matrix_handle_b;
+  detail::matrix_handle_manager matrix_handle_c1;
```
zhimingwang36:
```suggestion
  detail::matrix_handle_manager matrix_handle_axb;
```

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1352,426 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+/// (CSR format) product:
+/// C = alpha * A * B + beta * D
+/// \param [in] desc The descriptor of this calculation.
+/// \param [in] m The rows number of A, D and C.
+/// \param [in] n The columns number of B, D and C.
+/// \param [in] k The columns number of A and rows number of B.
+/// \param [in] alpha Scaling factor.
+/// \param [in] info_a Matrix info of the matrix A.
+/// \param [in] nnz_a Non-zero elements number of matrix A.
+/// \param [in] row_ptr_a An array of length row number + 1.
```
zhimingwang36:
```suggestion
/// \param [in] row_ptr_a An array of length \p m + 1.
```

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1352,426 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+  oneapi::mkl::sparse::omatadd_buffer_size(
+      queue, oneapi::mkl::transpose::nontrans, oneapi::mkl::transpose::nontrans,
+      info->matrix_handle_c1.get_handle(), info->matrix_handle_d.get_handle(),
+      info->matrix_handle_c.get_handle(),
+      oneapi::mkl::sparse::omatadd_alg::default_alg,
+      info->omatadd_desc.get_handle(), ws_size);
+  *buffer_size_in_bytes = ws_size;
+}
+
+/// Calculate the non-zero elements number of the result of a
```
zhimingwang36: +D?

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
```diff
@@ -1354,6 +1352,426 @@ void csrgemm(descriptor_ptr desc, oneapi::mkl::transpose trans_a,
...
+#endif
+  );
+  std::int64_t nnz_c = 0;
+  oneapi::mkl::sparse::omatadd_get_nnz(
+      queue, oneapi::mkl::transpose::nontrans, oneapi::mkl::transpose::nontrans,
+      info->matrix_handle_c1.get_handle(), info->matrix_handle_d.get_handle(),
+      info->matrix_handle_c.get_handle(),
+      oneapi::mkl::sparse::omatadd_alg::default_alg,
+      info->omatadd_desc.get_handle(), nnz_c);
+  int nnz_c_int = nnz_c;
```
zhimingwang36, line 1663: Please add warning in case: this api is used in none migration code scenario?

clang/lib/DPCT/SrcAPI/APINames_cuSPARSE.inc
```diff
@@ -70,8 +70,8 @@ ENTRY(cusparseCreateBsric02Info, cusparseCreateBsric02Info, false, NO_FLAG, P4,
 ENTRY(cusparseDestroyBsric02Info, cusparseDestroyBsric02Info, false, NO_FLAG, P4, "comment")
 ENTRY(cusparseCreateBsrilu02Info, cusparseCreateBsrilu02Info, false, NO_FLAG, P4, "comment")
 ENTRY(cusparseDestroyBsrilu02Info, cusparseDestroyBsrilu02Info, false, NO_FLAG, P4, "comment")
-ENTRY(cusparseCreateCsrgemm2Info, cusparseCreateCsrgemm2Info, false, NO_FLAG, P4, "comment")
-ENTRY(cusparseDestroyCsrgemm2Info, cusparseDestroyCsrgemm2Info, false, NO_FLAG, P4, "comment")
+ENTRY(cusparseCreateCsrgemm2Info, cusparseCreateCsrgemm2Info, true, NO_FLAG, P4, "DPCT1026/DPCT1027")
```
zhiweij1:
```suggestion
ENTRY(cusparseCreateCsrgemm2Info, cusparseCreateCsrgemm2Info, true, NO_FLAG, P4, "comment")
```

clang/runtime/dpct-rt/include/dpct/sparse_utils.hpp
zhiweij1: no

zhimingwang36 (approved):
LGTM

PR #2655, opened 2025-02-07
[SYCLomatic] Add cast, when required, while migrating `cuMemFree`

the-slow-one:
In USM restricted mode, `cuMemFree` is migrated to `sycl::free`. In the case when the argument to `cuMemFree` isn't of type `CUdeviceptr` (says it's u64) the  `sycl::free` need appropriate casting to `char *` (in other words `dpct::device_ptr`).


ziranzha (approved):

clang/lib/DPCT/RulesLang/RulesLang.cpp
```diff
@@ -6260,6 +6260,32 @@ void MemoryMigrationRule::freeMigration(const MatchFinder::MatchResult &Result,
   if (isPlaceholderIdxDuplicated(C))
     return;
 
+  if (Name == "cuMemFree_v2") {
```
zhiweij1: Can you implement this logic with rewriter? The migration of `cuMemFree_v2` uses rewriter already.

the-slow-one: Done.

clang/lib/DPCT/RuleInfra/CallExprRewriterCommon.h
```diff
@@ -155,6 +155,37 @@ template <class SubExprT> class CastIfNotSameExprPrinter {
...
+  template <typename StreamT> void print(StreamT &Stream) const {
+    if (!SubExpr)
+      return;
+
+    const Expr *CleanSubExpr = SubExpr->IgnoreUnlessSpelledInSource();
+
+    if (const Type *SubExprType = CleanSubExpr->getType().getTypePtr()) {
+      if (const NamedDecl *TypeName = getNamedDecl(SubExprType)) {
+        if (TypeName->getNameAsString() != GivenTypeName) {
+          Stream << "reinterpret_cast<dpct::device_ptr>(";
```
zhiweij1: Use `ReplacementTypeName` instead of `dpct::device_ptr`?

zhiweij1 (approved):

PR #2658, opened 2025-02-11
[SYCLomatic] Fix the issue that global device memory is invisible in other translation unit after migration

intwanghao:


ziranzha (approved):

zhimingwang36 (approved):
LGTM

PR #2662, opened 2025-02-13
[SYCLomatic] Remove -fgpu-exclude-wrong-side-overloads to avoid other parsing errors

zhiweij1:
This option was added to fix one parsing error.
But now we found it introduced other errors.
So in the PR, we remove it and choose another way to fix the previous issue.


tomflinda (approved):
LGTM

ziranzha (approved):

zhimingwang36 (approved):

PR #2667, opened 2025-02-17
[SYCLomatic] Enable the migration of API cudaLaunchHostFunc

intwanghao:


ziranzha (approved):

clang/runtime/dpct-rt/include/dpct/kernel.hpp
```diff
@@ -35,6 +35,8 @@ namespace dpct {
 typedef void (*kernel_functor)(sycl::queue &, const sycl::nd_range<3> &,
                                unsigned int, void **, void **);
 
+typedef void (*host_callback)(void *);
```
zhimingwang36: it's make a more common name, which can be used in other context but callback, like host_func

zhimingwang36 (approved):

PR #2670, opened 2025-02-18
[Bug][Python Migration] Added support for parsing Python syntax inside blocks like (), [] & {}

TejaX-Alaghari:
This PR adds support for parsing Python syntax inside blocks like 

- tuples - ()
- lists - []
- dicts - {}

### DIFF
--- a/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
+++ b/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
@@ -9,10 +9,10 @@
 #include "UserDefinedRules/PatternRewriter.h"
 #include "AnalysisInfo.h"
 #include "Diagnostics/Diagnostics.h"
+#include "FileGenerator/GenFiles.h"
 #include "MigrateScript/MigrateCmakeScript.h"
 #include "MigrateScript/MigratePythonBuildScript.h"
 #include "UserDefinedRules/UserDefinedRules.h"
-#include "FileGenerator/GenFiles.h"
 
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Path.h"
@@ -24,7 +24,6 @@
 #include <unordered_map>
 #include <variant>
 #include <vector>
-
 
 namespace clang {
 namespace dpct {
@@ -62,8 +61,26 @@ static bool isWhitespace(char Character) {
 
 static bool isNotWhitespace(char Character) { return !isWhitespace(Character); }
 
+static bool isLeftDelimiter(char Character) {
+  return Character == '{' || Character == '[' || Character == '(';
+}
+
 static bool isRightDelimiter(char Character) {
   return Character == '}' || Character == ']' || Character == ')';
+}
+
+static char getRightDelimiter(char Character) {
+  char rightChar = '\0';
+
+  if (Character == '{') {
+    rightChar = '}';
+  } else if (Character == '[') {
+    rightChar = ']';
+  } else if (Character == '(') {
+    rightChar = ')';
+  }
+
+  return rightChar;
 }
 
 static int detectIndentation(const std::string &Input, int Start) {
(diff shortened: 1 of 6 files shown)


clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
```diff
@@ -66,6 +65,20 @@ static bool isRightDelimiter(char Character) {
   return Character == '}' || Character == ']' || Character == ')';
 }
 
+static bool getRightDelimiter(char Character) {
```
zhiweij1:
```suggestion
static char getRightDelimiter(char Character) {
```

TejaX-Alaghari: That char to bool change for return type was a mistaken replacement when I committed

Changed it back properly now

zhiweij1 (changes requested):
From an overall view, I think this PR is a little strange.
There are too many special processes in the parser.
For example: can we just pass `MatchPattern` into `parseBlock` and use it in the `parseCodeElement` call?
@tomflinda

clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
```diff
@@ -295,24 +324,36 @@ static int parseCodeElement(const MatchPattern &Suffix,
...
+      char rightDelimiter = getRightDelimiter(Character);
 
-    if (Character == '[') {
-      Index = parseBlock('[', ']', Input, Index, Mode);
-      continue;
-    }
-
-    if (Character == '(') {
-      Index = parseBlock('(', ')', Input, Index, Mode);
-      continue;
```
zhiweij1: And I think to parse the content inside, `Index` should be `++Index`

tomflinda:
> From an overall view, I think this PR is a little strange. There are too many special processes in the parser. For example: can we just pass `MatchPattern` into `parseBlock` and use it in the `parseCodeElement` call? @tomflinda

@zhiweij1 yes, I think so.

clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
TejaX-Alaghari: Index is getting incremented inside parseBlock [here](https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp#L253)

Did you mean the same or more additional increment?

zhiweij1: > Index is getting incremented inside parseBlock [here](https://github.com/oneapi-src/SYCLomatic/blob/SYCLomatic/clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp#L253)

You are right.

clang/lib/DPCT/UserDefinedRules/PatternRewriter.cpp
```diff
@@ -418,6 +459,12 @@ static void applyExtenstionNameChange(
   for (; Pos > 0 && !isWhitespace(Input[Pos]); Pos--) {
   }
   Pos = Pos == 0 ? 0 : Pos + 1;
+  auto Size = Input.size();
+  // Skip the dangling left delimiters
+  for (; Pos < Size && isLeftDelimiter(Input[Pos]); Pos++) {
+  }
+  if (Input[Pos] == '[')
+    Pos++;
```
tomflinda: Take the "out = func([["foo.cpp", "bar.cu", "baz.cpp"]])" for example.

Instead of skipping '{', we should locate the root cause of why the target "foo.cpp" could not be matched.

TejaX-Alaghari: Updated the logic now to find the file beginnings correctly

TejaX-Alaghari:
> From an overall view, I think this PR is a little strange. There are too many special processes in the parser. For example: can we just pass `MatchPattern` into `parseBlock` and use it in the `parseCodeElement` call? @tomflinda

Re-implemented the logic as suggested now.

tomflinda (approved):
LGTM

zhiweij1 (approved):

PR #2712, opened 2025-03-14
[SYCLomatic] Use .x()/.y()/... instead of [...] to migrate bfloat16 member access

zhiweij1:


clang/lib/DPCT/RulesLang/Math/RewriterHalfPrecisionConversionAndDataMovement.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include "CallExprRewriterMath.h"
 #include "CommonMacroDefinition.h"
+#include "RuleInfra/CallExprRewriterCommon.h"
```
zhiweij1: remove

ziranzha (approved):

tomflinda (approved):
LGTM

keryell:
What is the motivation for this?
`vec` are somewhat deprecated and `marray` should be used instead.

zhiweij1:
> What is the motivation for this? `vec` are somewhat deprecated and `marray` should be used instead.

Hi @keryell , this is because all cuda vec types are currently mapped to sycl::vec. Before this PR, only bfloat16 vec used marray since the math functions of bfloat16 vec only supported marray interface at that time.
But now the math functions of bfloat16 vec also supported vec interface so we can unify the migration to avoid issue when meeting code like:
```
template<class T1, class T2>
__device__ inline void scale4(T1* frag_s_1) {
  T2 s_val_1_2;
  s_val_1_2.x = frag_s_1[0];
  s_val_1_2.y = frag_s_1[1];
}
 
__global__ void k() {
  half *a1;
  __nv_bfloat16 *a2;
 
  scale4<__nv_bfloat16, __nv_bfloat162>(a2);
  scale4<half, half2>(a1);
}
```
marray only has [] operator but no .x()/.y() member functions.

Totally changing the migration of all cuda vec to sycl marray is another method. But it is a big task. Currently, we decide to keep the old style.

keryell:
Thanks for the context!

PR #2717, opened 2025-03-18
[LIT][PTX][NFC] Lit case for prefetch ASM instruction

TejaX-Alaghari:
Added the lit case for prefetch.L1 with %1 instead of %0


tomflinda (approved):
LGTM

PR #2732, opened 2025-03-23
[Bug][LIT][NFC] Fix test mechanism for lit tests

TejaX-Alaghari:
This PR aims to fix the lit test mechanism used to check the similarity between input and expected files

In the current lit test cases of SYCLomatic, we -
1. diff the contents of input and expected files and
2. check the begin and end statements in diff.txt

But the 2nd step is essential useless without the `FileCheck` command that specifies the input file to check the begin and end statements in.

We can fix this in 2 ways -
1. Add `FileCheck %s --input-file %T/diff.txt` before `// CHECK: begin`
2. Remove `// CHECK` lines altogether and rely on `diff` command to find the differences between input and expected file contents

The first commit in this PR uses 1st way
The second commit uses 2nd way upon suggestion from @tomflinda to simplify the test case

### DIFF
--- a/clang/test/dpct/cmake_migration/case_001/case_001_framework.cpp
+++ b/clang/test/dpct/cmake_migration/case_001/case_001_framework.cpp
@@ -11,31 +11,31 @@
 // RUN: diff --strip-trailing-cr %S/expected.txt %T/out_1/input.cmake >> %T/diff.txt
 // RUN: diff --strip-trailing-cr %S/sub_expected.txt %T/out_1/subdir/sub_input.cmake >> %T/diff.txt
 // RUN: echo "end" >> %T/diff.txt
-// CHECK: begin
-// CHECK-NEXT: end
+// RUN: FileCheck %s --match-full-lines --input-file %T/diff.txt
 
 // RUN: dpct -in-root ./ -out-root out_2 ./input.cmake ./subdir/sub_input.cmake --migrate-build-script-only --migrate-build-script=CMake
 // RUN: echo "begin" > %T/diff.txt
 // RUN: diff --strip-trailing-cr %S/expected.txt %T/out_2/input.cmake >> %T/diff.txt
 // RUN: diff --strip-trailing-cr %S/sub_expected.txt %T/out_2/subdir/sub_input.cmake >> %T/diff.txt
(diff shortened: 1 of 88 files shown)


clang/test/dpct/cmake_migration/case_001/case_001_framework.cpp
```diff
@@ -11,31 +11,31 @@
 // RUN: diff --strip-trailing-cr %S/expected.txt %T/out_1/input.cmake >> %T/diff.txt
 // RUN: diff --strip-trailing-cr %S/sub_expected.txt %T/out_1/subdir/sub_input.cmake >> %T/diff.txt
 // RUN: echo "end" >> %T/diff.txt
-// CHECK: begin
-// CHECK-NEXT: end
+// RUN: FileCheck %s --match-full-lines --input-file %T/diff.txt
```
tomflinda: I'm not sure why the new line should be added, as the original check already covers its functionality.

clang/test/dpct/cmake_migration/case_003/case_003_add_executable.cpp
```diff
@@ -5,6 +5,7 @@
 // RUN: echo "begin" > %T/diff.txt
 // RUN: diff --strip-trailing-cr %S/expected.txt %T/out/input.cmake >> %T/diff.txt
 // RUN: echo "end" >> %T/diff.txt
+// RUN: FileCheck %s --match-full-lines --input-file %T/diff.txt
```
tomflinda: Same as this and the other rests.

clang/test/dpct/cmake_migration/case_001/case_001_framework.cpp
TejaX-Alaghari: The original check is effectively redundant without FileCheck command that parses the file contents of diff.txt
So, I've added them

clang/test/dpct/cmake_migration/case_003/case_003_add_executable.cpp
TejaX-Alaghari: Same response as above

tomflinda (approved):
LGTM

zhimingwang36 (approved):

PR #2746, opened 2025-03-27
[PTX] Enable migration of mma PTX ASM instruction for 11 shapes

TejaX-Alaghari:
This PR adds support for the migration of mma PTX ASM instruction for 11 shapes

- m8n8k4
- m8n8k16
- m8n8k32
- m8n8k128
- m16n8k4
- m16n8k8
- m16n8k16
- m16n8k32
- m16n8k64
- m16n8k128
- m16n8k256


clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -2055,6 +2055,79 @@ class joint_matrix {
...
+/// \param [in] b0 The 1st element from B matrix to be multiplied with A matrix
+/// \param [in] b1 The 2nd element from B matrix to be multiplied with A matrix
+/// \param [in] c0 The 1st element from C matrix to be added with d0
+/// \param [in] c1 The 2nd element from C matrix to be added with d1
+/// \param [in] c2 The 3rd element from C matrix to be added with d2
+/// \param [in] c3 The 4th element from C matrix to be added with d3
+/// \param [in] item The sycl::nd_item index space class
+template <typename MulType, typename ABType, typename CDType, typename ItemT>
+__attribute__((optnone)) void
+mma(CDType *d0, CDType *d1, CDType *d2, CDType *d3, ABType a0, ABType a1,
```
tomflinda: According the PTX spec: https://docs.nvidia.com/cuda/parallel-thread-execution/#matrix-multiply-accumulate-operation-using-mma-instruction, there are 11 kinds of shapes for mma:
```
m8n8k4
m8n8k16
m8n8k32
m8n8k128
m16n8k4
m16n8k8
m16n8k16
m16n8k32 
m16n8k64
m16n8k128
m16n8k256
```
Pls make sure your helper function has the capability to support them besides the shape of "m16n8k16".

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -2055,6 +2055,79 @@ class joint_matrix {
   matrix_accessor x;
   const size_t num_elements;
 };
+
+/// Multiplies 2 16x16 & 16x8 matrices and accumulates the result to a 16x8 b32
+/// matrix
```
tomflinda: For this helper function to work correctly, the size of the subgroup for the kernel function must be 32, the description for this helper function should clarify this.

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -2055,6 +2055,79 @@ class joint_matrix {
...
+/// \param [in] a3 The 4th element from A matrix to be multiplied with B matrix
+/// \param [in] b0 The 1st element from B matrix to be multiplied with A matrix
+/// \param [in] b1 The 2nd element from B matrix to be multiplied with A matrix
+/// \param [in] c0 The 1st element from C matrix to be added with d0
+/// \param [in] c1 The 2nd element from C matrix to be added with d1
+/// \param [in] c2 The 3rd element from C matrix to be added with d2
+/// \param [in] c3 The 4th element from C matrix to be added with d3
+/// \param [in] item The sycl::nd_item index space class
+template <typename MulType, typename ABType, typename CDType, typename ItemT>
+__attribute__((optnone)) void
```
zhiweij1: why add `__attribute__((optnone))` ?

clang/runtime/dpct-rt/include/dpct/math.hpp
TejaX-Alaghari: Sure, I'll add the info in the comment

clang/runtime/dpct-rt/include/dpct/math.hpp
TejaX-Alaghari: Without this attribute, the recv_a & recv_b data set by select_from_group() inside the for loop is being reset to 0 after exiting the loop due to optimizations from compiler.

There's also another option to make the array volatile, I'll try it and get back

clang/runtime/dpct-rt/include/dpct/math.hpp
TejaX-Alaghari: As this is the initial PR, I've added support for the shape used in top apps

Will add support for remaining shapes soon

clang/runtime/dpct-rt/include/dpct/math.hpp
keryell: I have the feeling the comment does not describe the case when C and D are different.

TejaX-Alaghari:
Added support for m8n8k4

---------     ---------   ----------  -----------
| Shape |     |   A   |   |    B   |  |  C / D  |
---------     ---------   ----------  -----------
m8n8k4          
.f16         .f16      .f16/.f32
                      .f64         .f64        .f64

clang/runtime/dpct-rt/include/dpct/math.hpp
```diff
@@ -2055,6 +2057,277 @@ class joint_matrix {
...
+/// \param [in] b1 The 2nd element from B matrix to be multiplied with A matrix
+/// \param [in] c0 The 1st element from C matrix to be added with d0
+/// \param [in] c1 The 2nd element from C matrix to be added with d1
+/// \param [in] c2 The 3rd element from C matrix to be added with d2
+/// \param [in] c3 The 4th element from C matrix to be added with d3
+/// \param [in] item The sycl::nd_item index space class
+template <typename MulType, typename ABType, typename CDType, typename ItemT>
+void mma(CDType *d0, CDType *d1, CDType *d2, CDType *d3, ABType a0, ABType a1,
+         ABType b0, ABType b1, CDType c0, CDType c1, CDType c2, CDType c3,
+         const ItemT &item) {
```
zhiweij1: How about designing the interface like:
```
template <typename MulType, typename ABType, typename CDType, typename ItemT,
          unsigned m, unsigned n, unsigned k>
void mma(CDType *d, const ABType* a, const ABType* b, const CDType* c, const ItemT &item);
```
Then we can merge the impls of different shapes in a single function. And in that function, we can use `if constexpr(...)` to set different branches for different shapes.

TejaX-Alaghari:
The advice from @zhiweij1 is to design the mma interface as below to unify migration of all the shapes under a single interface- 

image


with the mma function being -

image

clang/runtime/dpct-rt/include/dpct/math.hpp
zhiweij1:
```
sturct dpct_int4;
template <typename MulType, typename ABType, typename CDType, typename ItemT,
          unsigned m, unsigned n, unsigned k>
void mma(volatile void *d, const void * a, const void * b, const void * c, const ItemT &item);
```

zhiweij1: Please focus on this shape first:
```
autoGPTQ:         

mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32
 
llama.cpp:

./llama.cpp/ggml/src/ggml-cuda/mma.cuh:175:        asm("mma.sync.aligned.m16n8k16.row.col.s32.s8.s8.s32 {%0, %1, %2, %3}, {%4, %5}, {%6}, {%0, %1, %2, %3};"
 
```

tomflinda:
> The advice from @zhiweij1 is to design the mma interface as below to unify migration of all the shapes under a single interface-
> 
> image
> 
> with the mma function being -
> 
> image

OK to me.

PR #2758, opened 2025-04-01
[SYCLomatic] Support Class type of user-defined rule can use null in the Out field

zhiweij1:


tomflinda (approved):
LGTM

clang/tools/dpct/extensions/pytorch_api_rules/pytorch_api.yaml
```diff
@@ -156,7 +156,7 @@
   Kind: Class
   Priority: Takeover
   In: at::Tensor # The underlying type of torch::Tensor is at::Tensor
-  Out: torch::Tensor
+  Out: null
```
zhimingwang36: null is a new keyword here, how about if Out field is not given, then it is treat as ignored?

zhiweij1:
local lit passed

tomflinda:
local verification passed

zhimingwang36 (approved):

PR #2779, opened 2025-04-11
[SYCLomatic] For math function (impl-ed by rewriter) used in template function, only migrate the template version when there is no instantiation

zhiweij1:


tomflinda (approved):
LGTM

zhimingwang36 (approved):

PR #2790, opened 2025-04-16
[SYCLomatic]  If viewed from the translation unit, the header file <bits/stdc++.h> is included before SYCL header file, then insert the SYCL header file at the beginning of the main source file of the translation unit. 

zhiweij1:
Background:  if <bits/stdc++.h> is included before the SYCL header file,  it results in a compiler error. 

The behavior before this PR:  
If a file contains CUDA syntax, the tool inserts the SYCL header file in that file.  Also,  SYCL header file needs to be inserted at the beginning of the main file (to avoid `bits/stdc++.h` is included before SYCL header file). The solution has an issue: `sycl.hpp` may be inserted in some pure C++ code **without** checking if `bits/stdc++.h` is included or not.

This PR will be further refined as follows: 
In the translation unit, if `bits/stdc++.h` is included before the SYCL header file, insert the SYCL header file at the beginning of the main file.


tomflinda (approved):
LGTM

clang/lib/DPCT/AnalysisInfo.cpp
```diff
@@ -2537,6 +2543,10 @@ std::unordered_set<std::string>
     DpctGlobalInfo::CustomHelperFunctionAddtionalIncludes = {};
 std::unordered_map<clang::tooling::UnifiedPath, std::bitset<32>>
     DpctGlobalInfo::HeaderInsertedBitMap = {};
+bool DpctGlobalInfo::IsAfterBitsStdcxx = false;
```
zhimingwang36, line 2556: Please add comments for the purpose,  what corner case to address.

zhimingwang36 (approved):

PR #2797, opened 2025-04-22
[SYCLomatic] Fix the issue that __half type not migrated during the migration of BlockRadixSort API 

intwanghao:


zhiweij1 (approved):

clang/runtime/dpct-rt/include/dpct/group_utils.hpp
```diff
@@ -360,14 +360,18 @@ class group_radix_sort {
   }
 
 private:
+  template <typename KT>
+  using key_type =
```
ziranzha: How about moving the definition into the function? I don't see it used elsewhere.

ziranzha: How about moving the definition into the function? I don't see it used elsewhere.

ziranzha (approved):

PR #2803, opened 2025-04-24
[SYCLomatic] Add migration support for cudaKernelNodeParams, cudaKernelAddKernelNode, cudaGraphKernelNodeSetParams, cudaKernelNodeGetParams

daiyaan-ahmed6:


clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
```diff
@@ -28,20 +28,132 @@ extern DpctOption<opt, bool> AsyncHandler;
...
+                                         std::move(ReplacementStr)));
+                    emplaceTransformation(new InsertAfterStmt(BO, ")"));
+                    return;
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
```
ziranzha: Can we make it more readable? It has too many levels.

daiyaan-ahmed6: Done

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
```diff
@@ -28,20 +28,140 @@ extern DpctOption<opt, bool> AsyncHandler;
...
+      if (FieldName.empty()) {
+        report(ME->getBeginLoc(), Diagnostics::API_NOT_MIGRATED, false,
+               DpctGlobalInfo::getOriginalTypeName(ME->getBase()->getType()) +
+                   "::" + ME->getMemberDecl()->getName().str());
+        return;
+      }
+      if (FieldName == "func") {
+        auto BO = dyn_cast<BinaryOperator>(
+            getParentAsAssignedBO(ME, *Result.Context));
+        if (!BO) {
```
ziranzha, line 105: Getter isn't supported for "func", is that right?

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
```diff
@@ -28,20 +28,140 @@ extern DpctOption<opt, bool> AsyncHandler;
...
+        auto *ME_LHS = dyn_cast<MemberExpr>(LHS);
+        if (!ME_LHS) {
+          return;
+        }
+        auto *Base = ME_LHS->getBase()->IgnoreImpCasts();
+        auto *DRE = dyn_cast<DeclRefExpr>(Base);
+        if (!DRE) {
+          return;
+        }
+        auto *VD = dyn_cast<VarDecl>(DRE->getDecl());
```
ziranzha, line 118: Just replace the member, the base doesn't need change.

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
```diff
@@ -51,5 +171,47 @@ void GraphRule::runRule(const MatchFinder::MatchResult &Result) {
   EA.applyAllSubExprRepl();
 }
 
+const Expr *GraphRule::getParentAsAssignedBO(const Expr *E,
```
ziranzha: We already have this function, why define it again?

clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -94,6 +137,92 @@ class graph_mgr {
     }
   }
 
+  void add_kernel_node(dpct::experimental::node_ptr *node,
+                       dpct::experimental::command_graph_ptr graph,
+                       dpct::experimental::node_ptr *dependencies,
+                       std::size_t numberOfDependencies,
+                       dpct::experimental::kernel_node_params *params) {
+    node_graph_params_map[*node] = std::make_pair(graph, params);
```
ziranzha, line 145: Why need to save `graph`, I don't see the usage.

clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -94,6 +137,92 @@ class graph_mgr {
...
+    execGraph = new sycl::ext::oneapi::experimental::command_graph<
+        sycl::ext::oneapi::experimental::graph_state::executable>(
+        graph->finalize());
+    queue->submit(
+        [&](sycl::handler &cgh) { cgh.ext_oneapi_graph(*execGraph); });
+  }
+
+  void instantiate(dpct::experimental::command_graph_exec_ptr *execGraph,
+                   dpct::experimental::command_graph_ptr graph) {
+    exec_graph_map[*execGraph] = graph;
```
ziranzha, line 196: `execGraph` is an out param, needs to be created in the function.

clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -94,6 +137,92 @@ class graph_mgr {
...
+    }
+    graph_kernel_node_params_map[graph].emplace_back(*node, params);
+  }
+
+  void launch(dpct::experimental::command_graph_exec_ptr execGraph,
+              sycl::queue *queue) {
+    // Retrieve the graph associated with execGraph
+    auto graph = exec_graph_map[execGraph];
+    auto &kernel_params_vector = graph_kernel_node_params_map[graph];
+    for (std::size_t i = 0; i < kernel_params_vector.size(); i++) {
```
ziranzha, line 157: Why add the kernel nodes when launch, not on `add_kernel_node` called?

clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -94,6 +137,92 @@ class graph_mgr {
...
+            });
+          },
+          sycl::ext::oneapi::experimental::property::node::depends_on(
+              dependencies)));
+      if (i + 1 < kernel_params_vector.size()) {
+        auto &next_node_params = kernel_params_vector[i + 1].second;
+        auto next_dependency = next_node_params->get_dependencies()[i];
+        next_node_params->update_dependency(next_dependency, new_node);
+      }
+      node_kernel_params_pair.first = new_node;
```
ziranzha, line 185: Where is the usage of `new_node`?

clang/runtime/dpct-rt/include/dpct/graph.hpp
```diff
@@ -94,6 +137,92 @@ class graph_mgr {
...
+  }
+
+  void kernel_node_set_params(dpct::experimental::node_ptr node,
+                              dpct::experimental::kernel_node_params *params) {
+    node_graph_params_map[node].second = params;
+  }
+
+  void get_node_type(dpct::experimental::node_ptr node,
+                     sycl::ext::oneapi::experimental::node_type *nodeType) {
+    if (node_graph_params_map.find(node) != node_graph_params_map.end()) {
```
ziranzha, line 215: Why don't we use `node->get_type()` for kernel?

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
daiyaan-ahmed6: Added getter for func

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
daiyaan-ahmed6: Defined in general in RulesLang.h. Done.

clang/runtime/dpct-rt/include/dpct/graph.hpp
daiyaan-ahmed6: execGraph cannot be created as a new node until cudaGraphLaunch. Do you recommend to create a fake node to return?

clang/runtime/dpct-rt/include/dpct/graph.hpp
daiyaan-ahmed6: During `add_kernel_node` we do not have queue information. Queue information is needed to launch a host task on graph. So This PR adds the nodes in a vector during `add_kernel_node` and actually adds them to the graph during `
launch`.

clang/runtime/dpct-rt/include/dpct/graph.hpp
daiyaan-ahmed6: In DPCT, we can only launch the kernel node as a `host_task` due to kernel function pointer limitation. But some apps like llama.cpp have an if condition check that confirms if the node is of type `kernel`. So, if the node was added using `add_kernel_node`, then `get_node_type` returns the node as type `kernel` instead of `host_task`

clang/lib/DPCT/RulesLang/RulesLangGraph.cpp
daiyaan-ahmed6: Updated.

clang/runtime/dpct-rt/include/dpct/graph.hpp
daiyaan-ahmed6: In Launch, we need to add the nodes to the graph using the `graph.add({});` API

clang/runtime/dpct-rt/include/dpct/graph.hpp
daiyaan-ahmed6: The `new_node` is created and then assigned to the pointer address of the node passed in `add_kernel_node`

PR #2857, opened 2025-05-27
[DO NOT MERGE][SYCLomatic] reMigration

zhiweij1:
(1) Added a gitdiff2yaml binary
(2) Impled the logic to load the yaml file generated by (1)
(3) Impled the merge logic
(4) Added some unit tests
(5) Added a lit case controlled by the new option `--remigration` (Left some TODOs to refine, but the main logic works)


clang/lib/DPCT/IncMigration/ReMigration.cpp
```diff
@@ -0,0 +1,777 @@
...
+    } else {
+      // ItB->first == ItA->first
+      if (ItB->second == ItA->second) {
+        // No conflict, just add one of them
+        Result.push_back(genReplacement(ItA->first, ItA->second));
+        ++ItA;
+        ++ItB;
+        continue;
+      }
+      // Conflict line(s)
```
zhiweij1, line 392: We can do some refinements here.
If Repl_B and Repl_D have a conflict at one line, and no user-change in this line, we can use Repl_B directly.

PR #2861, opened 2025-05-29
[SYCLomatic][Python Migration][Bug] Fix case_010_include_rule_file lit case

TejaX-Alaghari:
This PR fixes the flaky Python rule file inclusion bug which fails when including rule files

The root cause of this issue is due to using `llvm::StringRef` objects for tracking the processed YAML files in a `llvm::DensetSet`.
The contents of the `llvm::StringRef` objects change for each call to `readYAMLFile()` function.
And at times, the string data pointed by the `llvm::StringRef` objects stored in the `llvm::DenseSet` are altered unintentionally causing rule files to be skipped from including.

This issue is fixed by using `std::unordered_set<std::string>` in place of `llvm::DenseSet<llvm::StringRef>` where `std::string` objects have the advantage of owning the memory of char array they store.


zhiweij1 (approved):

tomflinda (approved):
LGTM

PR #2885, opened 2025-06-10
Added launch_config objects to pass kernel properties

TejaX-Alaghari:
This PR uses `launch_config` objects to pass compile-time kernel properties during kernel launch ([sycl_ext_oneapi_enqueue_functions](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_enqueue_functions.asciidoc#:~:text=sycl%3A%3Arange.-,namespace%20sycl%3A%3Aext%3A%3Aoneapi%3A%3Aexperimental%20%7B%0A%0Atemplate%20%3Ctypename%20KernelName%2C,basic%20kernel%2C%20using%20the%20launch%20configuration%20specified%20by%20a%20launch_config.,-namespace%20sycl%3A%3Aext)).

This behavior replaces the existing deprecated behavior of passing the properties as an argument directly during kernel submission ([queue.hpp](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/queue.hpp#L2796)).

### DIFF
--- a/clang/test/dpct/sync_api_noneusm.cu
+++ b/clang/test/dpct/sync_api_noneusm.cu
@@ -105,9 +105,12 @@ int main() {
   // CHECK: {
   // CHECK-NEXT: auto exp_props = sycl::ext::oneapi::experimental::properties{sycl::ext::oneapi::experimental::use_root_sync};
   // CHECK-EMPTY:
-  // CHECK-NEXT:  dpct::get_out_of_order_queue().parallel_for(
-  // CHECK-NEXT:  sycl::nd_range<3>(sycl::range<3>(1, 1, 2) * sycl::range<3>(1, 1, 2), sycl::range<3>(1, 1, 2)),
-  // CHECK-NEXT:  exp_props,     [=](sycl::nd_item<3> item_ct1) {
+  // CHECK-NEXT:  sycl::ext::oneapi::experimental::nd_launch(
+  // CHECK-NEXT:    dpct::get_out_of_order_queue(),
+  // CHECK-NEXT:    sycl::ext::oneapi::experimental::launch_config(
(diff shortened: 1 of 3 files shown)


clang/test/dpct/sync_api_noneusm.cu
```diff
@@ -105,9 +105,12 @@ int main() {
   // CHECK: {
   // CHECK-NEXT: auto exp_props = sycl::ext::oneapi::experimental::properties{sycl::ext::oneapi::experimental::use_root_sync};
   // CHECK-EMPTY:
-  // CHECK-NEXT:  dpct::get_out_of_order_queue().parallel_for(
-  // CHECK-NEXT:  sycl::nd_range<3>(sycl::range<3>(1, 1, 2) * sycl::range<3>(1, 1, 2), sycl::range<3>(1, 1, 2)),
-  // CHECK-NEXT:  exp_props,     [=](sycl::nd_item<3> item_ct1) {
+  // CHECK-NEXT:  sycl::ext::oneapi::experimental::nd_launch(
```
ziranzha, line 108: Why use `nd_launch` here instead of `parallel_for(launch_config...)`?
By the way, if we use `nd_launch`, we should not use `KernelPrinter`; use `CallExprPrinter` instead.

TejaX-Alaghari: `parallel_for(launch_config...)` is used when the work-item range is specified using `sycl::range` object

When `sycl::nd_range` object is used `nd_range(launch_config...)` to be used ([ref](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_enqueue_functions.asciidoc#nd-range-kernels:~:text=Effects%3A%20Enqueues%20a%20kernel%20function%20to%20the%20sycl%3A%3Aqueue%20or%20sycl%3A%3Ahandler%20as%20a%20basic%20kernel%2C%20using%20the%20number%20of%20work%2Ditems%20specified%20by%20a%20sycl%3A%3And_range.))

PR #2978, opened 2026-02-23
Update to official oneDPL histogram implementation

danhoeflinger:
This PR updates the histogram implementation from a temporary feature preview to use the official oneDPL interfaces, and removes the temporary histogram code.

It also removes a `size()` function from device iterator which causes problems with some internals of oneDPL separating ranges and iterator implementations.  This size function seems strange (size of an iterator?), and only appears in one of the `device_iterator` implementations.  

Technically this is removing part of a public interface of `device_iterator`, but I'm not sure how it makes any sense, and makes usage within oneDPL problematic.


clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
```diff
@@ -1899,9 +1427,9 @@ ::std::enable_if_t<
     internal::is_hetero_execution_policy<::std::decay_t<Policy>>::value>
 histogram_even(Policy &&policy, Iter1 d_samples, Iter2 d_histogram,
                int num_levels, T lower_level, T upper_level, Size num_samples) {
-  internal::dpl_histogram::histogram(::std::forward<Policy>(policy), d_samples,
-                                     d_samples + num_samples, d_histogram,
-                                     num_levels - 1, lower_level, upper_level);
+  oneapi::dpl::histogram(::std::forward<Policy>(policy), d_samples,
+                                     d_samples + num_samples, num_levels - 1,
```
mmichel11: Do we need to run clang-format again here?

mmichel11 (commented):
Changes LGTM

clang/runtime/dpct-rt/include/dpct/dpl_extras/algorithm.h
danhoeflinger: mmm, yes.  Fixed.

timmiesmith (approved):
LGTM. Thank you.

The iterator.size() method doesn't make sense, but is likely an old workaround. It is enabling the USM-based device_iterator to be treated as a SYCL buffer backed device_iterator. If the iterator is defined over a buffer you can use it.get_buffer().size() to get the size of the input range.  The USM-based device_iterator is returning `*this` for get_buffer() and thus needed a nonsense size method in order for that usage in an algorithm to compile.

timmiesmith:
CI Windows failures are unrelated to the PR.

danhoeflinger:
> The iterator.size() method doesn't make sense, but is likely an old workaround. It is enabling the USM-based device_iterator to be treated as a SYCL buffer backed device_iterator. If the iterator is defined over a buffer you can use it.get_buffer().size() to get the size of the input range. The USM-based device_iterator is returning `*this` for get_buffer() and thus needed a nonsense size method in order for that usage in an algorithm to compile.

@timmiesmith 
Hmm, this makes some sense but also sets off some alarm bells for me.  I'm wondering if we should fully remove the `get_buffer` which returns `*this` as well, or if we need to replace it with some proxy which also provides the `size()` function to the best of its ability.  Similar to the removed one here, I don't think we have access to this info in reality.

Do you know where if anywhere this functionality is actually used?  Since `is_hetero = std::false_type` (and it is "passed directly"), I don't expect it to be used within oneDPL at all.  I'm not finding any place in SYCLomatic, or the SYCLomatic-tests. There are plenty of `get_buffer()` calls for oneDPL buffer_wrappers (sycl_iterators), and some for `device_vector`, but I see none for `device_pointer`.

danhoeflinger:
> @timmiesmith Hmm, this makes some sense but also sets off some alarm bells for me. I'm wondering if we should fully remove the `get_buffer` which returns `*this` as well...

I've removed get_buffer() and this still looks good in the tests. I don't think it is used anywhere.

timmiesmith (approved):
LGTM. Thank you.

dmitriy-sobolev (approved):
Histogram replacement looks good to me.
As for `size()` and `get_buffer()`, I trust Timmie's judgement.

PR #2980, opened 2026-02-27
Use indirectly_device_accessible not "is_passed_directly"

danhoeflinger:
Replace is_passed_directly unofficial interface with the new indirectly device accessible customization point everywhere in SYCLomatic.


timmiesmith (approved):
LGTM. Thank you for updating the iterator support here to use current oneDPL support.

timmiesmith (approved):
LGTM. Thank you.

tomflinda (approved):

PR #2997, opened 2026-04-10
Add backwards compatibility with older oneAPI compilers for non-uniform group APIs

mmichel11:
Follow-up to https://github.com/oneapi-src/SYCLomatic/commit/fe2a343d69a6668919a27d5be6ae161b1bbbc2ba as we have use cases to maintain compatibility with different compiler versions.

If a user is using a compiler built from https://github.com/intel/llvm, then this patch expects it to be a recent build using the latest version of the non-uniform group APIs. All oneAPI compilers that support the non-uniform group extension are supported.


timmiesmith (approved):
LGTM. Thank you.

danhoeflinger (approved):
LGTM, thanks

